PR #244, opened 2018-02-01
Add a hook to add a new output handler

mdboom:
This provides a way for third-party library code (or notebook code) to add a custom output handler.  This is complementary to the approach already on master whereby a special method is looked for on the return value, since it could be used, for example, to provide special outputs for vanilla data types, or at least notebook-unaware data types.

This is very much a work-in-progress, but I thought I'd submit now to get feedback / advice.

An example has been added to the "Output Handling" notebook, which uses the custom output handler plugin here:

https://raw.githubusercontent.com/mdboom/Iodide-playground/6f638bdbce330c4a46cd3ba8a674fa4d94829ffd/custom-output-handler.js

~~- This introduces the concept of a "public API" to the notebook.  This is where I imagine we would start adding functions so third-party code (or notebooks themselves) could hook into or manipulate the notebook.  This inserts an object `nb_api` object to the `window` object where this stuff would live.  I'm always wary of injecting things into the global namespace, but I couldn't find a better way -- I'm probably just showing my JS ignorance here.  That said, must Javascript "plugin"/"module" concepts are around compiletime, not runtime extension as we need to do here.~~

~~- The `nb_api` object I injected (as well as the `nd` object that master already has), only seems to be available and show up in the declared variables inspector after explicitly asking for it (i.e. after running a cell with `nb_api` in it).  Is that normal?  Downside is that you can't load the plugin before running a cell.~~

- When adding a hook, it would probably be a good idea to validate it against some expected interface.  I know the `interface` keyword isn't yet in the language.  Does anyone have experience with libraries for this like `implement-js`?  

EDIT: The new plugin API is this.  If a cell or external dependency has the specially-named function `iodidePlugin`, it is called, passing in the `iodideApi` object, which has all of the functions in the public API.  (This is a model that a lot of Python plugin frameworks use, such as the one in Sphinx).  The advantage here is that there's no pollution of the global namespace (by us at least), and registering plugin hooks isn't a side effect of simply importing the code.  I know it's not quite right where I'm looking for the `iodidePlugin` callback, since it won't find it in the second plugin it loads, but I'm getting closer...

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,11 +1,18 @@
 // The "Public API" for notebooks. This lets notebooks and third-party plugins
 // extend and manipulate the notebook
 
-import React from 'react'
 import { addOutputHandler } from './components/output.jsx'
 
-export const nb_api = {
+export const iodideApi = {
   'addOutputHandler': addOutputHandler
 }
 
-export default nb_api
+// This is called on every new object in a cell or external dependency.
+// If it looks like an iodideSetup plugin callback, it is called.
+export function handlePlugin(name, value) {
+  if (name === 'iodideSetup' && typeof value === 'function') {
+    value(iodideApi)
+  }
+}
+
+export default iodideApi
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -14,9 +14,6 @@ import Page from './components/page'
 import { store } from './store'
 
 import './tools/nd'
-import nb_api from './api'
-
-window.nb_api = nb_api
 
 render(
   <Provider store={store}>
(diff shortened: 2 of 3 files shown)


hamilton:
I can't wait until we get this running! Will be great to start providing some graphing libraries for folks.

@mdboom do you think we should also support having the `render` function also allow us to return [`Element`](https://developer.mozilla.org/en-US/docs/Web/API/Element) objects, like `HTMLElement` or `SVGElement`? It's also possible we could return React elements as well, which might be the absolute fastest way to let users write worthwhile components. I'm not saying the output renderers would have to be `jsx` but after transpilation we could make it easy to use React components.

This doesn't have to be a requirement for this PR - just spitballing some ideas.

mdboom:
Superceded by #343

PR #272, opened 2018-02-08
WIP: Fixes / Refactors for ExternalResourceCell

hamilton:
This will be a few commits centered around making the `ExternalResourceCell` better, testable, and not-broken.

### DIFF
--- a/src/reducers/cell-reducer-utils.js
+++ b/src/reducers/cell-reducer-utils.js
@@ -107,12 +107,13 @@ function addExternalDependency(dep) {
     elem.rel = 'stylesheet'
     elem.type = 'text/css'
     elem.href = src
-    if (elem.sizes.length === 0) {
-      outElem.status = 'error'
-      outElem.statusExplanation = 'style sheet is empty'
-    } else {
-      outElem.status = 'loaded'
-    }
+    outElem.status = 'loaded'
+    // if (elem.sizes.length === 0) {
+    //   outElem.status = 'error'
+    //   outElem.statusExplanation = 'style sheet is empty'
+    // } else {
+    //   outElem.status = 'loaded'
+    // }
   } else {
     outElem.status = 'error'
     outElem.statusExplanation = 'unknown dependency type.'


hamilton:
Ending the work day here:

screen shot 2018-02-08 at 5 33 38 pm

hamilton:
At this point I think I shouldn't spend too much more time on `ExternalDependencyCell` if we have plans to deprecate it with #288 soon. For now, I've made a number of improvements to it -

- smarter handling of the input parsing
- cleaned up the css line bug
- added the globals introduced by the script loading
- refactored the `ExternalDependencyCell` and added a more nuanced output handler, pulled out of the file for use when we get to stdlib

hamilton:
Reviewers (@bcolloran or @wlach or @mdboom) the best way to check this out is to put this into an `ExternalDependencyCell`:

```
https://cdnjs.cloudflare.com/ajax/libs/d3/4.13.0/d3.min.js
https://cdnjs.cloudflare.com/ajax/libs/timelinejs/2.36.0/js/timeline-min.js
https://cdnjs.cloudflare.com/ajax/libs/timelinejs/2.36.0/js/blahblahblah.js
```

src/components/cell-type-external-resource.jsx
```diff
@@ -1,84 +1,38 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import { bindActionCreators } from 'redux'
```
bcolloran: looks like bindActionCreators is not used?

src/components/cell-type-external-resource.jsx
```diff
@@ -1,84 +1,38 @@
...
-    })
-    return (
-      <div className="dependency-output">
-        {outs}
-      </div>
-    )
+export class ExternalResourceCellUnconnected extends React.Component {
+  static propTypes = {
+    cellId: PropTypes.number.isRequired,
+    value: PropTypes.array.isRequired,
```
bcolloran: i'm getting "warning.js:33 Warning: Failed prop type: The prop `value` is marked as required in `ExternalResourceCellUnconnected`, but its value is `undefined`.";

is this prop not required?

src/components/output-handler-external-resource.jsx
```diff
@@ -0,0 +1,66 @@
...
+import lightBaseTheme from 'material-ui/styles/baseThemes/lightBaseTheme'
+import getMuiTheme from 'material-ui/styles/getMuiTheme'
+import CheckCircle from 'material-ui/svg-icons/action/check-circle'
+import ErrorCircle from 'material-ui/svg-icons/alert/error'
+import UnloadedCircle from 'material-ui/svg-icons/content/remove'
+
+// this is merely a presentational component, and as such can be passed
+// all the data it needs.
+export default class ExternalResourceOutput extends React.Component {
+  render() {
```
bcolloran: add propTypes

src/components/cell-type-external-resource.jsx
bcolloran: i.e., should it just be  `value: PropTypes.array`?

hamilton: I'll remove `.isRequired` from it since it will often be `undefined`. Not sure why I didn't get this error before.

src/components/output-handler-external-resource.jsx
hamilton: fixed

PR #280, opened 2018-02-10
Add a dialog that requires confirmation before deleting a notebook

horcrux2301:
Added an alert that pops up before deleting a file.

iodide gif

### DIFF
--- a/src/components/menu-component.jsx
+++ b/src/components/menu-component.jsx
@@ -71,7 +71,7 @@ class MainMenu extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      isModalOpen: false,
+      isDeleteNotebookDialogOpen: false,
     }
     this.insertCell = this.insertCell.bind(this)
     this.cellUp = this.cellUp.bind(this)
@@ -79,25 +79,25 @@ class MainMenu extends React.Component {
     this.runCell = this.runCell.bind(this)
     this.deleteNotebook = this.deleteNotebook.bind(this)
     this.runAllCells = this.runAllCells.bind(this)
-    this.switchModal = this.switchModal.bind(this)
-    this.closeModalAndDeleteNotebook = this.closeModalAndDeleteNotebook.bind(this)
+    this.switchDeleteNotebookDialog = this.switchDeleteNotebookDialog.bind(this)
+    this.closeDialogAndDeleteNotebook = this.closeDialogAndDeleteNotebook.bind(this)
   }
 
-  switchModal() {
+  switchDeleteNotebookDialog() {
     this.setState({
-      isModalOpen: !this.state.isModalOpen,
+      isDeleteNotebookDialogOpen: !this.state.isDeleteNotebookDialogOpen,
     })
   }
 
-  closeModalAndDeleteNotebook() {
+  closeDialogAndDeleteNotebook() {
     this.setState({
-      isModalOpen: !this.state.isModalOpen,
+      isDeleteNotebookDialogOpen: !this.state.isDeleteNotebookDialogOpen,
     })
     this.props.actions.deleteNotebook(this.props.title)
   }
 
   deleteNotebook(notebook) {
-    this.switchModal()
+    this.switchDeleteNotebookDialog()
   }
 
   runAllCells() {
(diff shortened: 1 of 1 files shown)


src/components/menu-component.jsx
```diff
@@ -104,10 +124,27 @@ class MainMenu extends React.Component {
 
 
   render() {
+    const actions = [
```
hamilton: Could you change this variable name to `deleteNotebookDialogOptions`? `actions` are already a term understood in the Redux world.

src/components/menu-component.jsx
```diff
@@ -104,10 +124,27 @@ class MainMenu extends React.Component {
...
+      <FlatButton
+        label="Cancel"
+        primary
+        onClick={this.switchModal}
+      />,
+      <FlatButton
+        label="Submit"
```
hamilton: `label="delete"`

src/components/menu-component.jsx
```diff
@@ -68,16 +70,34 @@ function menuComponents(items, parentComponent) {
...
+    this.state = {
+      isModalOpen: false,
+    }
     this.insertCell = this.insertCell.bind(this)
     this.cellUp = this.cellUp.bind(this)
     this.cellDown = this.cellDown.bind(this)
     this.runCell = this.runCell.bind(this)
     this.deleteNotebook = this.deleteNotebook.bind(this)
     this.runAllCells = this.runAllCells.bind(this)
+    this.switchModal = this.switchModal.bind(this)
```
hamilton: Let's use the term `dialog` instead of `modal` on this file to agree with the material ui naming convention.

src/components/menu-component.jsx
```diff
@@ -104,10 +124,27 @@ class MainMenu extends React.Component {
...
+        onClick={this.closeModalAndDeleteNotebook}
+      />,
+    ];
     const mc = menuComponents(menuItems, this)
 
     return (
       <ToolbarGroup firstChild={this.props.firstChild}>
+        <Dialog
+          open={this.state.isModalOpen}
+          title="Are you sure you want to delete this file?"
```
hamilton: - Could you remove the text from the `title` and put it in line between `<Dialog etc...> {text here} </Dialog>`? This doesn't need a title.
- The text should read "Delete notebook?"

src/components/menu-component.jsx
hamilton: We need a more descriptive name for this, like `this.switchDeleteNotebookDialog`, so it is not ambiguous what the dialog does.

src/components/menu-component.jsx
```diff
@@ -68,16 +70,34 @@ function menuComponents(items, parentComponent) {
...
+    this.closeModalAndDeleteNotebook = this.closeModalAndDeleteNotebook.bind(this)
+  }
+
+  switchModal() {
+    this.setState({
+      isModalOpen: !this.state.isModalOpen,
+    })
+  }
+
+  closeModalAndDeleteNotebook() {
```
hamilton: modal -> dialog

src/components/menu-component.jsx
```diff
@@ -68,16 +70,34 @@ function menuComponents(items, parentComponent) {
 class MainMenu extends React.Component {
   constructor(props) {
     super(props)
+    this.state = {
+      isModalOpen: false,
```
hamilton: `isDeleteNotebookDialogOpen` is probably a better naming convention than this, for now.

We will likely refactor this down the road to make room for other dialogs, but for until we refactor how we're dealing with menus I think this approach works just fine.

hamilton:
Forgot to put this in my review, but ... thank you for the PR!

horcrux2301:
I have added all the changes. I will be more careful in selecting the names for variables and functions from next time.

hamilton:
@horcrux2301 no problem - thanks for the changes. Your PR is demonstrating that, regarding UI changes, the iodide core members should add the material ui guide as a reference to `CONTRIBUTING.md` (in this case, the [dialog reference](https://material.io/guidelines/components/dialogs.html) would have been useful) so it is clear to all how to proceed with these kinds of improvements without much feedback.

Will review now.

hamilton:
Thanks. This looks good to me. 

Because we are overriding the keybindings, we cannot cancel the dialog with the `esc` key (which is a default for `material-ui` components), so I'm going to file a separate issue for this.

bcolloran:
nice work, thanks @horcrux2301!

Issue #273, opened 2018-02-08
change "run all cells" implementation to step through cells in order rather than recursively updating state


bcolloran:
looks like MD cells can are not fully rendered before the js in subsequent cells is executed -- when I "run all", i'm observing a condition in which dom elements defined inline are not found when I query for then in the next code cell

bcolloran:
investigated this a little more-- the issue is that ATM `RUN_ALL_CELLS` is a single action in the reducer, so it applies *all* the changes to state in one action, i.e. between react render cycles.

The solution is to have a separate chunk of code that just sequentially fires the `RUN_CELL` action in order on each cell

PR #307, opened 2018-02-12
Run all fix 2

bcolloran:
hamilton, please sanity check me on this approach to "run all". it feels ugly, but i rat holed for probably at least 6 hours trying to come up with a pretty way to make sure that all React DOM updates are flushed before evaling subsequent JS cells, and I never really found one. (for future reference: there is no obvious place to put a "componentDidUpdate" that listens for this).

even though i'm not loving this solution, is does seem to fix the bug in #273, which is an important bug, so i'm willing to live with it and fix the foibles later


bcolloran:
discussed with hamilton, merging

PR #313, opened 2018-02-12
Deprecate DOM cell type

### LINKED ISSUE #306: deprecate DOM cell type and remove all traces of it from codebase
We currently have a cell of type "DOM", which lets you create divs or some other element with an id. I'm personally not convinced that this is especially useful, since as far as I can tell you can get the same behavior by creating a markdown cell and creating some html inline:

```
<div id="mycoolelement">
</div>
```

Is there some motivation for these cells that I'm not aware of?

pradeepgangwar:
Fixes #306 

### Changes done:
- Removed all traces of DOM cell.
- Fixed examples.


bcolloran:
very nice! thank you!!

PR #325, opened 2018-02-13
Bug in history item

### LINKED ISSUE #323: No `id` in `cell` object in `history`
Short summary of the bug. 
Bug in MASTER

### What I Did
1. Open Console
1. Write `a = 10` in the notebook
1. Press `Shift+Enter`

### What I Expected
There should be no errors in the console

### What Happened 
The following error occured.
```
Warning: Failed prop type: The prop `cell.id` is marked as required in `HistoryItem`, but its value is `undefined`.
    in HistoryItem (created by NotebookHeader)
    in NotebookHeader (created by Connect(NotebookHeader))
    in Connect(NotebookHeader) (created by Page)
    in div (created by Page)
    in Page (created by Connect(Page))
    in Connect(Page)
    in Provider
```
### What is causing the bug in my opinion
In the following code
```
histContents = this.props.history.filter(cell => cell.content.length).map((cell, i) => {
        // TODO: Don't use array indices in keys (See react/no-array-index-key linter)
        const cellComponent = <HistoryItem display ref={`cell${cell.id}`} cell={cell} id={`${i}-${cell.id}`} key={`history${i}`} />  // eslint-disable-line
        return cellComponent
      })
```
I tried to console log the cell, the cell object looks like this
```
{ cellID: 0, lastRan: Tue Feb 13 2018 11:35:33 GMT+0530 (IST), content: "a = 10"}
```
so there is no `id` in `cell` object. I think this is causing the error.

pbteja1998:
Fixes #323


bcolloran:
looks like #323 is already fixed elsewhere; thanks @pbteja1998  and @hereisnaman  for figuring it out!

Issue #330, opened 2018-02-13
explore refactoring away OneRowCell and TwoRow cell


bcolloran:
these container components may be obsolete now, but i'm not sure. follows from #219

bcolloran:
closing, we're gonna do this

Issue #329, opened 2018-02-13
need per-row UI to allow hiding/showing cell rows in presentation mode


bcolloran:
follows from #219

bcolloran:
we need to discuss how we might want this to work.

hamilton:
I have a pretty clear idea for this which also solves #312.

hamilton:
@bcolloran is this relevant anymore, given the direction we're going?

bcolloran:
nope!

Issue #317, opened 2018-02-13
[META] add some onboarding for first-time notebook users


hamilton:
Inspired by #315 - new user completely baffled about what the heck to do.

- there is no way for new users to know what to do first when they're dropped in a notebook.
- there is no indication of what they can put in a cell when they're dropped in. This would be fixed by both default gray-out text and #312 / #187 
- we should make it easy to get right to links, guides, etc. when they're in the notebook.
- any annotations should be hideable with one click such as "I'm good - don't show this again"

bcolloran:
I'm working on this right now

On Feb 12, 2018 4:33 PM, "Hamilton Ulmer" <notifications@github.com> wrote:

Inspired by #315 <https://github.com/iodide-project/iodide/issues/315> -
new user completely baffled about what the heck to do.

   - there is no way for new users to know what to do first when they're
   dropped in a notebook.
   - there is no indication of what they can put in a cell when they're
   dropped in. This would be fixed by both default gray-out text and #312
   <https://github.com/iodide-project/iodide/pull/312> / #187
   <https://github.com/iodide-project/iodide/issues/187>
   - we should make it easy to get right to links, guides, etc. when
   they're in the notebook.
   - any annotations should be hideable with one click such as "I'm good -
   don't show this again"

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/iodide-project/iodide/issues/317>, or mute the thread

bcolloran:
closed with #377

Issue #315, opened 2018-02-12
Add some kind of introduction text or link


jaredkerim:
Short summary of the bug. Please include the IODIDE version number where you found the bug

Iodide version: How do i find this

### What I Did

I opened iodide

### What I Expected

I expected some text or link to explain what to do next

### What Happened 

I got some buttons and an input dialog with no instructions

jaredkerim:
And some dummy data with examples would be nice

hamilton:
Thanks for the issue! We actually have all these things available but when you're dropped into a notebook there is no real indication of what to do next. We should figure out how to handle this.

lol "Iodide version: How do i find this" - @bcolloran this is a pretty important issue to add, if we're using this issue template.

hamilton:
bcolloran has a great tutorial notebook that landed in a recent PR. We should also surface this tutorial in the docs so it is obvious how to get started.

bcolloran:
i'll take on the task of making sure that the demo notebook is deployed as the landing page.

mdboom:
Yeah, demo notebook as landing page is a good quick fix. It would be fun to use something like "[Bootstrap Tour](http://bootstraptour.com)" eventually to give users a quick walk through of the interface.

bcolloran:
closed with #377

PR #332, opened 2018-02-13
Collapse state fix 2

### LINKED ISSUE #219: rework defaults and UI for cell collapse/scrollable/expanded states
I want to rethink how cell collapse state works, particularly WRT presentation mode. Since we allow readers (of view-mode docs) to access the full code behind an analysis just by clicking the 'edit' button, we don't have a strong need to make the JS viewable in-line at all (even as a collapsed pill), so i'm thinking this should be an opt-in feature. that is, by default, state manipulation cells (js, resource, css, etc) should be completely invisible in presentation mode (as should raw cells). The notebook author should be able to make these cells visible if they want to do so for the flow of their narrative, but that should not be the default.

I'm less sure about this part, but I think this also goes for outputs -- they should probably be hidden too, since often in the course of hacking on a notebook you'll want to have intermediate results inline to help you find your way through a problem, but you don't want to have to clean them up for view mode; and in fact, doing could make it harder for people digging into your code in edit mode to follow it. So perhaps the general principle for view mode should be: other than MD and DOM cells (the outputs of which are always shown by default), a NB author has to make an explicit choice to have something shown in View mode. (And conversely, I guess a NB author could choose to hide MD or DOM cells for some reason)

To get the UX to work for these options, we might need to re-introduce a per-cell options menu

see also #144

### LINKED ISSUE #269: Update the evictions notice analysis to apply the modern cell types

bcolloran:
i know i need to get out of the habit of these insanely large PRs, esp now that we have so many contributors, so I apologize for this. Sometimes, tho, it ends up feeling more efficient if I can work safely in a big branch and know that that work will be internally consistent.

In any case, this should fix #269, #224, and #255, which will close out milestone 0.3a entirely (the long awaited JSMD example files are here!)

it also satisfactorily fixes #219, and #144, cleaning up a bunch of stuff around cell rows and cell containers, in the process cleaning up a bit of the corresponding css. there's follow up work to be done in #329 and #330, but it can wait.

once this is merged, i'll be a whisper away from fixing #317 and #315 with a revision of the "what does a web notebook look like" example that I've been working on, and which can be used as the new default intro NB on the GH page. You can see it in the pre-loaded examples. (and once we merge this PR, it will be easy to submit PRs to enhance that and the other examples, b/c they'll be diffable JSMD! woo!)

i'm hoping one of you can can kick the tires on this and we can merge it early tomorrow. All the tests are green, the main thing to do is check out this branch, try the example notebooks, and make sure everything looks how we want.


hamilton:
Great - will review this once I’m in this morning!

hamilton:
I am testing this branch and get the following error:

screen shot 2018-02-13 at 9 54 07 am

src/jsmd-tools.js
```diff
@@ -111,29 +135,48 @@ function parseJsmd(jsmd) {
...
-  ['javascript', 'js'],
-  ['markdown', 'md'],
-  ['external dependencies', 'resource'],
-  ['raw', 'raw'],
-])
+function stateFromJsmd(jsmdString) {
+  const parsed = parseJsmd(jsmdString)
+  let { cells } = parsed
+  const { parseWarnings } = parsed
+  console.log(parseWarnings)
```
hamilton: let's not log parseWarnings unless this array has some contents, and perhaps we use `console.warn` over the contents.

hamilton:
There seems to be a bug in this implementation at the moment that @bcolloran is investigating where new notebook content doesn't show up in presentation mode - @wlach / @mdboom I'd hold off taking a look for now.

src/jsmd-tools.js
bcolloran: good idea, fixed

PR #333, opened 2018-02-13
External Resource Cell (redo PR)

hamilton:
Paves over the rebase nightmare I introduced in #272 but has all the same work.


hamilton:
@mdboom I agree. It is confusing, and is something I was hoping to address in today's meeting but got swept up! I suspect we will begin to handle these issues once we hone in on the handlers.

Issue #337, opened 2018-02-13
CSSCell does not update the style tag


hamilton:
If I update a `CSSCell` with new styles and attempt to execute it again, the style does not update. You can use the inspector to verify.

hamilton:
Fixed with #338

PR #338, opened 2018-02-13
fixes the CSSCell update bug and corrects the test cases

hamilton:
- fixes the bug described in #337 
- updates the cell tests

example below. Run the css cell, then change a value, then run the css cell again to see if the output style changes. I suggest changing `border-radius` from `10px` to `100px`.

```
# markdown cell:
<div id='test' />

# css cell:
div#test {
  width:300px;
  height:300px;
  background-color:tomato;
  border-radius:10px;
}

```


hamilton:
thx for the review @mdboom

PR #339, opened 2018-02-13
Automatically deploy to Github pages from Travis

### LINKED ISSUE #334: auto-deploy master version of app for pages
We found that it is kind of a shame that the notebook served in https://iodide-project.github.io/iodide/ must be explicitly updated whenever we want to give users the newest changes. Given we are moving very quickly, we will just serve master to `gh-pages`.

@mdboom is taking the lead on this, with any needed assistance from @wlach.

mdboom:
Fixes #334

This automatically deploys each commit on master to Github pages, to the URL:

https://iodide-project.github.io/master/

These are deployed to a separate repository at iodide-project/master, since filling up the main dev repo with build products for every commit would soon balloon in size.  Even in the `master` repo, the commit is "orphaned" so the history is not actually retained.

I did make one substantial change to what `npm build` does.  Rather than pulling the version from the `package.json` file, it queries the git repo.  If the current commit is a tag, it uses that for a version number, otherwise it uses the git commit hash.  This ensures that the version in the build product is always exactly what it says it is.  (Presently, it creates a 0.2.0 release from all commits leading up to the 0.2.0 release, which is sure to cause confusion).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ script:
 
 before_deploy:
   - npm run build
-  - cp prod/iodide.${TRAVIS_COMMIT}.html prod/index.html
+  - cp prod/iodide.${TRAVIS_BRANCH}.html prod/index.html
 
 deploy:
   provider: pages
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,32 @@
 language: node_js
+
 node_js:
   - "lts/*"
   - "8"
+
 install:
   - npm install
+
 script:
   - npm run lint
   - npm test
+
+before_deploy:
+  - npm run build
+  - cp prod/iodide.${TRAVIS_COMMIT}.html prod/index.html
+
+deploy:
+  provider: pages
+  github-token: "$GITHUB_TOKEN"
+  skip-cleanup: true
+  keep-history: false
+  repo: iodide-project/master
+  verbose: false
+  local-dir: prod
+  target-branch: master
+  on:
+    branch: master
+
+env:
+  global:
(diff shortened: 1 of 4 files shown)


bcolloran:
Oh this new workflow sounds great, much cleaner! Thanks @mdboom! Just to confirm I'm understanding (I'm new to all this, have never used tags before)

(1) if a commit is a "tag" it will build under its tagged version number to with resource URIs set to 'https://iodide-project.github.io/iodide/dist/' so that we can manually deploy it in the `docs/dist` directory from the "stable" main URL at "https://iodide-project.github.io/iodide"

(2) for all other commits to master, it'll build with resource URIs set to 'https://iodide-project.github.io/master/' and be autodeployed at the "latest" URL "https://iodide-project.github.io/master"; and new commits to master will just be added to the iodide/master repo, so exported notebooks will continue to point at the js/css resources that were associated with the NB at the time of export?

If I'm understanding that right, it sounds *perfect*, and I'll go ahead and change the URL in the readme to point at "https://iodide-project.github.io/master"

mdboom:
@bcolloran: Your understanding is mostly correct.  However, I forgot about exporting notebooks, and I think this approach might need some tweaking for that.  If a notebook is exported from `https://iodide-project.github.io/master`, the resources will point back to `https://iodide-project.github.io/master/iodide.${GIT_COMMIT_HASH}.js` etc.  When the master repo is updating with a new version, the old hash will disappear and it will 404.

We could change it so the files are "iodide.master.js" instead (i.e. don't include the git hash). That does mean that reloading an exported notebook later may break or change behavior.  But I guess it's safe to say that notebooks exported from master are just sort of unsupported?

bcolloran:
@mdboom gotcha sounds good. I think it's fair to say that notebooks on master are unsupported, and that we won't offer guarantees about notebooks exported from master. I think your suggestion of using "iodide.master.*" sounds like a good way forward. Let's merge after that little tweak.

mdboom:
@bcolloran: I've changed this so the output files on a non-tagged commit use the branch name rather than the hash.  It won't actually *work* until we merge, since it's set to only deploy from the master branch and right now this PR is a branch-merging-into-the-master-branch.

bcolloran:
well then let's merge it! :-)

mdboom:
Ugh.  Looks like git-rev-sync doesn't really work on Travis (because they set up the branches weirdly).  Will have a fix shortly.

PR #340, opened 2018-02-14
updated the ReactTable implementation to JSON.stringify other objects…

hamilton:
…, and .toString any Date objects

To test, try putting this into a `JSCell`:

```javascript
var x = [
    {date: new Date(), obj: {a:10,b:20}, test: 'hello', test2: 1, test3: null},
    {date: new Date(), obj: {a:30,b:40}, test: 'goodbye', test2: 39394394, test3: null},
]
x
```

Tests forthcoming, but I wanted to fix this bug before we had too many dogfooders blocked by it.

### DIFF
--- a/src/components/output.jsx
+++ b/src/components/output.jsx
@@ -27,7 +27,20 @@ const dataFrameHandler = {
   shouldHandle: value => nb.isRowDf(value),
 
   render: (value) => {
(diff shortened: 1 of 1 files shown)


hamilton:
@mdboom I'm good with that, especially once we deprecate `ExternalResourceCell`.

Issue #288, opened 2018-02-11
iodide needs a stdlib for common notebook-centric operations


bcolloran:
we've talked for a while about having a small library of built-in functions; following from our discussion the other day, i think our approach will feel more unified (most stuff happens in code, not in special cells), and our app code will be simplified a bit if we move things to a set of library functions that are called within the normal eval paradigm.

These are just a few ideas to get the conversation rolling (we can pick any namespace we think works best, but in these examples i'll just use `io.`; could also be "IO.", "iodide.", "nb.", "NB.", etc) --

`io.DOM(elt type, elt attrs, showInPresentation = true, ???)` - immediately create and return a reference to a DOM element (dangerously mutating the DOM outside of the react flow-- i *think* we can get away with this, b/c muting DOM within MD elts seem to work ok, even when you trigger re-renders of parent elts by e.g. moving a cell up/down). This would append a new child to a div that we would add to each cell (between the input and output rows), and which could be selected by e.g. `document.body.querySelector("div.cell.selected div.printarea")` or whatever.

`io.print(obj, showInPresentation = false)` - extension to `io.DOM` that would print `obj` into, let's say, a `<pre>` container? note that i don't favor globally overriding or somehow capturing `console.log`. i believe we should whenever possible avoid changing _anything_ about the behavior of regular browser APIs unless we have no choice, and this is a perfectly legible alternative.

`io.show(obj, showInPresentation = false)` - extension to `io.DOM` that would pretty print `obj` using any registered output renderers.

`io.script(url)`; `io.stylesheet(url)` -- request a script or stylesheet and do whatever we currently do (just append a <script> or <link css> tag in <head>?). we could also offer something like `x = io.script(url, namespace=true)` that returns an object containing everything that is dumped into the window object when you load a script via script tag. i'm wary of using the term "require" because i think it implies dependency resolution or something (but i'm not sure about that)? or maybe `io.require` would be used for files wrapped in module syntax? i'm really not sure how this stuff works.

`io.getFile(url, timeOut = 5000ms)` or `io.awaitFile(url, timeOut = 5000ms)` -- synchronously load and return data object. flags for parsing csv, json, ... ?

[...]

Issue #166, opened 2017-11-08
come up with app error reporting mechanism


hamilton:
We report cell execution errors in the output of the cell, which is the right spot. But we do not have a way to report any other app errors. JSON parse errors when importing notebooks, for instance. There will likely be other ones as well.

There are a few options. I'm in favor of using the [Material UI Snackbar](http://www.material-ui.com/#/components/snackbar) component, which pops up a small, unobtrusive component that clicks away.

hereisnaman:
@hamilton @bcolloran I was working on solving saving unnamed notebook issue https://github.com/iodide-project/iodide/issues/208. As @bcolloran said there in the thread, it would be good to show a snackbar in case of explicit saves. I can build a general use component for it, which will integrate with both jsmd execution errors and other app errors. 

I need to know that how will we show and scroll to the relevant output in the history window when clicked on trigger(click on snackbar or auto), inpage link ? Please suggest.

bcolloran:
we have this now!!

Issue #38, opened 2017-09-12
error objects should be rendered by the output renderer when returned by an "eval" call


bcolloran:
i attempted to put in a tiny bit of error help with a simple try/catch that writes some error info to a string, but maybe the lib used to display other outputs can be extended to handle error objects.

PR #343, opened 2018-02-14
Output handler api

mdboom:
This is a replacement for #244, since Github asked me to move my repository to be public.

This follows the stdlib proposal in #288.

Also supports regular DOM Elements as suggested by @hamilton.  React components would be nice, but thought it might be worth discussing the best way to provide React to third-party code -- as a member in `io.` maybe?

### DIFF
--- a/examples/output-handling.jsmd
+++ b/examples/output-handling.jsmd
@@ -86,7 +86,7 @@ methods:
    can be an HTML string, or DOM Element.
 
 You then tell Iodide about your custom output handler by passing it to
-`io.addOutputHandler`.
+`iodide.addOutputHandler`.
 
 If you find yourself using the same output handler in a lot of notebooks, this
 is a great candidate for something to put in an external library.
@@ -107,7 +107,7 @@ const GeoLocationOutputHandler = {
     }
 }
 
-io.addOutputHandler(GeoLocationOutputHandler)
+iodide.addOutputHandler(GeoLocationOutputHandler)
 
 %% md
 Now that we've defined the custom output handler, it will be used to display any
--- a/examples/output-handling.jsmd
+++ b/examples/output-handling.jsmd
@@ -41,9 +41,20 @@ for (let i=0; i<10; i++){range.push(i)}
 A = range.map( (x,i) => range.map( (y,j) => (Math.random()+i-.5)))
 
 %% md
+## Objects with a render method
+
 Objects can define their own custom output by adding a `render` method.
 
-In the example below, a custom class, `GeoLocation`, is used to represent a geographic location by latitude and longitude.  When rendering the output, the Google Maps Static Image API is used to display a  thumbnail map for the location.
+This method may return any of the following:
+
+- A string containing HTML
+
+- A [DOM Element](https://developer.mozilla.org/en-US/docs/Web/API/Element)
+
+In the example below, a custom class, `GeoLocation`, is used to represent a
+geographic location by latitude and longitude. When rendering the output, the
+Google Maps Static Image API is used to display a thumbnail map for the
+location.
 
 %% js
 var GeoLocation = class GeoLocation {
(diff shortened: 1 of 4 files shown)


src/api.js
```diff
@@ -0,0 +1,10 @@
+// The "Public API" for notebooks. This lets notebooks and third-party plugins
+// extend and manipulate the notebook
+
+import { addOutputHandler } from './components/output'
+
+export const io = {
```
hamilton: @mdboom would love to get your thoughts on this. I am thinking it might be wise to avoid the `io` namespace because [socket.IO](https://socket.io/) typically uses it. It's not a _default_ namespacing thing, but given it is common, do you think we should use something else? I guess this is a matter of taste.

mdboom: I'm not plugged into the Javascript ecosystem enough to really know...  Happy to make this whatever we decide.

examples/output-handling.jsmd
```diff
@@ -41,21 +41,81 @@ for (let i=0; i<10; i++){range.push(i)}
...
+io.addOutputHandler(GeoLocationOutputHandler)
+
+%% md
+Now that we've defined the custom output handler, it will be used to display any
+object with `lat` and `lon` members, even though it's not an instance of a
+specialized class.
+
+%%js
+var x = {lat: 37.387315, lon: -122.060009}
+x
```
hamilton: Idea: do you think we need a story for being able to coerce the output handler to a specific type? I might have this neat geo output handler, for instance, but in my workflow I may want to just check my lat/lon occasionally, so it might be nice to to have a stdlib fcn that effectively says "send this to the output but with this handler". This could probably get kind of tricky when we think about when the output handler state is saved.

hamilton: Similar use case - I have an array of lat/lon objects, which might get consumed by a custom output handler to plot them on all on a map, but I may also need to check the lat/lon in a table as well. being able to just say "set the output to table" would be helpful.

I don't know if this is in-scope for your PR, but just wanted to bring it up.

examples/output-handling.jsmd
```diff
@@ -41,21 +41,81 @@ for (let i=0; i<10; i++){range.push(i)}
...
+   can be an HTML string, or DOM Element.
+
+You then tell Iodide about your custom output handler by passing it to
+`io.addOutputHandler`.
+
+If you find yourself using the same output handler in a lot of notebooks, this
+is a great candidate for something to put in an external library.
+
+%% js
+var React = require('React')
```
hamilton: this breaks the cell at the moment

examples/output-handling.jsmd
mdboom: That's an interesting idea.  One could imagine a function like:
```
io.displayOutput(value, 'handler_name')
```
to force a particular output handler.

I have also wondered if it would be helpful to have a little dropdown to the left of the output cell that lets you choose an output handler (if the heuristic guess is wrong).

examples/output-handling.jsmd
mdboom: Oops.  Was playing around and didn't mean to leave that in there.

examples/output-handling.jsmd
hamilton: Feels like both are probably necessary!

hamilton:
@mdboom I think this is fantastic! Made test to play with the idea a bit. Works great:

feb-14-2018 08-45-36

Two non-blocking questions:

- what should our API ns be? I'm still a bit spooked by `io` because I have personally used socket.io extensively in the past, as have many JS programmers, and though the namespacing from socket.io is not explicit, many programmers fall back on it. Should we go with `IO` instead to be safe? Or just ignore this entirely?
- For those added via `io.addOutputHandler`, should there be a way to coerce an output handler to be use on the input-side? In my case, maybe I should just expect to return  `new TestHandler(obj)` or whatever and let that be the point of coercion, and not the automatic part, if that's what I want. We've talked about also allowing a user to select a different output handler in a dropdown as well, so perhaps this is redundant. If we go this route, we should make it easy to do this with the built-in output handlers already part of the magic as well (tables, matrices, etc.). My comment above suggests a use case for this. Would love to get @wlach's thoughts on this, since I think both of us have a Metrics Graphics output handler kicking in our heads.

mdboom:
@hamilton: I think I have the same questions you have, without answers, so I'll let others weigh in.

mdboom:
Also, @hamilton, we should put the demo in the .gif above in an example notebook.  It would be good to have a real CSS-y thing.

hamilton:
Agreed, but I should make it about 10% more disco first

src/components/output.jsx
```diff
@@ -151,11 +160,26 @@ export default class CellOutput extends React.Component {
...
 
     for (const handler of handlers) {
-      if (handler.shouldHandle(value)) {
-        const resultElem = handler.render(value)
-        if (resultElem !== undefined) {
+      let shouldHandle = false
+      try {
+        shouldHandle = handler.shouldHandle(value)
+      } catch (error) {
+        console.log('An output handler threw an exception in shouldHandle')
```
wlach: We should probably propagate this error up somehow... ditto for the error below.

mdboom: I thought about that, but it's sort of a "soft error".  If one of the output handlers borks, it should gracefully fall back to the generic handlers that should hopefully do a decent job of displaying just about anything.  I think that's better than an exception.  I guess the best thing would be to get the generic output *plus* a little inline message above it saying that an output handler failed.  That sort of goes back to the broader question of how to display errors (#166, #38)

src/api.js
bcolloran: i'm in favor of just using this for now so that we can start experimenting, but not getting attached to `io` and being open to change it after thinking about it and discussing it some. see #350

src/components/output.jsx
```diff
@@ -151,11 +160,26 @@ export default class CellOutput extends React.Component {
...
+      }
+      if (shouldHandle) {
+        var resultElem = handler.render(value)
```
bcolloran: linter nit: `let` not `var` :-)

i don't know why, but this lint error does show up in my editor but doesn't seem to alert in mt build terminal (I'm guess you had the same experience). Maybe the linter is misconfigured for `npm run start`.

mdboom: Yeah, I see that, too.  Created #353 to track that.

mdboom:
I think I agree with @wlach here about using the namespace `iodide`.  I've implemented that change, but not opposed to changing again if we all don't agree.

bcolloran:
this is great, excited to dogfood it!

PR #344, opened 2018-02-14
Recursively call output handlers

mdboom:
This is a follow-on to #340, that, rather than special casing types within the compound output handlers (data frame, array, etc.), it just calls output handlers recursively on each value.

It lets you do some pretty cool things, like this:

screenshot2018-02-14-14 07 23

Of course, this may be too magical or more trouble than it's worth, but I thought it worth sharing for discussion.

### DIFF
--- a/examples/output-handling.jsmd
+++ b/examples/output-handling.jsmd
@@ -99,7 +99,7 @@ const GeoLocationOutputHandler = {
 
     render: function(value) {
         // return a HTMLElement, as created by document.createElement
-        var img = document.createElement('img')
+        let img = document.createElement('img')
         img.src = 'https://maps.googleapis.com/maps/api/staticmap?center=' + value.lat + ',' + value.lon + '&zoom=17&size=600x300&maptype=roadmap'
         return img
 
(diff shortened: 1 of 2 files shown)


hamilton:
I don't have much time today to review this idea more extensively, but I think it seems like a pretty good idea, and one I had when I was making the bug fix in #340. Specifically, I think `Date` objects should have their own output formatters to make them just a bit easier to work with / inspect certain parts of the object, and the `Object` format for the `ReactTable` handler really should just give you the tree view we have for regular object outputs.

One downside would be if you have a registered apiHandler for, say `{lat:..., lon:...}` but for the table you really just want the tree view to inspect the object.

@bcolloran what do you think?

hamilton:
One way to handle the issue of having registered handlers is we can include a `useInDataTable` option which is set to `false` by default.

mdboom:
The `useInDataTable` idea is a good one, but maybe it could be generalized to all containers, such as 1-D arrays.

I say this knowing very little about CSS, but I wonder if ideas from adaptive layouts would be useful here.  i.e., if the container that the object is being rendered to is small it's rendered a certain way, and large a different way.

bcolloran:
I think this is the right appcroah @mdboom. Hamilton and I were talking, and we think your idea sounds right -- just add a flag for all "renderRecursively" contexts that defaults to false but can be switched on if desired. In the case of the handlers for built-in types (see #360), we could set them to render in recursive contexts by default.

mdboom:
I've updated this based on the ideas here with just one tweak:

I added a flag `inContainer` (rather than `renderRecursively`), which I think is more descriptive.  It is true when the value is being rendered inside a container.  It can choose to either (a) not render, and pass along to another handler, or (b) render in an alternate smaller form.  The whole concept can be ignored by handlers, in which case the rendering is always the same.

I added a modified `GeoLocation` example which renders a smaller map when it's inside a container vs. outside one.

bcolloran:
Sounds great, `inContainer` works for me.

Mike, I'd be happy to go ahead and merge this now if you like, but i'm not sure where you're at now that you found the devtools stuff in #360 -- do you want to go with this PR, or are you now thinking of taking another pass at the idea using that stuff?

Or I guess to put it differently: I think this looks good, please merge it when you feel ready.

PR #347, opened 2018-02-14
fixes the row count issue with tables < 25 rows

hamilton:
There are some weird formatting issues if your data table gets tweaked (for instance from using `.slice`). the output handler keeps the previous number of rows from the previous object that was fed into the component.

This is a pretty easy fix - just add `minRows={0}` and `ReactTable` knows exactly what to do. Demo below:

feb-14-2018 14-22-39


hamilton:
Thanks! Going to merge now.

Issue #363, opened 2018-02-16
[META] refactor towards 3 row JS cells


bcolloran:
Here are a bunch of small steps. Each of the subtasks below should be a small PR that would make a good first issue.

If you want to work on one of the subtasks, feel free to say so in a comment below. please claim just one at a time until you have a filed a PR; once you have completed that PR, you can claim another one.

## 1: refactor cells

Objective: refactor all cell types to render their `CellContainer`s and `CellRow`s directly (instead of using `oneRowCell` and `twoRowCell`). 

These subtasks can be done in parallel by multiple contributors.

Each of these changes should be filed as separate PRs; the PR's will only be accepted if they include updated tests for the *unconnected* component's complete contract. For more on testing an unconnected component's contract, look at the existing tests and see the following links
- https://hackernoon.com/unit-testing-redux-connected-components-692fa3c4441c and
- https://medium.com/wehkamp-techblog/unit-testing-your-react-application-with-jest-and-enzyme-81c5545cee45

Integration tests for actions are _not_ required with these PRs.

- [x] cell-type-css -- remove OneRowCell
- [x] cell-type-raw -- remove OneRowCell
- [x] cell-type-markdown -- remove OneRowCell
- [x] cell-type-external-resource -- remove twoRowCell
- [x] cell-type-javascript -- remove twoRowCell

## 2:  remove  `oneRowCell` and `twoRowCell`

Once section (1) is done, these subtasks can be done in parallel by multiple contributors.

Objective: once each of the cell components has been refactored, we can remove the `oneRowCell` and `twoRowCell` components (these aren't really needed anymore, they were needed for managing "pill" representation of cells in presentation mode, which is gone).

- [x] completely remove twoRowCell from codebase #357 
- [x] completely remove oneRowCell from codebase #356 

## 3
These tasks still need a bit more thought. Please wait before working on these sub-issues
- [x] refactor state and state prototypes to handle different (arbitrary?) numbers of rows. Right now, every cell is assumed to have two rows (even tho this is not actually the case)
- [x] add a middle cell row to JS cell

i think those are all the steps. only the first part of step (3) is likely to be hairy.

pradeepgangwar:
Can #357 and #356 be solved independently. I see many thing depend on them.

bcolloran:
@pradeepgangwar i updated the description of the issue -- i think it should be possible to take it one step at a time and proceed towards #357 and #356

djbarnwal:
[...]

PR #364, opened 2018-02-17
Remove twoRowCell from cell-type-javascript

djbarnwal:
This PR removes `twoRowCell ` from `cell-type-javascript` as mentioned in Issue #363 . Also, the corresponding test file has been updated accordingly. I haven't written a lot of test code in Enzyme so any review to make this code better would be appreciated :)


bcolloran:
@djbarnwal the tests look perfecr to me! I don't see any way to make that code better, great job and thanks for the PR! :-)

PR #366, opened 2018-02-17
Correct prop types for PageMode

pradeepgangwar:
screenshot from 2018-02-18 01-02-40

CellContainer, CellRow and CellEditor have `title-edit` option missing from PageMode prop.


bcolloran:
good catch, thank you!

PR #379, opened 2018-02-20
Split notebook-header components

### LINKED ISSUE #376: Break up notebook header in separate components
right now, notebook-header.jsx contains I think about four top-level react components; it should really be split up so that each top-level component is in its own file.

Initially, a PR that just splits these up will be accepted, but it would be possible to go above and beyond by:
* making sure that each new component only has access to the minimal amount of the state it needs (by removing unneeded variable in each components `mapStateToProps` function). This is basically issue #176
* adding unit tests for each new component. For now, just testing the unconnected component would be a great step (since we still haven't figured out our strategy for testing connected component actions, see #375 ).

### LINKED ISSUE #380: deprecate h1 presentation-title from notebook-header
Presentation view has this explicit h1 added according to how your name your file. Title elements should be explicitly added by the user, so we should deprecate this.

pradeepgangwar:
Fixes #376 
Fixes #380

### DIFF
--- a/src/components/editor-menu.jsx
+++ b/src/components/editor-menu.jsx
@@ -1,0 +1,84 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { Toolbar, ToolbarGroup, ToolbarTitle } from 'material-ui/Toolbar'
+
+import IconButton from 'material-ui/IconButton'
+import HistoryIcon from 'material-ui/svg-icons/action/history'
+import ArrowDropDown from 'material-ui/svg-icons/navigation/arrow-drop-down'
+
+import darkBaseTheme from 'material-ui/styles/baseThemes/darkBaseTheme'
+import getMuiTheme from 'material-ui/styles/getMuiTheme'
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider'
+
+import MainMenu from './menu-component'
+import ViewModeToggleButton from './view-mode-toggle-button'
+
+import Title from './title'
+
+import { prettyDate } from '../notebook-utils'
+
+
+class EditorMenu extends React.Component {
+  static propTypes = {
+    viewMode: PropTypes.oneOf(['editor', 'presentation']),
+    sidePaneMode: PropTypes.oneOf(['history', 'declared variables']),
+    mode: PropTypes.oneOf(['title-edit', 'command', 'edit']),
+    actions: PropTypes.shape({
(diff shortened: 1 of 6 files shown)


src/components/editor-menu.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+import MainMenu from './menu-component'
+import ViewModeToggleButton from './view-mode-toggle-button'
+
+import Title from './title'
+
+import { prettyDate } from '../notebook-utils'
+
+
+class EditorMenu extends React.Component {
```
hamilton: Since we are breaking these components out into separate files, this seems like a good time to rename this `EditModeToolbar`, since that's more descriptive. I would also rename this file `edit-mode-toolbar.jsx`.

src/components/presentation-menu.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import lightBaseTheme from 'material-ui/styles/baseThemes/lightBaseTheme'
+import getMuiTheme from 'material-ui/styles/getMuiTheme'
+import MuiThemeProvider from 'material-ui/styles/MuiThemeProvider'
+
+import ViewModeToggleButton from './view-mode-toggle-button'
+
+class PresentationMenu extends React.Component {
```
hamilton: We should rename this to `PresentationModeToolbar` for now. (It's really just a single button, so not really a toolbar.) The name of the file can be `presentation-mode-toolbar.jsx`.

src/components/presentation-menu.jsx
```diff
@@ -0,0 +1,29 @@
...
+  }
+  render() {
+    return (
+      <div className="presentation-header">
+        <div className="view-mode-toggle-from-presentation">
+          <MuiThemeProvider muiTheme={getMuiTheme(lightBaseTheme)}>
+            <ViewModeToggleButton textColor="black" actions={this.props.actions} viewMode={this.props.viewMode} />
+          </MuiThemeProvider>
+        </div>
+        <h1 className="presentation-title" style={{ color: this.props.title === undefined ? 'gray' : 'black' }}>{this.props.title || 'new notebook'}</h1>
```
hamilton: We could get rid of this `h1` as per #380 :)

pradeepgangwar:
@hamilton Fixed code based on review. Please review again.

hamilton:
Looks good - thanks @pradeepgangwar!

Issue #176, opened 2017-12-05
compartmentalize state in NotebookHeader


bcolloran:
I think I've figure out out why the perf has gotten so bad again, even on my branch with a lot components farther down the tree. it boils down to:
* NotebookHeader is still connected to a large part of the store, which means that it ends up re-rendering on every key press.
* I think i started working on this perf refactor before we changed to the material menu -- at that time, the thing that was costing us the most upon was having to re-render all of the codemiror instances. b/c our menu was pretty light, i didn't look in depth at what it was connected to, and it didn't matter that much that it re-rendered every time. but now that it's heavier, it surfaces the cost of rerendering each time.

This should be easy to resolve, we just need to go into all of the menu components and make sure they are attached to the most absolutely minimal part of 'state' possible. For example, i can't spot a reason that we need to pass a copy of `state.cells` into the menu at all; it looks like we are just using that to look up the currently selected cell id, so it's way overkill, but that part of state changes every time a key is pressed in the editor...

I'll see if i can make some headway on this tonight, but i might need your help @hamilton since i'm not as familiar with how the menus work at this point.

Let's see if we can make some progress on this on tuesday.

Issue #165, opened 2017-11-08
fuzzy text menu search a la figma


hamilton:
see #150. We will be adding more menu options in the coming months, and it will be nice to have something similar to what figma has:

![figma](https://media.giphy.com/media/xT9IgHl36aTKrbtVrG/giphy.gif)

bcolloran:
i'm going to close this for now -- we bth know that this option exists if we ever actually need it, but for now (and for the last year+) we really haven't. so wontfix until the need exists

Issue #152, opened 2017-10-19
transform keybindings file to be an object


hamilton:
The keybindings should be an importable object, whose events are shared with other parts of the system, such as the menu.

PR #385, opened 2018-02-21
create new NotebookTask object to combine keybindings and menu items into a single set of user tasks

hamilton:
This is the beginning of a major refactor of how the keybindings and menus work. The premise is to have all the user actions ("tasks" for now, since "actions" is a Redux term and we don't want to muddy the waters) unified in one simple sort-of-declarative place. This will allow us to do cool things like add a fuzzy text search through all possible actions (those with & without keybindings), provide automatically generated documentation of various tasks and how to get to them, etc.

This required making a new object, the `NotebookTask`, along with many tests. These are really just fancy JS objects that have custom getters, for now. They also, however, validate inputs (which is the focus of most of the tests written for them).

Ultimately I aim to move toward fixing #176, #165, #152, and a few others I'll tag in a second once I look them up.

The order of operations is this:

- __this PR__ create `src/notebook-task.js` + `test/notebook-task.test.js`
- __this PR__ create `src/task-definitions.js`, which defines ALL the tasks
- __this PR__ replace the old jupyter keybindings with `src/task-definitions.js`
- __ANOTHER PR__ write tests for `src/task-definitions.js`
- __ANOTHER PR__ replace `menu-content.js` with `task-definitions.js`
- __ANOTHER PR__ create a `menu-item.jsx` component that wraps material-ui stuff but is generic enough that we can swap out the content for something else
- __ANOTHER PR__ menu-component / MainMenu becomes just a `<menu><item/>...</menu>` 

As of this PR, the first three above are done. We can tackle the rest in another PR.

Comments and criticism welcome! I'm sure I didn't do something very elegantly somewhere and I'd rather it be good now than good later.


src/task-definitions.js
```diff
@@ -0,0 +1,347 @@
...
+  },
+  keybindingCallback() {
+    if (isCommandMode()) dispatcher.changeCellType('css')
+  },
+})
+
+TASKS.EditMode = new NotebookTask({
+  title: 'Change to Edit Mode',
+  keybindings: ['enter', 'return'],
+  displayKeybinding: '⎋',
```
bcolloran: i don't recognize this symbol: '⎋'. Can we just use the word `enter` written out? i think it will be a little clearer to more people.

In Jupyter's keybd shortcut display (and other places), they wrap all the shorcut in little box. i think it's very clear and looks fine:

image

src/task-definitions.js
```diff
@@ -0,0 +1,347 @@
...
+      dispatcher.insertCell('javascript', 'below')
+      dispatcher.selectCell(getCellBelowSelectedId(), true)
+    }
+  },
+})
+
+TASKS.DeleteCell = new NotebookTask({
+  title: 'Delete Cell',
+  keybindings: ['shift+del', 'shift+backspace'],
+  displayKeybinding: '\u21E7 \u232b',
```
bcolloran: if we are going to use unicode, probably we should just use unicode literals rather than codepoints -- but same thought here as for "enter" below-- might be better to use text; i think it's probably clearer across platforms etc

src/task-definitions.js
```diff
@@ -0,0 +1,347 @@
...
+  keybindings: ['up'],
+
+  keybindingCallback(e) {
+    if (isCommandMode()) {
+      // e.preventDefault blocks kbd scrolling of entire window
+      if (e.preventDefault) {
+        e.preventDefault()
+      } else { // internet explorer
+        e.returnValue = false
+      }
```
bcolloran: this pattern:
```
      if (e.preventDefault) {
        e.preventDefault()
      } 
```
and this pattern:
```
   if (e.preventDefault) {
+        e.preventDefault()
+      } else { // internet explorer
+        e.returnValue = false
+      }
```

are showing up a lot. I should have cleaned it up when i last worked on keybindings-jupyter, but i think that (A) the latter can always safely replace the former (since apparently its an IE workaround) and (B) we can wrap the latter in a little function like:
```
function preventDefaultWrapper(e) {
    if (e.preventDefault) {
      e.preventDefault()
    } else {
      e.returnValue = false
   }
}
```

and just replace those 3 or 5 lines of cruft with a one liner `preventDefaultWrapper(e)` throughout the file. (If you want to go with this, you can of course call the function whatever you want

bcolloran:
just brainstorming some more re "task"--

sublime text calls its `ctrl+shift+p` thing the "command palette" -- we could use "command", but we already have a "command mode" (but the command mode is maybe the number one place you'd issue commands, so maybe that term is not too bad/would not feel too overloaded?)

there is "event". or since `explicit > implict` ==> "uiEvent" or "userEvent" or something;

perhaps "launcher", which kind of connote something that launches one or more redux actions?

and ftr, i can live with with "task" if that ends up sounding best to you :-)

src/task-definitions.js
```diff
@@ -0,0 +1,347 @@
...
+        dispatcher.addCell('javascript')
+        dispatcher.selectCell(getCellBelowSelectedId(), true)
+      }
+    }
+  },
+})
+
+TASKS.MoveCellUp = new NotebookTask({
+  title: 'Move Cell Up',
+  displayKeybinding: '\u21E7 \u2191',
```
mdboom: Rather than having `displayKeybinding` and `keybindings`, which could accidentally become out of sync, would it make sense to have a function that converts the "ascii" `keybindings` value into Unicode automatically (much as is already done for the Macintosh ⌘ symbol in `commandKey` above?

Also, should we have a preference for actual codepoints rather than `\u` escaped ones here?  I've always felt that `\u` escapes are only desirable if you have to use editors or tools that don't handle unicode well, but this is 2018... :)

src/task-definitions.js
```diff
@@ -0,0 +1,347 @@
...
+  displayKeybinding: 'a',
+
+  callback() {
+    dispatcher.insertCell('javascript', 'above')
+    dispatcher.selectCell(getCellAboveSelectedId(), true)
+  },
+  keybindingCallback() {
+    if (isCommandMode()) {
+      dispatcher.insertCell('javascript', 'above')
+      dispatcher.selectCell(getCellAboveSelectedId(), true)
```
mdboom: Maybe replace with `this.callback()` to avoid code duplication here?  (I see above from some other tasks why both `callback`/`keybindingCallback` have to exist, unfortunately...)

src/keybindings.js
```diff
@@ -1,17 +1,16 @@
 import Mousetrap from 'mousetrap'
-import jupyterKeybindings from './keybindings-jupyter'
+import TASKS from './task-definitions'
```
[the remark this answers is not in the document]
hamilton: I'm happy to bring this down to just `tasks` - I don't have a strong feeling one way or another on this.

src/task-definitions.js
```diff
@@ -0,0 +1,347 @@
...
+
+  callback() {
+    dispatcher.changeCellType('css')
+  },
+  keybindingCallback() {
+    if (isCommandMode()) dispatcher.changeCellType('css')
+  },
+})
+
+TASKS.EditMode = new NotebookTask({
```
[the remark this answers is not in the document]
hamilton: Very good call - these should remain verbs!

src/task-definitions.js
```diff
@@ -0,0 +1,347 @@
...
+  let ctr = 'Ctrl '
+  if (OSName === 'MacOS') {
+    ctr = '⌘ '
+  }
+  return ctr + key
+}
+
+const TASKS = {}
+
+TASKS.EvaluateCell = new NotebookTask({
```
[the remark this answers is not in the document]
hamilton: Yes good call - these should be camel, not pascal.

src/task-definitions.js
```diff
@@ -0,0 +1,347 @@
...
+    dispatcher.evaluateCell()
+  },
+})
+
+TASKS.EvaluateAndSelectBelow = new NotebookTask({
+  title: 'Evaluate and Select Below',
+  keybindings: ['shift+enter'],
+
+  keybindingCallback() {
+    if (viewModeIsEditor()) {
```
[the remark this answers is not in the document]
hamilton: I completely agree. I really don't like that we're using both `callback` and `keybindingCallback`. These should be unified, since for the most part these are duplicates of each other w/ a couple of additional checks for `keybindingCallback`. This would be MUCH cleaner without two callbacks.

src/task-definitions.js
hamilton: This is a mistake - that's technically the `esc` key not the `return` key, but honestly `Esc` explains everything and this is evidence enough that the char for it is not supported widely.

src/task-definitions.js
hamilton: Yep, these are holdovers copy-pasted from the menus. Much of this depends on how they are rendered in the `material-ui` components. @mdboom's comment below about acutally just having a fcn that makes a display keybinding is smart (I had that thought while doing this but figured I'd keep them separate for now since I wasn't sure how to deal with multiple keybindings). I think we should have a function that makes the display keybindings from this, and that allows us to deal with formatting stuff without having to be so explicit.

src/task-definitions.js
hamilton: YES. My comment above re: two callbacks is relevant here.

src/task-definitions.js
hamilton: 100%.

src/task-definitions.js
hamilton: Yeah, I really want to figure out how to get rid of two callbacks. I HATE THAT WE HAVE TWO CALLBACKS and I won't sleep well until there is only one per `NotebookTask`.

hamilton:
Thanks for the first batch of comments y'all! I've got a couple spare hours later today and I'm going to address all of them and ask y'all for another review.

I think it's important we get this right, since effectively I'm asking for about half the app to be written around these tasks (menus, keybindings).

src/task-definitions.js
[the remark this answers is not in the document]
hamilton: Will have to think a bit more about how this should be best-approached.

src/task-definitions.js
bcolloran: i actually think the two callbacks is a nice explicit pattern in cases where it's need -- i think mdboom just means that in this case it could be

```
keybindingCallback() {
    if (isCommandMode()) {
      this.callback()
    }
}
```

but i guess if we have a `NotebookCommandModeTask` wrapper class, it could by default just use `callback()` wrapped in `isCommandMode()` unless an expiicit `kebindingCallback` is provided. so maybe we can avoid two callbacks in many/most cases with wrapper classes. But i think having the flexibility and clarity of `keybingCallback` is great. I really like the approach you've come up with for this whole refactor :-)

mdboom: Yes, that's what I meant @bcolloran, but you said it better. :)

hamilton: Yeah, I suppose my issue with the two callbacks is almost always, with `myTask.keybindingCallback` what we really want is `callback` with an explicit `keybindingPrecondition`. The getter for `keybindingCallback` can remain the same API-wise, but what I want is just to say is

```javascript
const newTask = new NotebookTask({
   title: 'whatever',
   callback: ()=>{fireSomeAction()},
   keybinding:...,
   keybindingPrecondition: isCommandMode, // or some other function
})

```

And the API is such that `newTask.keybindingCallback` returns an anonymous function that looks like this:

```javascript
return function() {
   if (this.keybindingPrecondition(...arguments)) this.callback(...arguments);
}

```


These are small nits, and are not probably worth stopping the show for it, and at any rate your little code snippet actually is nice enough and reduces the complexity by ensuring things aren't out of sync. But this seems just about as explicit and feels a bit cleaner to me.

src/task-definitions.js
hamilton: iirc the material ui components don't play nicely with codepoints, which is why I went with these.

src/task-definitions.js
mdboom: @hamilton: I like what you propose as well.  I think it goes even further toward elegance.  But I'm happy to say "save for another PR" to avoid holding up this one.

mdboom:
One other thought I had.  Long term, once we're doing things like i18n, we'll probably want a way to change keybindings.  This ties the keybindings themselves to the actions they perform pretty tightly.  Probably fine for now, but want to ensure we're seeing around that corner.

hamilton:
@mdboom @bcolloran I addressed issues that y'all brought up. Just tested all the keybindings manually and everything seems to work as expected. Let me know how y'all would like to proceed with this!

hamilton:
@bcolloran @mdboom I went ahead and implemented the `preventDefaultKeybinding` option, as well as `keybindingPrecondition` for now. If I declare `callback` with either (or both) of these options, and I don't  declare an explicit `keybindingCallback`, the value `myTask.keybindingCallback` will be constructed with all of these. I prefer this design for some of these keybinding callbacks because it makes `task-definitions.js` feel a bit more declarative. You know exactly how we compose each one of these tasks, and you don't have to look up another class definition elsewhere either.

I'm going to merge this in a few hours.

bcolloran:
looks good to me!

PR #387, opened 2018-02-21
disable conflicting ctrl-d sublime shortcut in code editor

### LINKED ISSUE #159: disable the "ctrl-d" multiple cursors shortcut within codemirror code editors
ctrl-d for declared variables pane conflicts with codemirror sublime multiple cursors shortcut

fkohlgrueber:
This is my first PR in this project and I'd greatly appreciate feedback of any form. 

I fixed #159 by defining `ctrl-D` as a shortcut that executes an empty function. This feels a bit hacky to me, but I didn't find a better solution. Simply not defining the shortcut within the `sublime` layout doesn't work because it's also defined in the `default` layout (which is used as a fallthrough by the `sublime` layout).

I'm looking forward to your feedback. Thanks in advance.

### DIFF
--- a/src/codemirror-keymap-sublime.js
+++ b/src/codemirror-keymap-sublime.js
@@ -136,7 +136,7 @@
   }
 
 
-// these shorcuts conflict with notebook shortcuts
+// these shortcuts conflict with notebook shortcuts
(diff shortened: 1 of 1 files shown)


bcolloran:
works for me -- we can live with a solution like this, dealing with the codemirror internals can be a headache. :-)

PR #391, opened 2018-02-22
FOR DEMO ONLY: Use react-inspector instead of JSONTree

mdboom:
With confirmation that Firefox devtools' [devtools-reps](https://github.com/devtools-html/devtools-core/tree/master/packages/devtools-reps) isn't really meant to be used standalone, I did a little Googling for similar, and found [react-inspector](https://github.com/xyc/react-inspector).  It claims to be "the power of [Chromium's] DevTools inspectors right inside your React app", but it doesn't look like it's actually based on the Chromium DevTools code, just inspired by it.  It doesn't do a lot of the nice things that `devtools-reps` does (like the different length reps) or have as much coverage of object types (such as `Error` objects and tracebacks).  It doesn't have the extensible architecture of Chromium DevTools, our thing, or that Firefox devtools apparently has planned either.

tl;dr: I don't think this really advances the ball, but I'm posting this here in case in playing with it others may see an advantage to this that I don't.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5357,6 +5357,11 @@
         }
       }
     },
+    "is-dom": {
+      "version": "1.0.9",
+      "resolved": "https://registry.npmjs.org/is-dom/-/is-dom-1.0.9.tgz",
+      "integrity": "sha1-SDgy1SlyBz3hK5/j9gMghw2oNw0="
+    },
     "is-dotfile": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/is-dotfile/-/is-dotfile-1.0.3.tgz",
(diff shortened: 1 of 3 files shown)


hamilton:
Have been testing out your PR a little. Here's what I've found:

1. it's nice to have something that looks / works like a console output. I think that gets this closer to feeling like something familiar.
2. The `ObjectInspector` is really slow with new clicks, and I think this makes it feel very clunky.
3. I DO think that the `DOMInspector` is not clunky and very very cool, and would actually be QUITE useful for people for inspecting the details of some DOM element in an easy way. Something to consider stealing from this lib.
4. this little anim demonstrates all of these issues succinctly:

dom element useful

mdboom:
No reason we couldn't just use the DOMInspector for DOM elements and continue using JSONTree for JS objects.

mdboom:
So am I right that the thing to do is: Use `react-inspector` for DOM elements, but continue to use `JSONTree` for Javascript elements?

bcolloran:
sounds reasonable to me

hamilton:
update from team meeting: we're going to wait a few days to see if we can use the native dev tools code instead of any of this.

PR #398, opened 2018-02-23
392 fix

### LINKED ISSUE #392: ctrl-d should just delete the character under the cursor, not popup declared variables pane
Probably based on 20+ years of using emacs, using CTRL-D to delete the current character under the cursor is pretty hard-embedded in my muscle memory. Virtually every text editing environment I use (including Atom and Firefox's default text entry box) also supports it.

I know there's been a bunch of churn on this (#159, #387) which I hate to revert, but this is a huge frustration point whenever I use the environment (sorry for not filing the bug until now). I think we just need to find another shortcut for the "declared variables" pane. For all intents and purposes I think this keybinding is "reserved"-- it's no doubt a minority of people that use it, but those who do (like me) *really* care about it.

bcolloran:
this will fix #392 ... kind of...

@wlach , it turns out that the default ctrl-d keybinding in sublime text is to select other instances of the word under the  cursor. i'm not sure if we should single out ctrl-d as the one exception to the sublime keymap, or maybe just leave it until we can put in keymap selection?

### DIFF
--- a/src/codemirror-keymap-sublime.js
+++ b/src/codemirror-keymap-sublime.js
@@ -136,7 +136,7 @@
   }
 
 
-// these shortcuts conflict with notebook shortcuts
+// these shorcuts conflict with notebook shortcuts
   // cmds[map[ctrl + "Enter"] = "insertLineAfter"] = function(cm) { return insertLine(cm, false); };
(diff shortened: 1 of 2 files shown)


wlach:
@bcolloran yeah I think we shouldn't mess with the defaults. I'd say let's land this as-is and figure out keybindings later (I know there was talk of switching to a different code editing component, which might change the defaults/requirements).

PR #399, opened 2018-02-23
Python  🐍 🐍 🐍 🐍 🐍 🐍

mdboom:
This is a quick-and-dirty hack to add Python cells, so we can play with the interesting work going on in the pyodide respository.  Eventually, this should be replaced with an API that allows third-party libraries to add support for new languages, rather than hardcoding the notion of python all over the place like this.

Anyway, this is turning out to be really cool.  Kick the tires and let me know what you think.  We should probably file any issues related to the Python stuff in the pyodide repository.

### DIFF
--- a/examples/python.jsmd
+++ b/examples/python.jsmd
@@ -36,16 +36,46 @@ The basic data types (None, bool, ints, floats, lists, and dicts) are converted 
 [0, 1, 32.0, 'foo', {'a': 10, b: '20'}, b'bytes']
 
 %% md
-All other data types (such as class instances, functions, File objects etc.) are returned to Javascript as "opaque raw pointers".  They are there on the Javascript side, but you can't really do anything useful with them other than pass them around.
+## Custom data types
 
-(I have some ideas about how to make them useful, though, in the future).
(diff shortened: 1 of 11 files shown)


hamilton:
Obviously extreme hype for this. I seem to be getting the following error from the `ExternalResourceCell`:

screen shot 2018-02-23 at 2 03 04 pm

mdboom:
Hmm, I can't reproduce that error here, but it is a 21MB file being served over Github Pages, so kinda not surprised it's timing out.  As a workaround, you can checkout the pyodide-demo repository and copy the files to your `dev/` directory in Iodide and then use a local file path (`./python.asm.js`) as an external dependency to load it.  That's how I've been developing along.  I suspect we'll need to have a solution for getting that large file over the net more reliably.

hamilton:
Got it - will do!

bcolloran:
i'm going to close this since pyodode is up and running in another repo and we haven't touched this in a while. feel free to re-open if that is the wrong move.

mdboom:
Nope, that's fine.  I'll create a PR against iodide-examples to get something in over there.

PR #406, opened 2018-02-27
Updates localStorage save to use jsmd. Updated test.

### LINKED ISSUE #401: autosaves and ctrl-s saves to local storage should save/load jsmd
This is a small step towards #351 -- currently, we store full state objects as JSON in local storage. The objective of this issue is to only save jsmd to local storage, and to load jsmd from local storage.

The relevant functions are in the file "jsmd-tools.js". You'll also need to look at the notebook reducer.

bcolloran:
This will fix #401.

This PR blocks further work on row visibility updates, b.c something about saving states to JSON was breaking in my other branch (I think JSON didn't like the enums). Since this was on the list anyway, i figured I"d knock it out to unblock the other stuff.

### DIFF
--- a/src/jsmd-tools.js
+++ b/src/jsmd-tools.js
@@ -35,6 +35,7 @@ const cellTypeToJsmdMap = new Map([
 const jsmdValidNotebookSettings = [
   'title',
   'viewMode',
+  'lastSaved',
 ]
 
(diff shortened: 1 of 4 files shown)


bcolloran:
thanks Mike!

PR #407, opened 2018-02-27
Add date and time of export to jsmd meta

### LINKED ISSUE #371: add datetime at time of export to JSMD meta section
We need to add the date and time of export to the jsmd meta information when the jsmd is serialized at the of export. We also need to add tests around this new functionality.

The relevant files are: `src/jsdm-tools.js` and `test/jsmd-tools.test.js`

pradeepgangwar:
I need suggestions for this functionality, although i have added a working modal but i am not sure if that's a correct fix.

Fixes #371

### DIFF
--- a/src/jsmd-tools.js
+++ b/src/jsmd-tools.js
@@ -200,7 +200,7 @@ ${cell.content}`
       metaSettings[setting] = state[setting]
     }
   }
-  metaSettings.exportAt = exportDatetimeString
+  metaSettings.lastExport = exportDatetimeString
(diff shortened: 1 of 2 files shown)


bcolloran:
Nice work! i like the approach of making `stringifyStateToJsmd` take the date string as a parameter -- it's obvious now that I think about it, but it hadn't occurred to me. It's a nice functional style, very testable.

I made a couple small tweaks, but nothing major. Merging the PR!!

Issue #405, opened 2018-02-27
remove "import from JSON" from menu


bcolloran:
this is not needed anymore now that we have self-bundled html+jsmd files.

@hamilton, assigned this to you b/c you can easily knock it out while you're working on the other menu stuff.

hamilton:
yep - already removed in my branch :)

Issue #395, opened 2018-02-23
[META] change directory structure to better compartmentalize our growing file lists


hamilton:
- Now that we have many components for cells, we should put these all in a new dir under `src/components`.
- Once we break up `output.jsx` we should have a separate dir for those components / functions (#276)
- all the new menu work that goes into #393 should be in a new dir under `components` as well

hamilton:
Closed with #537.

Issue #393, opened 2018-02-22
refactor menus to use new UserEvents


hamilton:
This is built on PR #384 - the menus should consume the same set of `UserEvent`s that the keybindings use.

Issue #390, opened 2018-02-22
invalid prop types in notebookHeader and EditorModeToolbar


bcolloran:
looks like these should be `date`s

I'm not sure how i got to this error, might just need to trace backwards to see what the originating data type is.

could be dealt with as part of header/menu refactors --  #319, #176, etc

```
Warning: Failed prop type: Invalid prop `lastSaved` of type `date` supplied to `NotebookHeader`, expected `string`.
    in NotebookHeader (created by Connect(NotebookHeader))
    in Connect(NotebookHeader) (created by Page)
    in div (created by Page)
    in Page (created by Connect(Page))
    in Connect(Page)
    in Provider
iodide.master.js:398:7

Warning: Failed prop type: Invalid prop `lastSaved` of type `date` supplied to `EditorModeToolbar`, expected `string`.
    in EditorModeToolbar (created by NotebookHeader)
    in NotebookHeader (created by Connect(NotebookHeader))
    in Connect(NotebookHeader) (created by Page)
    in div (created by Page)
    in Page (created by Connect(Page))
    in Connect(Page)
    in Provider
iodide.master.js:398:7
```

PR #408, opened 2018-02-27
[WIP] menu UI refactor

hamilton:
This work ended up being a major refactoring of about half the component code base - all the menus, buttons, and UI components that are _not_ part of the cell set - to work with our `UserTask` class. 

It:

- [x] migrates to material-ui v1 (`@next`), which works a bit more smoothly / is more performant / is more composable than `@latest`
- [x] refactors all the menus to use the new `UserTask` item
- [x] matches styling and tasks of the previous menu
- [x] migrates the examples to use the new external url task
- [x] makes just about every component a container component, while following the `propTypes` / unconnected conventions found in the cells
- [x] finishes some of the reorg into the right directories
- [x] tests for the unconnected components are necessary

This should fix: #405, fix (partly) #395, #393, #390, #389, #319, #176, #160, and lay the groundwork for closing #162.


hamilton:
It turns out my approach simply breaks material ui's oldest `@latest` version. It wasn't obvious when we started this library that `@latest` was actually heavily discouraged in favor of `@next`, the beta version of v1. It turns out it makes a big difference. So I'm going to be porting over the entire app to `@next` and see if it will solve some of the various click / keyboard issues I'm facing now with these composed menu items. On the whole this is an annoyance and is quite a refactor to go ontop of this refactor, but it feels necessary. Hoping to have this ready for review by Thursday.

hamilton:
I've migrated most of the code base to the beta version of `material-ui`, and everything works as expected now, though some styling is slightly different. 

Drawers do not automatically close on `esc` - we should see if that should be expected behavior.

hamilton:
Worked through an extensive QA process w/ @bcolloran. As soon as the tests pass, we are going to merge this.

On the whole, this refactor solves a lot of problems around our UI components, but it also introduces some additional issues and surfaces some rough edges we should check on. Drawers are a bit rough now, since material-ui has updated.

bcolloran:
:100: great work @hamilton

PR #412, opened 2018-03-01
Row visibility

bcolloran:
This ended up being a bigger PR than i'd wanted, but it's not actually that bad. The focus here is on refactoring how row visibility works (in order to support a middle row for printing and other evaluation side-effects), but a bunch of stuff around jsmd parsing and serialization needed to be tuned up to support this as well. I also had to update a couple tests.

Once this is in place, the side-effect row should be pretty easy to pop in.


bcolloran:
@hamilton -- merging in you menu upgrades was zero work, no conflicts whatsoever.

This is passing everything, and when i run thru the QA steps we did this afternoon i don't see any problems, but i could use another set of eyes.

src/state-prototypes.js
```diff
@@ -1,85 +1,109 @@
...
+
 function newCellID(cells) {
   return Math.max(-1, ...cells.map(c => c.id)) + 1
 }
 
-// function newCellRow(collapseEditView, collapsePresentationView) {
-//   // these track the collapsed state of cell row
-//   // must be one of "HIDDEN", "SCROLL", "EXPAND"
-//   return { collapseEditView, collapsePresentationView }
+// function newCellRow(rowType, EXPLORE, REPORT) {
```
hamilton: we probably don't need this commented code

src/reducers/notebook-reducer.js
```diff
@@ -95,7 +95,6 @@ const notebookReducer = (state = newNotebook(), action) => {
       clearUserDefinedVars(state.userDefinedVariables)
       try {
         nextState = JSON.parse(window.localStorage.getItem(action.title))
-        console.log()
         console.log(`"${nextState.title}" is currently saved in localStorage as JSON.
```
hamilton: this should be a `console.warn` probably

bcolloran:
fixed those issues, merging.

this is the next to last step of #363, now just need to actually add the cell row, which should be a small PR

PR #435, opened 2018-03-05
Fix CSS cell proptype issue

djbarnwal:
Changing the `cell-type` of a newly created cell to `css` logs an error 
cellproptype

This PR fixes the above issue.


bcolloran:
Ah, yep. Nice catch, thanks @djbarnwal !! :-)

PR #436, opened 2018-03-05
Fix styling issues of history side pane with chrome

### LINKED ISSUE #417: history side pane has styling issues in Chrome when number of history items ecxeeds page height
This appears in `master` but only in Chrome, not Firefox. It appears to be an issue due to the migration to material-ui `1.0.0-beta35` in #408.

One solution may be to utilize the `Paper` component within the wrapping `SidePane` component.

__Steps to reproduce:__

1. Open any notebook in Chrome ([this link](https://iodide-project.github.io/master/?url=https://raw.githubusercontent.com/iodide-project/iodide-examples/master/output-handling.html) should do the trick)
2. run all cells
3. open the history pane

mar-01-2018 19-07-26

pradeepgangwar:
These small tweaks were able to fix this bug. I hope these changes are in acceptable range. Suggest changes if there is some other way.

Fixes #417 

screenshot from 2018-03-06 04-45-04

### DIFF
--- a/src/page.css
+++ b/src/page.css
@@ -470,8 +470,11 @@ div.last-saved {
 .pane-title {
   margin:5px;
   margin-right:0px;
-  display:flex;
-  align-items: center;
+  display:-webkit-box;
+}
+
+.pane-title h1 {
+  margin-top: 6px;
 }
 
 .pane-title .menu-button {


bcolloran:
Was this tested on Firefox too? (To ensure it looks the same in both cases?)

pradeepgangwar:
yes it works on Firefox too.

bcolloran:
Perfect, thanks @pradeepgangwar. @hamilton, as our menus/panes/styling expert, can you review when you have time?

hamilton:
Yes, this works on all platforms.

hamilton:
thanks @pradeepgangwar!

Issue #97, opened 2017-09-26
Schema validation for state when in dev-mode


bcolloran:
Should we consider using enums for things with enumerable values? There are at least:

cell types (MD, raw, js, ...)
modes (edit, command, ...)

Surely other stuffs. May not matter at the stage we're at, but I think pros do stuff like this as code bases swell in order to make it easier to quickly see where all the possible values of a thing are declared, and in order to avoid errors from sprinkling strings all over the place.

bcolloran:
there are at least four subparts to this refactor
- [ ] views (report/explore)
- [ ] cell types (js/md/css/...)
- [ ] cell row display modes (hidden/visible/scroll)
- [ ] app mode (command, cell edit)

bcolloran:
Did some reading about redux state and chatted with @hamilton -- I think the path forward here is to add schema validation to the redux state during reducer operations (when in dev mode probably). This will be better than enums, b/c it won't require rewriting each individual file, and we won't have to put actual enum objects in the state -- best practice in redux is apparently to keep things simple and serializable in the redux store, not to put complex js objects and functions in the store.

This will be kind of like "propTypes" checks for the whole state. I'm sure there are libraries that do this already. I'll look into it.

PR #437, opened 2018-03-05
state schema validation

bcolloran:
Ok, here's a first cut at schema validation for the the redux store.

It's pretty simple overall. There is some additional cruft in the code, but i left it there on purpose; once we get this merged, I'll do a series of small PRs to handle some of the FIXMEs and complete implementation of some of the ideas in the commented out sections.

Side note: this validates the the store against the supplied json schema on *every* reducer invocation (every user action). let's keep an eye open for performance degradations. Also, we should start to think about "production builds" with this and the logger middleware disabled (though we might be some time off from actually needing them)


bcolloran:
this will fix #97

src/state-prototypes.js
```diff
@@ -16,11 +16,69 @@ const StringEnum = class {
...
+  },
+  additionalProperties: false,
+}
+// cellSchema.required = Object.keys(cellSchema.properties)
+// cellSchema.minProperties = Object.keys(cellSchema.properties).length
+
+const stateSchema = {
+  type: 'object',
+  properties: {
+    title: {}, // FIXME change to string ONLY with default '' or 'untitled'
```
hamilton: Is there a reason this is an object by default instead of `null` or `undefined`?

bcolloran: yea it turns out that JSON schema doesn't have "undefined", so you can use a raw object to represent the union of all types, in this case "undefined" and "string".

I'll follow up with a series of PRs that change everything in state to use single fixed types rather than type unions.

PR #438, opened 2018-03-06
object and global environment-aware autocomplete for JS cells. dope.

bcolloran:


bcolloran:
consider this a first draft. ctrl+space initiates completions, but we can change that behavior (this was the default)

Try typing `window.` in a js cell, then `ctrl+space`. So many things defined in the window namespace!!

I'm stoked, this will cut down my use of the console to inspect objects a lot...

mdboom:
Very nice.  I'll have to figure out how to get this working on the Python side...

Issue #440, opened 2018-03-06
Visualization Algorithm notebook not working


imshubhamsingh:
I was trying to add new features to the notebook, it was working just an hour ago. Now Its broken. 

### What I Did
Just ran the notebook

### What Happened 

![image](https://user-images.githubusercontent.com/17807743/37011143-ca4f5698-2114-11e8-91d6-264d7a95fdaf.png)

### (OPTIONAL) notebook that reproduces bug 

I think this was caused due to addition of ajv in #437

hamilton:
Ah, yes. This would have caused it. We're still in a pretty early stage of this project, so we are hoping to get done as quickly as possible these major changes that break some things. Apologies.

@bcolloran do you have a suggested way to move forward on this? I'm sure it wouldn't be hard to suggest an easy fix to get @imshubhamsingh moving again.

imshubhamsingh:
@hamilton  and @bcolloran, Found the cause of the error, fixed it and filled a PR. The notebook is working now

PR #441, opened 2018-03-06
Added lastExport to stateSchema for issue #440

imshubhamsingh:
@bcolloran It looks like lastExport was missing from state schema for validation


bcolloran:
looks good, thanks for catching this @imshubhamsingh !!

PR #445, opened 2018-03-06
add a middle "side-effect-target" row to JS cells

bcolloran:
This will complete work on #363, adding a middle row to JS cell. The middle row contains a div with class "side-effect-target", which can be targeted by Iodide stdlib code with e.g.
`document.querySelector('div.selected-cell div.sideeffect div.side-effect-target')`

for a dumb example, try:
```
function print(str){
  let div = document.createElement('div')
  div.innerHTML = str
  document.querySelector('div.selected-cell div.sideeffect div.side-effect-target').append(div)
}
for (let i=0; i<10; i++){print(i)}
"return value"
```
for real usage, we'd need to be cleverer about clearing the side-effect target on each cell execution (but this won't be hard).

cell middle rows currently respect the "HIDDEN" overflow state but not the "SCROLL" overflow state, so there's a bit more CSS work to do

### DIFF
--- a/src/components/cell-type-javascript.jsx
+++ b/src/components/cell-type-javascript.jsx
@@ -23,6 +23,9 @@ export class JsCellUnconnected extends React.Component {
         <CellRow cellId={this.props.cellId} rowType="input">
           <CellEditor cellId={this.props.cellId} />
         </CellRow>
+        <CellRow cellId={this.props.cellId} rowType="sideeffect">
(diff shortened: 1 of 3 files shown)


mdboom:
Instead of "side effect" maybe "console"?  That might be confused with the browser's Javascript console, of course, but this is basically a console/tty, in effect, right?

Issue #418, opened 2018-03-02
scrollToCellIfNeeded may be kicking some fixed DOM elements off the page in some circumstances


hamilton:
This is a bug currently in master, surfaced while QA testing #408 in Chrome. This bug does not appear in Firefox.

__steps to reproduce:

1. open an example notebook. [This url](https://iodide-project.github.io/master/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fiodide-project%2Fiodide-examples%2Fmaster%2Fwhat-a-web-notebook-looks-like.html) will do the trick.
2. run all cells
3. select the top element, then repeatedly press `shift + down` until it happens.

The issue appears when the selected cell passes over the webGL cube.

![mar-01-2018 19-12-36](https://user-images.githubusercontent.com/95735/36881797-f7c771a2-1d84-11e8-912d-9d9984596944.gif)

PR #449, opened 2018-03-06
Presentation Mode: z-index of button, padding problems pushing content off page

hamilton:
This is the first of a number of PRs centered on getting the presentation view more usable out the box and more extensible generally. It:

- removes the white space that came at the top of the mode, which was there to let the "edit" button be visible re: z-indexing
- changes the z-index of the "edit" button so it is not covered by CodeMirror instances (visible when a markdown cell has not been rendered atm, and in the future will be an issue when we can set JS cell inputs as visible)
- temporarily removes the strange padding issues that caused content to go off the bottom part of the page.

The css for presentation view is fairly broken. We are going to have to reorganize how we do that so we can make this environment more useful for a wide variety of circumstances. But this should make things usable right now, at the very least.

### DIFF
--- a/src/components/toolbar/view-mode-toggle-button.jsx
+++ b/src/components/toolbar/view-mode-toggle-button.jsx
@@ -29,6 +29,7 @@ export class ViewModeToggleButtonUnconnected extends React.Component {
       <Button
         style={{ color: this.props.textColor || '#fafafa' }}
         onClick={this.toggleViewMode}
+        variant="flat"
         mini
       >
         {this.props.viewMode === 'presentation' ? 'Edit' : 'View'}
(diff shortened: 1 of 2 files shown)


bcolloran:
this fixes
  1. removes the white space that came at the top of the mode, which was there to let the "edit" button be visible re: z-indexing
  2. changes the z-index of the "edit" button so it is not covered by CodeMirror instances (visible when a markdown cell has not been rendered atm, and in the future will be an issue when we can set JS cell inputs as visible)

for me, but i'm not sure what this one is:

  3. temporarily removes the strange padding issues that caused content to go off the bottom part of the page.

i think this moves us forward, so i'm going to merge.
(note to self -- still need to figure out #418/#442 in a separate PR)

Issue #455, opened 2018-03-08
tighten up notebook header and make sure notebook title gets enough space


bcolloran:
there a lot of dead space in the header at this point, let's see if we can buy back a few vertical pixels and let's tighten up the horizontal spacing. it looks like the bar is 64px atm; i think somthing like 45-50 would be better proportioned for a desktop experience (my assumption is that the material UI lib uses phone/tablet defaults).

also, note that in the following image, the title text, which should read "Notebook title", is squeezed into a tiny horizontal area, cutting off the start/end of the string. We should make sure that the title text has enough space, and in the process we can tighten up the horizontal spacing around the buttons by a few px to keep the proportions looking good.

![image](https://user-images.githubusercontent.com/2380975/37126642-cafdcb52-2226-11e8-93e0-e45c44a83a10.png)

hamilton:
I'll take this on. It's not quite a bug, since the issue is I ported over our toolbar for my full screen width (which is how I have been using the notebook). There is a "last saved" element to the right of the title which is taking up just a bit too much space, it seems, if you tend to keep windows narrower. One option is to remove it entirely, and the other is to make it smaller. I'm going to opt for the latter for now.

Also, this should be two issues probably, not one, but I can take care of both in the same PR.

hamilton:
I could also remove the "saved 8 minutes ago" in favor of an icon that has a tooltip, which would give maximum space to the title (which I'm all for)

bcolloran:
yea that looks good, but of course it's hard to tell without seeing the window in a narrow view. For now, i think testing down to about 850 or 900px width is what i'd aim for.

ya i thought this was all related and would be better bundled as one issue/pr, but certainly go ahead and break it up if that seems better

hamilton:
No prob - this basically boils down to a responsive design task, shouldn't take long

Issue #124, opened 2017-10-10
tooltips everywhere


bcolloran:
not a priority for some time, but eventually we'll want nice tooltips all over the place to make the UI maximally learnable.

PR #457, opened 2018-03-08
toolbar: reduce height, give title element more space, and add tooltips

### LINKED ISSUE #454: tooltips missing since header refresh
@hamilton -- looks like the tooltips that used to exist for the header buttons are no longer there. I looks like we just need to wire the "title" properties of the usertask definitions to the button components. Will that make a good first issue, or do you prefer to handle it yourself?

hamilton:
Should close #454 / #124, and #455:

- height of toolbar is now 50px, and the buttons have been reduced in height / width
- the title now takes up more space by default. The toolbar is _not_ really responsive at this point, but it should work if you keep a browser window at half of your laptop screen width, more or less.
- tooltips everywhere.

### DIFF
--- a/src/components/toolbar/editor-mode-title.jsx
+++ b/src/components/toolbar/editor-mode-title.jsx
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types'
 import Helmet from 'react-helmet'
 
 import { connect } from 'react-redux'
-import Tooltip from 'material-ui/Tooltip'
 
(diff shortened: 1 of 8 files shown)


bcolloran:
Looks great! I want to go ahead and merge this, so i don't want to block on the only issue i notice, which is that the tooltips are a bit small/difficult to read, which i'll file as a separate issue that we can handle later.

bcolloran:
added #458 re tooltip text size

PR #466, opened 2018-03-10
Increase tooltip size

### LINKED ISSUE #458: tooltip text is a bit small
on my screen it looks like its only 7-8 px, very small. should be standard 11-12ish probably.

pradeepgangwar:
I am not sure whether this implementation (directly implementing via CSS) is fine or not. But this got it working. Attached are screenshots for reference.

Fixes #458 

### Before 
screenshot from 2018-03-10 06-08-38

### After
screenshot from 2018-03-10 06-09-13

### DIFF
--- a/src/page.css
+++ b/src/page.css
@@ -506,6 +506,10 @@ div.last-saved {
   font-size: 12px
 }
 
+#tooltip-icon {
+  font-size: 0.800rem;
+}
+
 /************************************************/
 /* side pane styles */
 


bcolloran:
the result looks good to me, and as far as i know, setting this in the CSS should be ok-- but i've requested a review from @hamilton since he knows the material-ui and menu code well

hamilton:
This is a sensible way of dealing with it. I think we should _probably_ change these from `id=tooltip-icon` to `class=toolbar-tooltip` since I'm now realizing we're using ids as if they were classes, and change the `<Tooltip>` component to have `classes={{root: 'toolbar-tooltip'}}`, but this should work for the time being.

hamilton:
Thanks @pradeepgangwar!

PR #468, opened 2018-03-10
fixes #462

### LINKED ISSUE #462: create an option for "open in presentation mode"
We need this before we begin shipping our first substantive Mozilla analyses. 

This requires adding:

- something in the store / jsmd export to open the notebook in presentation mode
- one menu item that makes it easy (for now) to export in presentation mode as default
- appropriate tests, validation, etc.
- running all cells when loaded in presentation mode (#162)

bcolloran:
the run all on load is hacky, but run all has always been hacky and weird. Works on the "what a we NB looks like" example, but @hamilton please test on some of your notebooks. testing steps:
(1) get a notebook in exportable shape
(2) `shift+E` or use the menu to "export as report" (save file under a new name probably)
(3) open just saved file

expected:
- should open in report view
- should run all with the same results as if you pressed the "run all" button and switched to report mode

the run all stuff makes me nervous, and i'm sure there are racy bugs around it that we'll need to track down later, but i think this should work as well as run all ever does (in explore view)...

### DIFF
--- a/src/evaluate-all-cells.js
+++ b/src/evaluate-all-cells.js
@@ -2,6 +2,11 @@ import actions from './actions'
 
 export default function evaluateAllCells(cells, store) {
   cells.forEach((cell) => {
+    if (cell.cellType === 'css') {
+      store.dispatch(actions.evaluateCell(cell.id))
+    }
+  })
+  cells.forEach((cell) => {
     if (cell.cellType === 'markdown') {
       store.dispatch(actions.evaluateCell(cell.id))
     }
@@ -9,7 +14,7 @@ export default function evaluateAllCells(cells, store) {
   window.setTimeout(
     () => {
       cells.forEach((cell) => {
-        if (cell.cellType !== 'markdown') {
+        if (cell.cellType !== 'markdown' && cell.cellType !== 'markdown') {
           store.dispatch(actions.evaluateCell(cell.id))
         }
       })
--- a/src/evaluate-all-cells.js
+++ b/src/evaluate-all-cells.js
@@ -1,0 +1,19 @@
+import actions from './actions'
+
+export default function evaluateAllCells(cells, store) {
+  cells.forEach((cell) => {
+    if (cell.cellType === 'markdown') {
+      store.dispatch(actions.evaluateCell(cell.id))
+    }
+  })
+  window.setTimeout(
+    () => {
+      cells.forEach((cell) => {
+        if (cell.cellType !== 'markdown') {
+          store.dispatch(actions.evaluateCell(cell.id))
+        }
+      })
+    },
+    42, // wait a few milliseconds to let React DOM updates flush
+  )
+}
--- a/src/task-definitions.js
+++ b/src/task-definitions.js
@@ -8,6 +8,7 @@ import { isCommandMode,
   getCellBelowSelectedId,
   getCellAboveSelectedId, prettyDate, formatDateString } from './notebook-utils'
 import { stateFromJsmd } from './jsmd-tools'
+import evaluateAllCells from './evaluate-all-cells'
 
 const dispatcher = {}
 for (const action in actions) {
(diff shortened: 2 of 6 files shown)


src/evaluate-all-cells.js
```diff
@@ -0,0 +1,19 @@
+import actions from './actions'
+
+export default function evaluateAllCells(cells, store) {
+  cells.forEach((cell) => {
```
hamilton: I'm going to make a bold claim here. The FIRST set of cells that should be run is actually the CSS Cells, if possible, then the markdown cells. Otherwise we get a flickering between the default styles and the ones applied via CSSCells. I believe we'd do it right here in this function. It's also possible that we can very easily just do all the markdown + css cells together and that'd take care of the issue!

src/task-definitions.js
```diff
@@ -266,6 +248,14 @@ tasks.exportNotebook = new UserTask({
   callback() { dispatcher.exportNotebook() },
 })
 
+tasks.exportNotebookAsReport = new UserTask({
+  title: 'Export Notebook as report',
+  keybindings: ['shift+e'],
```
hamilton: Could we make this keybinding `meta+shift+e`?

src/task-definitions.js
```diff
@@ -266,6 +248,14 @@ tasks.exportNotebook = new UserTask({
...
+  displayKeybinding: 'Shift E',
```
hamilton: I know this is extremely ugly right now, but could we replace `Shift` with `\u21E7 `? I need to do a small refactor that automatically generates the right human-readable keybinding, but for now this works on the other menu items that use `Shift`.

src/evaluate-all-cells.js
bcolloran: Yep, that seems reasonable, and easily changed

src/task-definitions.js
bcolloran: Oh duh yes of course

src/task-definitions.js
bcolloran: Hmmm I'm looking at some mac keyboards right now and they all label that key with the text "shift". I've looked at a couple other cross platform programs, and the convention seems to be to write out the word rather than use a symbol. If using a symbol is the convention on mac, we can a add that later but I'm in favor of keeping the explicit string as the base case for other platforms (and for mac until Mike's suggesting is implemented, since "shift" will at least not be confusing)

bcolloran:
Thanks Hamilton, I'll fix up those bits in a little while and merge

src/task-definitions.js
bcolloran: hmm, i'm going back and looking at the menus, and i'm actually finding the use of symbols in general to be somewhat ambiguous, esp in the case of the up and down arrows, where there are a couple arrows next to each other.

image

I just went to go look at the FF menus on my girlfriend's Mac, and i see that using these symbols is the standard there... but she didn't even know what all of them meant. With the exception of the "open apple" symbol (which is unambiguous on a mac kbd) and *maybe* the shift arrow, I don't think we should be using symbols for the kbd shorcuts, and definitely not as the default for all platforms. If we do use them it should be as a variant for mac only.

As long as i'm in here, i'm going to change these default displays as well. I'll leave the mac symbols in the comments so that we won't have to hunt them down once we're ready to implement the mac variants.

bcolloran:
@hamilton, i'm going to hold off on merging since i also updated the kbd shortcut strings.

Tbh, i don't think we should follow the mac convention here at all, since my intuition and user research sample of one leads me to believe that the mac convention of using symbols is likely to be more confusing for average users, even if they always use mac. But i'm quite sure we should default to the non-mac display style, and build the mac style as the special case (if we want to use it at all).

In any case, I don't want to merge without consulting you since there are additional changes that are outside of what we've discussed so far

hamilton:
@bcolloran Thanks for checking that. You're 1005 right - the Mac keybindings for shift should only be available on macs, if at all, so moving over to `Shift` for the `displayKeybinding` across the board is a good idea. We can circle back on a more OS-based nuanced approach once we get there. We're mostly there on the most important one at any rate - cmd / ctrl for the `meta` keybinding depending on your OS.

bcolloran:
cool, thanks hamilton, merging now

Issue #424, opened 2018-03-02
new HistoryPane does not seem to update when new history items are added


hamilton:
Steps to reproduce:

1. go to [this example notebook](https://iodide-project.github.io/iodide-examples/what-a-web-notebook-looks-like.html)
2. open the history pane in the top right corner
3. click the "run all cells" button (the fast-forward icon)
4. observe that nothing is added to history pane. Then click the x
5. open the history pane again. The state is finally updated.

![mar-02-2018 07-48-31](https://user-images.githubusercontent.com/95735/36907840-797de8e8-1dee-11e8-9106-f3dc177c2f16.gif)

hamilton:
fwiw this may have also been a bug before #408 was merged. The Declared Variables Pane updates as expected.

bcolloran:
i'm pretty sure this existed before #408. I think it's just about how/when we feed those props to the history pane, and when it triggers a react update

djbarnwal:
So I was going through the code briefly and it came to my notice that the reducer was [mutating the state](https://github.com/iodide-project/iodide/blob/2439821725487e6b632419bd6cc54fab51a2dce8/src/reducers/cell-reducer.js#L141-L145). The `Object.assign()` function creates a shallow copy. This has been mentioned in [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

> For deep cloning, we need to use other alternatives because Object.assign() copies property values. If the source value is a reference to an object, it only copies that reference value.

I will send a PR to fix this issue.

imshubhamsingh:
I checked it, the data coming from reducer through mapStateToProps is correct, The problem is that somehow the history data is being persisted. Then I thought there must be some bug in Drawer component of material-ui. But that also working fine as I recreated a scenario in [codesandbox](https://codesandbox.io/s/64v8qkro1n).

hamilton:
Great catch @djbarnwal!

bcolloran:
fixed by #470

PR #470, opened 2018-03-10
Fix Reducer

djbarnwal:
The first commit fixes Issue #424  
While working on the issue I noticed how external dependencies are added in the state. This brought in light another issue. On dispatching `CLEAR_VARIABLES` action, though the variables get cleared, executing the external dependency cell again doesn't work as expected as the source was already in the state. The second commit fixes this issue.


bcolloran:
this all looks great and works as expected. many thanks @djbarnwal!

Issue #415, opened 2018-03-02
uncaught exception with scrolling to cell


hamilton:
Error around `scrollToCellIfNeeded`

```
cell-reducer-utils.js:24 Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null
    at scrollToCellIfNeeded (file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:79932:21)
    at cellReducer (file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:65940:100)
    at reducers.reduce (file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:65668:59)
    at Array.reduce (<anonymous>)
    at file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:65668:42
    at dispatch (file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:45133:22)
    at Object.dispatch (file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:65644:7806)
    at Object.dispatcher.(anonymous function).params [as selectCell] (file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:12209:95)
    at UserTask.callback (file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:12348:16)
    at HTMLUnknownElement.boundFunc (file:///Users/hamiltonulmer/Code/mozilla/iodide/dev/iodide.dev.js:38655:9)
```

imshubhamsingh:
@hamilton I've a temporary fix for it as of now. The main problem is that we are not doing anything for asynchronous codes. So while adding the cell into the nodebook `tasks.addCellBelow` dispatches two function simultaneously because of which the error is showing. The cell is created and selected at the same time, not one after the other. We should really add **redux-thunk** to the project, It'll solve this problem as well as the usage of iodide-api in future.

bcolloran:
@imshubhamsingh thanks for looking into this issue!

I think there is something deeper going on here --

[...]

PR #473, opened 2018-03-11
temporary fix for #415

imshubhamsingh:
Just a temporary fix for #415.


bcolloran:
see comment at #415 -- let's try another approach to this

Issue #178, opened 2017-12-05
codemirror slow with non-fixed sized editors


bcolloran:
see #176 -- re: "there are a bunch of really slow calls to endOperations (with sub-calls endOperation_R2 and endOperation_finish)"

this looks like it has to do with something weird about the scrolling model of codemirror instances:
https://github.com/codemirror/CodeMirror/issues/4076
That issue is still open. Not sure what to do about it...

Issue #92, opened 2017-09-23
code mirror perf problems


hamilton:
We have seen that CodeMirror (or at least how we're using it / how it interacts with our app) has some performance issues.

We are welcoming PRs for replacing CodeMirror with any of the other JS-based editors.

bcolloran:
try the following to see the performance choke:
1) on /dev, open the following NB with the url trick ?url=https://iodide-project.github.io/iodide-examples/eviction-notices-by-sf-neighborhood--1999-2015.html 
2) load the notebook, then go down to the css cell at the bottom and type a bit.
3) typing speed will be slow
4) also try adding cells at the end of the NB

Note: this may be a Firefox-specific problem, so a fix for this issue MUST TEST ON FF AND CHROME

hamilton:
@bcolloran on second thought, I really don't see any performance issues with CodeMirror on large cells. I'd be happy to close this issue, since the main problem - severe hangs which negatively impacts the UX - seems to have been solved with your PR.

hamilton:
@bcolloran I believe this has been fixed. Going to close. If it seems slow again we can reopen.

PR #477, opened 2018-03-12
performance imporovements, making props more stable

### LINKED ISSUE #403: data frame viewer causing terrible app performance?
### What I Did

loaded a couple data sets (5550 rows x 44 cols; 70k x 6)

returned values of these tables and a mutated dataset

app slowed way way way down; editing etc was basically unusable.

reran these cell, but returned a string in each, and app performance came right back.

this needs to be tested/verified, but initially it looks like something in the data table viewer is killing perf.

### LINKED ISSUE #420: outputs retroactively handled by output handlers
steps:
- load https://iodide-project.github.io/iodide-examples/output-handling.html
- run all
- 3rd cell with `let obj = {'lat': 42, 'lon': 49}` shows tree viewer
- switch to report view ("view" button)
- switch back to explore view ("edit" button)
- 3rd cell with `let obj = {'lat': 42, 'lon': 49}` now shows image

this is probably one of those weird react quirks where triggering an component re-render also rerenders all children, which in this case will use the most recently registered output handlers.

### LINKED ISSUE #476: `onContainerClick` prop in  cell-editor appears to be unused; doublecheck this
this method was only used by the mardown cell to call `enterEditMode`.

also, markCellNotRendered should be handled in the from the somewhere other than this; probably, the implementation should be changed s.t. we:
- we mark a cell "dirty" if it's content changes from the last time it was evaled
- instead of keeping track of a `rendered` bool, we have a `evalCount` int that counts the number of times each cell is evaled. this will also help in the "shouldComponentUpdate" method of CellOutput -- we can only update in case the evalCount has increased

bcolloran:
this PR fixes #476 and fixes #420 for sure, and i think it fixes #403.

Hamilton, please verify the steps in #420 are now working. Please also do any other perf stress testing you can think of to see if #403 seems better, and to see if #178 and #92 might be obselete too (the editor perf seems pretty good to me at this point, but i think you had scenarios that we're known to cause problems?)

### DIFF
--- a/src/components/cell-container.jsx
+++ b/src/components/cell-container.jsx
@@ -19,6 +19,14 @@ class CellContainer extends React.Component {
     }).isRequired,
   }
 
+  // shouldComponentUpdate(nextProps) {
+  //   return (
+  //     this.props.selected !== nextProps.selected ||
+  //     this.props.pageMode !== nextProps.pageMode ||
+  //     this.props.cellType !== nextProps.cellType
+  //   )
+  // }
+
   handleCellClick = () => {
     if (this.props.viewMode === 'editor') {
       const scrollToCell = false
@@ -29,6 +37,7 @@ class CellContainer extends React.Component {
   }
 
   render() {
+    console.log(`CellContainer rendered: ${this.props.cellId}`)
     const cellClass = `cell-container ${
       this.props.cellType
     } ${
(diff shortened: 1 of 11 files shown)


bcolloran:
Sounds good. I'll leave the CodeMirror issues open and close the other ones

Issue #237, opened 2018-01-30
presentation mode should have a good fixed width, line / paragraph spacing


hamilton:

hamilton:
This was fixed with #483 - closing now.

Issue #210, opened 2017-12-15
[META] overall styling improvements


hamilton:
- [ ] have a few different themes for presentation mode: something that resembles a "dashboard" theme, something that resembles a "NYT data presentation" theme (which is perhaps a general report theme).
- [x] consider using css grids + polyfill for better cell support (#207) 
- [x] implement `CSSCell` for custom style hacking in-notebook (#189)
- [x] redesign some elements of presentation mode so that it "works" out the box the way users expect
- [ ] ~~remove saved notebooks from menu and push into an overlay (#209)~~

bcolloran:
obselete

Issue #199, opened 2017-12-13
[META] presentation mode design changes


hamilton:
In terms of functionality, presentation mode is mostly-implemented. In terms of readability and general aesthetics, it needs a bit of work.

- [x] The block-level element styling doesn't feel 100% right at the moment. Let's fix the spacing and follow typical design principles around these things.
- [x] Having near full-width for text and code in presentation mode makes the text fairly unreadable. This should not be the default.
- [ ] Allowing the option for a Table of Ccontents would be very useful for longer reports and presentations.
- [x] We should make sure all the correct css classes are in-place for easy theming of presentations.

hamilton:
I'm going to close this since #483 addresses everything in-scope for this. I would say the ToC is `wontfix` for now.

PR #483, opened 2018-03-14
add a usable default stylesheet & partition out presentation css from app style

### LINKED ISSUE #331: CSS might have regressed without bootstrap reset and basic styles.
might be worth seeing whether we like the basic bootstrap styles (without all of their crazy buttons and grids and stuff). Presentation mode is looking a little shabby

hamilton:
This should close #331, #237, parts of #210, parts of #199. 

I've added a new default stylesheet for `user-markdown` which provides sensible report-style formatting.

Steps to test it out for yourself:

- run your dev server, and use what is in `dev/`
- add `?url=https://gist.githubusercontent.com/hamilton/436ca12593652f830dcb5b141d336b37/raw/283a7a9436d52742c4a109719b802eb262f2f5b1/default-presentation-2.html` to load an example (run all cells in presentation mode is broken so you will have to "run all" then switch to presentation mode"
- examine how we're handling style now.


bcolloran:
it all looks really good, i only see one thing that looks a little off, which is the extra padding around user-markdown, i think maybe these styles.

```
.user-markdown:last-child {
    padding-bottom: 60px;
}
.user-markdown:first-child {
    padding-top: 30px;
}
```

this results in an extra 90px of whitespace between paragraphs if you have two consecutive MD cell.

Take a look at
` ?url=https://iodide-project.github.io/iodide-examples/what-a-web-notebook-looks-like.html `
for an example

peek 2018-03-13 23-12

Other than that little quibble, oh man it looks great!

bcolloran:
Oh, i decided to spot check some other notebooks, (take a look at `?url=https://iodide-project.github.io/iodide-examples/output-handling.html`), and i noticed another issue:

If you're in "EDIT" view and your window is moderately wide (1150px or more) it's kind of jarring that the MD is centered rather than being aligned to the left side of the cell, so it might be best not to change the horizontal alignment in edit view (but i think having an 800px max width on the container is probably fine for edit view, as long as it's aligned to the left)

Also, the extra top/bottom padding really jumps out in that example.

image

hamilton:
@bcolloran 

- good catches re: the margining on the top and bottom. There's still a bit of funky css inherent in presentation mode that I may need to work out, so it is clear that this solution was probably a bit too messy. Without that bit of css there is no comforting whitespace below or above the cell container, but the solution should be something a bit better than this (and obviously not as weird with multiple MD cells!)
- I like this idea about the editor & auto margining. It does indeed feel weird centered in the editor. This should be a really easy fix - set the margin to not auto for editor.

hamilton:
@bcolloran made those changes. Let me know what you think!

- editor mode rendered markdown is now left-aligned with the left edge of CodeMirror instances (so this is the gutter left edge of a js cell, which is less elegant, but the absolute left edge of a MD cell editor, which looks fine). The width is still ___px.
- the between-cell stuff is fixed, and I added a bit of padding between cells in presentation mode to look a lot more seamless (as if it were one big mdcell).

bcolloran:
:clap: 

man, i really looks great! so very much improved, thank you hamilton! merging!

PR #484, opened 2018-03-14
App info messages

### LINKED ISSUE #448: should we move console.warn out of the jsmd tools?
original point: I assumed the parse warnings were being triggered by an outdated jsmd blob used in testing, but that's actually the right idea - we want to induce the parse warnings.

bcolloran:
@hamilton, this is a pretty light weight PR that adds a pane for making app info, warnings, and error messages recoverable after an alert.

In addition to the venerable #166 about app errors, i've realized that i need this making saving to gists sensible. it will also allow me to fix #448.

note that this is only part one of a full app error/warn/info solution; we'll also want the little snackbars that alert you when there are new messages, and that have a "more" link to pop the full info pane with the detailed message. but for saving anonymous gists, the alert is less important initially than the scrollback-- we need to be able to print out 
- the url of the gist
- as well as the sharable `iodide.html?gist=...` link,
and both in a location the user can get back to.

i think this is pretty simple, just followed your template. lmk if anything seems wrong.

### DIFF
--- a/src/components/toolbar/app-info-pane.jsx
+++ b/src/components/toolbar/app-info-pane.jsx
@@ -1,0 +1,34 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { connect } from 'react-redux'
+
+import SidePane from './side-pane'
+
+// import DeclaredVariables from '../declared-variables'
+import tasks from '../../task-definitions'
+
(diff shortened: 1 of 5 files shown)


bcolloran:
Thanks Hamilton. Ya def file those other issues, we'll sort it out.

Issue #475, opened 2018-03-11
[META] menu tune-ups


bcolloran:
Filing this as a single checklist issue rather than a bunch of issues to avoid clutter. I think this GIF should demonstrate all the things i'm noticing--

![peek 2018-03-11 15-29](https://user-images.githubusercontent.com/2380975/37259242-f88c7772-2540-11e8-9952-7f9149c8e920.gif)

- [x] drawing/hiding menus feel a little sluggish, but especially hiding. It's subjective of course, but to me it makes that part of the app feel slow rather than snappy.
- [ ] need to click on each submenu to display it, rather than just unfolding submenus on hover, which is the design convention for desktop UIs
- [x] when a submenu is open, clicking outside the menus only closes the submenu; should close all menus. "clicking outside the menus" refers to clicks entirely outside of the menus and any submenus; clicking at menu higher in the heirarchy should only close menus lover in the heirarchy. clicking outside **all** menus should close all menus
- [ ] when you click outside the root menu (which closes it), the "menu" tooltip is shown again and remains shown until another click, or until some (but not all) uiTask is fired (kbd up/down to select cells doesn't make the tooltip go away, but 'enter' to go into cell edit mode does make the tooltip vanish)
- [x] "cell" submenu missing "add cell above     A"
- [x] submenu text sppears to be a bit smaller than top level menu text. increase size of submenu text (it's too small)

Please note that if you want to work on any of these subissues, you don't have to have one big PR tackling all of them. In fact, little PRs taking them one step at a time can be prefereable (but if you find that all these subissues are tangled up together, larger PRs addressing several at once will be ok)

bcolloran:
regarding the behavior in #485-- if you want to switch menu subsections -- e.g., you first click on the "notebooks" heading, which opens that submenu, but you then wish to look in the "cells" submenu, clicking the "cells" heading does not open the cells submenu, it closes all the menus. so, to specify the behavior a little more:

- [ ] when a submenu is open, clicking outside the menus only closes the submenu; should close all menus. "clicking outside the menus" refers to clicks entirely outside of the menus and any submenus; clicking at menu higher in the heirarchy should only close menus lover in the heirarchy. clicking outside **all** menus should close all menus

i updated the original item and unchecked the box so that we can track it here.

i'm going to revert that PR; the current behavior is actually a bit less desirable than the previous behavior.

hamilton:
Sounds good - sorry for not QAing more fully.

[...]

PR #488, opened 2018-03-15
lowered the transition time for menus

### LINKED ISSUE #486: reduce time of transitionDuration of menus
Part of #475.

[The documentation for `Menu`](https://material-ui-next.com/api/menu/) shows that we can reduce the time it takes to fade out by changing `transitionDuration`.

hamilton:
As per one sub-issue in #475, fixes #486  - menus don't feel snappy because the transition time is just too high. I reduced the main menu to 70ms and the submenus to 50ms, down from whatever it was before (probably 200ms by default).

@bcolloran let me know if this feels better. I can also investigate removing the animation delay that happens on click.


bcolloran:
yea feels much better. thank you!

PR #489, opened 2018-03-15
JsCell to generic CodeCell

bcolloran:
ok guys, this is step towards dynamic language additions. @mdboom, for the most part, i've tried to follow your template and the api you laid out, but make it fit within the Redux state.

Everything should work exactly as before, but not everything that we'll need to have in place to load another language is ready; i haven't tackled adding/altering keybindings, for example, nor do i have a functioning language import cell prototype. this PR is really just about generalizing the handling of language cells and adding the necessary stuff to the redux store. I'll tackle that stuff once we're all happy that this approach seems ok, and especially that this sketch of the API matches @mdboom's need closely enough to work ok.

I think the main thing to notice here is that the languages are for now defined by the following schema:
```
const languageSchema = {
  type: 'object',
  properties: {
    languageId: { type: 'string' },
    displayName: { type: 'string' },
    codeMirrorName: { type: 'string' },
    keybinding: { type: 'string' },
    module: { type: 'string' },
    evaluator: { type: 'string' },
  },
  additionalProperties: false,
}
```

and that evaluation happens just by looking up the module and the evaluator on the `window` object:

```
        const code = thisCell.content
        const languageModule = state.languages[thisCell.language].module
        const { evaluator } = state.languages[thisCell.language]

        try {
          output = window[languageModule][evaluator](code)
          thisCell.evalStatus = evalStatuses.SUCCESS
        } catch (e) {
          output = e
          thisCell.evalStatus = evalStatuses.ERROR
        }
```

So @mdboom, if i understand how pyodide works, that would be e.g.

```
{
  languageId: 'py',
  displayName: 'Python',
  codeMirrorName: 'python',
  module: 'pyodide',
  evaluator: 'evaluate',
  keybinding: 'p',
}
```

and evaluation of some `code` would happen as:

`window['pyodide']['evaluate'](code)`

i tried to match your API pretty closely, will this work? (fwiw-- the reason for choosing to do evals as string lookups off the window object rather than defining a callback function (which is clearly more natural) is that it's a best practice to keep everything in the redux store a trivially serializable object -- which of course helps wrt JSMD too)

Anyway, as of this PR, _everything should just work the same as it did previously_ -- no new functionality, but evaling is more generic. I intentionally broke this into a ton of little bite-sized commits with descriptions of what i was up to, so even tho i touched 19 files, it shouldn't be too bad to see what i was up to if you want to step through the commits. Might be less overwhelming that way.

@hamilton for QA, you can just try running any existing NBs and doing all the normal stuff. From a user's perspective, nothing should be look or function differently at this point.

please lmk if you have thoughts or anything looks funny.


src/actions.js
```diff
@@ -64,10 +64,11 @@ const actions = {
       content: text,
     }
   },
-  changeCellType(cellType) {
+  changeCellType(cellType, language = undefined) {
```
hamilton: should the default language just be `js` here? Seems like this is the place to show the default.

src/actions.js
```diff
@@ -143,6 +144,19 @@ const actions = {
       mode,
     }
   },
+  addLanguage(languageId, evaluate, displayName, codeMirrorName, keybinding) {
+    return {
+      type: 'ADD_LANGUAGE',
+      languageId,
```
hamilton: why do we need `languageId` bot here and in `languageDefinition`?

src/actions.js
```diff
@@ -143,6 +144,19 @@ const actions = {
...
+      languageDefinition: {
+        languageId,
+        evaluate,
+        displayName,
+        codeMirrorName,
```
hamilton: Maybe this should be called `syntaxHight` so as to semantically decouple it specifically from CodeMirror, yeah? I know this is kind of tough because let's say we change editors and we have to change the `codeMirrorName` value to some other string. At least we don't have to update all the symbols I guess.

mdboom: I see your point.  I think it's probably ok to go with the generic variable name (`syntaxHighlight`) and worry about the possibility of maybe needing to rename what languages are called by the syntax highlighting tool later.

src/components/cell-editor.jsx
```diff
@@ -112,9 +112,9 @@ class CellEditor extends React.Component {
       lineNumbers: true,
       keyMap: 'sublime',
       extraKeys: {
-        'Ctrl-Space': this.props.cellType === 'javascript' ? this.autoComplete : undefined,
+        'Ctrl-Space': this.props.codeMirrorName === 'javascript' ? this.autoComplete : undefined,
```
mdboom: Should this key off of the `languageId`?  I think that's probably the more iodide-specific and stable thing to use than `codeMirrorName`, which as @hamilton pointed out, may change if we change syntax highlighters.

Also, I think eventually (though not necessarily in this PR), we'll want to make the autocompletion replacable with a callback, too.

src/actions.js
bcolloran: duh yes of course! good call, thanks! :-)

src/actions.js
bcolloran: ah, surely we do not!

src/actions.js
bcolloran: @mdboom I ported this over from your code without thinking about the names much, but as i reflect on it now for a second I actually think your first intuition on the naming conventino was the right one -- it seems that using `codeMirrorName` is actually maybe semantically useful here. I'm not entirely sure how the (React) CodeMirror plugin situation works, but it looks like there are a few modules that modify the CodeMirror module upon import, and that they define magic keywords in CodeMirror that are specific to CodeMirror. So using `codeMirrorName` is kind of a red flag/reminder that if you are working on this stuff, looking through the Iodide code for our language naming conventions won't be enough, you'll have to look at the CodeMirror docs too in all likelihood.

src/components/cell-editor.jsx
bcolloran: Ah good point-- it actually does key off the `langauageId`, but down in the `mapStateToProps`.

+1 on replaceable autocomplete. that's definitely something we'll want to figure out.

src/actions.js
bcolloran: but maybe `codeMirrorMode` is better:
https://codemirror.net/doc/manual.html#option_mode
https://codemirror.net/mode/index.html

i think these are the CodeMirror-specific mode names that we'll need, and that the other plugins (autocomplete, comment, etc) key off of.

bcolloran:
I was going to hold off on merging this until i had a demo, but since it seems like folks are happy that this is on the right track, and since nothing seem broken in any of our QA, i'm going to go ahead and merge now, b/c i think we'll benefit from master having the build changes in the commit https://github.com/iodide-project/iodide/pull/489/commits/a15a09b9497e636263bce4404dff4cf677cc1a2e

PR #498, opened 2018-03-18
use highlight.js for code blocks within markdown

### LINKED ISSUE #496: syntax highlighting within markdown code blocks
I am writing some documentation w/ iodide for a small project I had started a few months ago, and am noticing that (at the moment) it is hard to get nice syntax highlighting in presentation mode automatically. using the GH-flavored markdown approach (putting the language name in triple tick blocks) works in CodeMirror, but the generated html is not syntax highlighted.

It is possible that this will all be side-stepped by "show input in presentation", which would give us a CodeMirror instance w/ all the usual highlighting. imo this would be a much more concise way of integrating code & prose.

All the same, there are probably use-cases where you want to syntax highlight a code block without having that available. Let's say I want to demonstrate a code block in Clojure or something, but not actually write any Clojure code in iodide.

It turns out `markdown-it` has [an easy way to do this](https://markdown-it.github.io/markdown-it/#syntax-highlighting) kind of thing.

With that in mind, I would like to open this up for discussion. Should we be applying syntax highlighting `markdown-it` plugins?

hamilton:
this closes #496. To test:

- run `npm install` to bring in `highlight.js`
- run on `dev`
 - use this notebook: `?url=https://gist.githubusercontent.com/hamilton/94d5ccd03b8eb3e6a7d59e499902335f/raw/abe7ff8e37c3b6c02d762f517f4249c236b8e4ea/highlight-js.html`


bcolloran:
@hamilton, this code looks fine, but it raises for the question of whether we should be building this into our core bundle, or supporting this as an instance of a more generic problem and building it as a plugin. I think that approach is probably right in the long run -- allow users to add markdown extensions at runtime (this, TOC, etc) so that (a) we don't have to build everything into core Iodide (to keep the bundle slim), and (b) users can be empowered to add little thingies that extend the makrdown parser functionality for all of their weird little niche use cases.

hamilton:
Agreed - I think there is a line to draw here for sure. But I also think
that this is not actually a niche use-case for our users - at least, it is
not any more niche than having KaTeX support by default. All the same, we
should discuss this during tomorrow’s meeting and find the right middle
ground for this going forward!

hamilton:
We've discussed this, and decided that this would indeed be better as a plugin. Going to close this PR for now.

PR #501, opened 2018-03-19
Anonymous gist

bcolloran:
this is the "minimum viable feature" for gist export:
- `ctrl-g` exports to an anonymous gist
- links to the gist and the runnable notebook (using the `?url=` trick) are available in the info pane (top right - 🛈 )

next steps mostly have to do with cleaning up app messages, but hopefully this will work. @wlach in particular, but everyone: lmk if this helps close the usability gap for you, or if there are other things you need for this to be usable.

as with my last few PR, I've tried to break this up into a bunch of little commits to make it easier to see what i'm up to. (I notice now that I've also included some stuff from another branch `code-plugin-cells` in the first few commits of this PR that is not really related to the gist export, but i'll be building off that anyway in another PR soon, so i guess i'll leave that in).


wlach:
So a few things:

1. I think this needs to be in a menu or somewhere else in the UI. A keybinding is just not discoverable enough.
2. Command-G actually resolves to search for the next instance of a token in Firefox and Chrome, so I don't think we should override it. I am not sure if we need a keyboard binding for this operation at all, to be honest.

src/actions.js
```diff
@@ -1,161 +1,216 @@
...
+    };
+    fetch('https://api.github.com/gists', {
+      body: JSON.stringify(gistData),
+      method: 'POST',
+    })
+      .then(response => response.json())
+      .then((json) => {
+        console.log(json)
+        dispatch(updateAppMessages(`Exported to Github gist: 
+<a href="${json.html_url}">gist</a> -
```
hamilton: both of these anchors should probably look like this: 

```html
<a href="${json.html_url}" target='_blank'>gist</a> 
```

The `target='_blank'` will open the link in a new tab in modern browsers.

bcolloran:
Thanks for the feedback guys! Good suggestions. I'll go ahead and remove the keyboard shortcut and add a menu item for now (longer term this is a good candidate for the command palette), and I'll make those other changes

Yeah, the thunk stuff is a pre requisite for the recommended way of doing asynchronous fetches in redux, which is why it's needed on this PR.

bcolloran:
Ok, i've added those changes you guys requested
- links now open in new tabs
- kbd shortcut removed
- menu item added.

plz let me know if you have any other needs around this!

hamilton:
We will add more in #502.

Couple of improvements for this PR:

- [ ] the side pane (for now) should open upon exporting.
- [ ] open a new tab w/ gist link in it.

bcolloran:
see also #502 about making the app messages prettier

hamilton:
@bcolloran turns out this entire idea is a nonstarter.

https://blog.github.com/2018-02-18-deprecation-notice-removing-anonymous-gist-creation/

bcolloran:
ok guys, i've removed the actual code about gists, but b/c this PR had a bunch of other changes that did not seem controversial, i'm going to merge

PR #507, opened 2018-03-19
Save notebook before running cells

### LINKED ISSUE #461: possibly unneeded peek at state? (maybe need small refactor of `saveNotebook`)
just noticed this in the code here:

https://github.com/iodide-project/iodide/blob/e1520924cdb9326aa01ed8c9f1957f0d6eafb3cf/src/task-definitions.js#L259

```
  253  
  254: tasks.saveNotebook = new UserTask({
  255    title: 'Save Notebook',
  256    keybindings: ['ctrl+s', 'meta+s'],
  257    displayKeybinding: commandKey('S'),
  258    preventDefaultKeybinding: true,
  259:   callback() { dispatcher.saveNotebook(store.getState().title) },
  260  })
  261  
```
I think we should not need that `store.getState().title` for any reason -- the reducer will have access to the store as it's processing the `saveNootbook` action.

@hamilton, do you recall any reason why this was needed? Even if it is currently needed, a small refactor of `saveNotebook` should be able to polish it out (I can take that on if it's needed)

### LINKED ISSUE #503: always autosave before executing cell
in case you hit a chunk of long running code or something that hangs and you need to bail out and reload the page, we should make it easy to recover the code you had been using previously.

i think this should be pretty easy to implement, probably just adding an additional `saveNotebook` dispatch during the cell evaluation user task

pradeepgangwar:
Fixes #503 
Fixes #461

### DIFF
--- a/src/reducers/notebook-reducer.js
+++ b/src/reducers/notebook-reducer.js
@@ -92,11 +92,13 @@ const notebookReducer = (state = newNotebook(), action) => {
     }
 
     case 'SAVE_NOTEBOOK': {
+      ({ title } = state)
       let lastSaved
       if (!action.autosave) {
         lastSaved = (new Date()).toISOString()
       } else {
         ({ lastSaved } = state)
+        title = AUTOSAVE + title
       }
       nextState = Object.assign({}, state, { lastSaved }, {
         cells: state.cells.slice().map((c) => {
@@ -107,13 +109,8 @@ const notebookReducer = (state = newNotebook(), action) => {
           }
           return newC
         }),
-      }, { title: state.title === undefined ? 'new notebook' : state.title })
+      }, { title: state.title })
       clearHistory(nextState)
-      if (action.title !== undefined) {
-        ({ title } = action)
-      } else {
-        ({ title } = state)
-      }
       window.localStorage.setItem(title, stringifyStateToJsmd(nextState))
       return Object.assign({}, state, { lastSaved }, getSavedNotebooks())
     }
--- a/src/task-definitions.js
+++ b/src/task-definitions.js
@@ -34,8 +34,8 @@ tasks.evaluateCell = new UserTask({
 
   callback() {
     dispatcher.changeMode('command')
-    dispatcher.evaluateCell()
     dispatcher.saveNotebook(true)
+    dispatcher.evaluateCell()
   },
 })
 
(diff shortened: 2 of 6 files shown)


bcolloran:
Thanks for giving this a try @pradeepgangwar.

It does save to local storage for "ctrl+enter" and "run all", but it looks like it still needs to be implemented for "shift+enter". Here's what i tried:
- open a new dev notebook at `dev/iodide.dev.html`
- change the NB title to something i haven't used before: "newname1"
- run the first cell with "shift+enter"
- check `Object.keys(localStorage)`
- the title "newname1" does not appear in the list of saved items.

Also, this should update the *autosave* not the normal "save" entry in localStorage.

finally, if you want to try something more ambitious, you might also be able to fix #461 while you're looking at stuff related to saving :-) (but of course you don't have to try for that bigger task if you are not interested).

Thanks!

pradeepgangwar:
@bcolloran I have updated the files and they work fine on my local machine. Evaluating cells changes autosave entry and normal 'save' entry remains untouched. Please review :smiley:

pradeepgangwar:
@bcolloran After latest commit on this PR I also tried addressing issue #461. Please have a look into it. I found it to be working on my machine.

src/reducers/notebook-reducer.js
```diff
@@ -89,11 +92,6 @@ const notebookReducer = (state = newNotebook(), action) => {
         }),
       }, { title: state.title === undefined ? 'new notebook' : state.title })
```
bcolloran: after your other nice refactoring, do we still need `{ title: state.title === undefined ? 'new notebook' : state.title }`? Can the title ever be undefined at this point? If not, we can remove all the checks for that, and we can even update the state schema in state-prototypes.js, the fixme around line 80:
`title: {}, // FIXME change to string ONLY with default '' or 'untitled'`

pradeepgangwar:
@bcolloran I have added all changes suggested by you in previous comments. Please review :smiley: 
I am unable to figure out why these tests are failing? :worried:  I haven't changed anything related to those files.

src/task-definitions.js
```diff
@@ -35,13 +35,17 @@ tasks.evaluateCell = new UserTask({
   callback() {
     dispatcher.changeMode('command')
     dispatcher.evaluateCell()
+    dispatcher.saveNotebook(true)
```
bcolloran: oops, we need to autosave _before_ we evaluate the cell.

Issue #508, opened 2018-03-19
changing title throws an error in the console


bcolloran:
when you try to edit the title and then leave the title text entry area, the following error is thrown:

```
editor-mode-title.jsx:24 Uncaught TypeError: __WEBPACK_IMPORTED_MODULE_4__task_definitions__.a.changeMode is not a function
    at TitleUnconnected.onBlur (editor-mode-title.jsx:24)
    at HTMLUnknownElement.boundFunc (ReactErrorUtils.js:63)
    at Object.ReactErrorUtils.invokeGuardedCallback (ReactErrorUtils.js:69)
    at executeDispatch (EventPluginUtils.js:83)
    at Object.executeDispatchesInOrder (EventPluginUtils.js:106)
    at executeDispatchesAndRelease (EventPluginHub.js:41)
    at executeDispatchesAndReleaseTopLevel (EventPluginHub.js:52)
    at Array.forEach (<anonymous>)
    at forEachAccumulated (forEachAccumulated.js:22)
    at Object.processEventQueue (EventPluginHub.js:252)
```

A fix for this should ideally add an appropriate test case.

PreciseSlice:
Hi there! I’m a first-time contributor and was hoping to help out with this issue. I am with a group of students looking to contribute. I noticed nobody was assigned to this issue, but if there’s already a solution in progress I’m happy to try helping out elsewhere. Thanks!

hamilton:
Thanks for the PR @PreciseSlice & Co. ~~~

PR #510, opened 2018-03-20
#508 Editor Mode On Blur Bug

cbdallavalle:
When the title was changed and the user clicked out of that text area, a webpack error was thrown. 

We discovered that in the `editor-mode-title.jsx` component, line 24 caused the error, which stated that `tasks.changeMode` was not defined. `Tasks.changeMode` called an object instead of a function. 

Line 24 is now fixed with the addition of `.callback()` so that the object's method is called and subsequently executed. 

Solves #508

Authors: 
@PreciseSlice 
@michellehoffman 
@cbdallavalle

### DIFF
--- a/src/components/toolbar/editor-mode-title.jsx
+++ b/src/components/toolbar/editor-mode-title.jsx
@@ -21,7 +21,7 @@ export class TitleUnconnected extends React.Component {
   }
 
   onBlur() {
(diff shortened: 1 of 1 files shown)


hamilton:
Great! This works as expected / easy fix. Thanks for the PR y'all.

Issue #497, opened 2018-03-18
the "last saved" text section clashes a bit with icon button


hamilton:
![mar-18-2018 07-35-09](https://user-images.githubusercontent.com/95735/37567091-0eb32382-2a7f-11e8-8a03-96fb04771fed.gif)

michellehoffman:
Hi there! I’m a first-time contributor and was hoping to help out with this issue. I noticed nobody was assigned to it, but if there’s already a solution in progress I’m happy to try helping out elsewhere. Thanks!

hamilton:
Hi @michellehoffman! It's all yours!

PR #512, opened 2018-03-21
#497 solved clash of text with icon button

michellehoffman:
On hover, the Declared Variables button had a border appear that overlapped with the "last saved" text. To fix this issue, we added a style attribute to the Typography component in last-saved-text.jsx and assigned it a right margin of 10px.

Before: 
497-unsolved

After:
497-solved

Authors: 
@PreciseSlice 
@cbdallavalle 
@michellehoffman


bcolloran:
Great, many thanks @PreciseSlice @cbdallavalle and @michellehoffman :-)

PR #513, opened 2018-03-21
#509 Autosaved and saved notebook names now updating in menu

### LINKED ISSUE #509: the autosaved notebook name does not appear to update on autosave
### What I Did

- change the title of a notebook to one that has never been saved before
- wait until the next autosave action appears in the console
- look in the "notebooks" menu

### What I Expected

the new title should appear next to "autosave"

### What Happened 

no new title appears

---

My guess from a quick look is that the menu is generated once by looking at localStorage when Iodide loads rather than having this info in the store so that redux knows to automatically update the menu as occurs for other items in the store. it might be that the list of autosaved NB titles should be added to the store at load, and then updated on each (auto)save.

jandrewtorres:
Fixes #509.

### Before:
Autosaved and saved notebooks would not be updated in "notebooks" menu after saving, only upon refresh.

### After:
Autosaved and saved notebooks are now updating in "notebooks menu" after saving.

### How this was accomplished:

1. Two fields were added to the stateSchema in state-prototypes.js: `autoSave` and `locallySaved`
1. These fields are updated accordingly in the notebookReducer actions.
1. The SavedNotebooksAndExamplesSubsection retrieves the values from the store and maps them to props and renders.

### Thoughts
This is a quick fix to get the bug fixed, ideally the saved notebooks list would not be in the notebook reducer. It maybe would be good to have a toolbar, or an application 'workspace' reducer/state to hold this list.


src/components/toolbar/saved-notebooks-and-examples-subsection.jsx
```diff
@@ -1,47 +1,63 @@
...
+  }
 
-export default class SavedNotebooksAndExamplesSubsection extends React.Component {
   render() {
+    const autoSaveMenuItem = this.getAutoSaveMenuItem()
+    const locallySavedMenuItems = this.getLocallySavedMenuItems()
+
     return (
       <NotebookMenuSubsection title="notebooks ... " {...this.props} >
         {autoSaveMenuItem ? <NotebookMenuHeader key="autosave" title="Auto-Saved" /> : undefined}
```
bcolloran: as long as you're in here, you maybe you could replace this 
```
( x ? a )
( x || b )
( x ? c )
```
pattern with more explicit if/else blocks

:-)

src/components/toolbar/saved-notebooks-and-examples-subsection.jsx
```diff
@@ -1,47 +1,63 @@
...
+        return p(b) - p(a)
+      })
 
-locallySaved.sort((a, b) => {
-  const p = (_) => {
-    let ls = localStorage.getItem(_)
-    if (!ls) return -1
-    ls = stateFromJsmd(ls)
-    return Date.parse(ls.lastSaved)
+      if (this.props.locallySaved.length) {
```
bcolloran: is this test needed? isn't props.locallySaved always an array of strings?

src/components/toolbar/saved-notebooks-and-examples-subsection.jsx
```diff
@@ -1,47 +1,63 @@
...
-let autoSaveMenuItem
-let autoSave
-if (autosave !== undefined) {
-  autoSave = getLocalStorageNotebook(autosave)
-  autoSaveMenuItem = <NotebookMenuItem task={autoSave} key={autoSave.title} />
-}
+  getLocallySavedMenuItems() {
+    let locallySavedMenuItems
+    if (this.props.locallySaved) {
+      this.props.locallySaved.sort((a, b) => {
```
bcolloran: can we move all this sorting stuff to the `getSavedProjects()` function in the notebook-reducer file? In general we're moving towards having our React components be as dumb as possible (makes them much easier to test), so ideally we'll just pass in the full sorted list as props to the component and render it. We should only ever have to sort this list when it changes, which we can do on load and then in subsequent reducer actions


see if you can get

move as much of `getLocallySavedMenuItems()` functionality out of the component as possible.

src/components/toolbar/saved-notebooks-and-examples-subsection.jsx
```diff
@@ -1,47 +1,63 @@
...
-})
-
-let locallySavedMenuItems
-if (locallySaved.length) {
-  locallySavedMenuItems = locallySaved.map((l) => {
-    const task = getLocalStorageNotebook(l)
-    if (task !== undefined) {
-      return <NotebookMenuItem task={task} key={task.title} />
+
+  getAutoSaveMenuItem() {
```
bcolloran: this function is probably not needed.

you can replace it and the block
```
        {autoSaveMenuItem ? <NotebookMenuHeader key="autosave" title="Auto-Saved" /> : undefined}
        {autoSaveMenuItem || undefined}
        {autoSaveMenuItem ? <NotebookMenuDivider key="autosave-divider" /> : undefined }
```

with something simpler in the render function like:

```
let autoSaveMenuItem
if (this.props.autoSave) {
autoSaveMenuItem = (<div> // FIXME: use React 16 fragments instead of useless container div
  <NotebookMenuHeader key="autosave" title="Auto-Saved" />  
  <NotebookMenuItem task={autoSave} key={autoSave.title} />
  <NotebookMenuDivider key="autosave-divider" />
</div> ) 
}
```

Now that i think of it, you can probalby do the same for getLocallySavedMenuItems()

(PS: i know that you are are just following the existing patterns, so not criticizing your PR -- i just didn't notice this stuff on the first code review, and now i'm seeing how we can clean this stuff up, so we may as well do it :-)  )

bcolloran:
These code changes look good to me. Tagging @hamilton for review since he's worked on the menus and the saving stuff. Great work on this so far @jandrewtorres!

@hamilton, i've done code review, but not QA yet. dunno if there are parts of the menu/save behavior you want to poke at.

jandrewtorres:
@bcolloran All suggestions you have made have been implemented :-). Thank you for the advice!

bcolloran:
Merged! thanks @jandrewtorres !

PR #514, opened 2018-03-22
eval migrated out of cell-reducer and into actions

bcolloran:
this PR pulls cell evaluation out of the reducers and into the actions, which will open the path for evaling plugin definition cells like any other cell and having that eval action dispatch other actions that update the global store. i've tested this on a number of example notebooks and everything seems to be working ok, but um since evaling is the core of the whole shebang, probably worth doing some more  thorough QA :-)

(also i messed up and left in a bit of gist code previously, so the removal of that code is in this PR too)


bcolloran:
great, thanks for reviewing @hamilton.

PR #518, opened 2018-03-24
#475 Added insert cell above to cell-menu-subsection.jsx

jandrewtorres:
Issue #475 mentions this addition. 

A quick one-liner 🥇. 

I wanted to use this feature frequently when I was having some fun creating an example notebook tonight and it would have been helpful if it was in the menu, so now it is.


bcolloran:
Great, thanks!

PR #519, opened 2018-03-26
Code plugin cell

bcolloran:
As of this PR it's you can eval a "language definition cell" (kbd shorcut "L" to change to that type), which, if it contains a correctly formatted JSON string matching this spec --

```
{
  "languageId": "py",
  "displayName": "python",
  "codeMirrorMode": "python",
  "keybinding": "p",
  "url": "https://raw.githubusercontent.com/iodide-project/pyodide-demo/master/pyodide.js", 
  "module": "pyodide", 
  "evaluator": "runPython"
}
```

-- will download the JS file at the given URL, will attach the keybinding, and will call
`window[module][evaluator]` on the code in the cells of this type when you eval them

you should be able to test this with a dumb example using the following gist with your dev build on this branch:

`/dev/iodide.dev.html?gist=bcolloran/40e4677559cd19eab9df06c7926a576b`

@mdboom does this do roughly what you need? One thing i'm not sure about: do the python builds require multiple files? If so, that should be much easier to add now that we have this all in place.

Next steps:
- add per-cell badge that shows cell type
- add new cell types to mouse menu
- make per-cell menus available from badge, and allow selection of cell type there too

EDIT (by @mdboom): Changed JSON snippet above to a working example to avoid confusion.


src/actions.js
```diff
@@ -230,6 +326,8 @@ export function evaluateCell(cellId) {
       dispatch(evaluateResourceCell(cell))
     } else if (cell.cellType === 'css') {
       dispatch(evaluateCSSCell(cell))
+    } else if (cell.cellType === 'plugin') {
+      dispatch(evaluateLanguagePluginCell(cell))
```
mdboom: I'd anticipate we'll have more plugin types than just language plugins (output plugins, for example).  I don't think we want a proliferation of cell types for each type of plugin.  Should we add an `evaluatePluginCell` function that dispatches to different known plugin types based on the content of the JSON?  I.e. add a field "pluginType" to the JSON that would be "language" in this case, but might be "output", etc.?  Maybe we want to move the rest of the payload under a "settings" key in that case, because it will be easier to validate against a JSON Schema that way (but that's fine if not).

bcolloran: yeah i think that's right. minimally, we already know we want to support MD extensions. I'll go make that change.

bcolloran:
"UX is not great" is awfully generous -- the UX for this is terrible :-)

+1 to all your suggestions -- directory of plugins, and also a nicer ux where you can at least select from valid pluginTypes and you'll be given boxes to fill in to correctly specify the plugin (and help info, etc).

If this works for what you need (esp wrt loading everything from a single URL), i'll merge.

PR #530, opened 2018-04-01
Cell type label

bcolloran:
this creates cell type labels, which initially was ok, but when i tried to add a tooltip it turned out to be a PITA b/c parts of our css were a mess. So i had to clean up some stuff there (moved all the cell and cell container styles to a separate file), which also meant i had to tweak cell rows a bit (got rid of most of the flex container stuff). And as long as i was messing with rows, i added tooltips to the row collapse buttons.

I think this one is a pretty straightforward review, but since i mucked about the css, please verify that your dashboards look ok

this clears the way to add a per-cell menu, and will also actually make it easier to add the "row visible in pres mode" toggle button


hamilton:
Hey - this looks good so far (as a quick-and-dirty cell type label), but all the other changes breaks the CSS for my recent work. The standard stylesheet looks good, so most reports should be fine. I'm happy to change the stylesheet there to cohere with these changes (obviously more sensible), but before we merge this change I need a bit more time on my end to make sure the changes will sync up in a way that won't disrupt the internal work that relies on master.

This does surface a need: I don't want to point to master for a production notebook (seems dangerous if we're going to do stuff like this without having any release schedule / release notes), but we don't really have a nice, clean way of pointing to a specific iodide version, and I think this will be a usability problem fairly soon, so perhaps we should start more aggressively versioning the notebook releases.

bcolloran:
dang, i was afraid this would break your dashboards, but i was holding out hope that it would be ok :-( i'll certainly hold off on merging until you are ready.

+1 re: now that we're rolling out production work, it's time to start implementing the versioning plan we sketched out a while back. i'll take a look at that today.

re: "quick and dirty" -- i actually like this ui :-/ (and building it was far from quick -- css morass, lol :-) ). I think that it's understated and doesn't add a lot of visual clutter, but that it does provide the at-a-glance hint about cell types that we've been missing, and that it will make a good launch point for the cell menus. Could you expand on what you dislike about it?

hamilton:
Sorry, I didn’t mean to imply that the ui was bad! I think it’s great and
we can expand a menu out of it to allow cell operations like changing the
cell type. This is totally in the right direction and I can also change my
work to cohere w/ your css first thing tomorrow.

bcolloran:
oh ok, no worries :-) 

no rush on merging this (i'm underwater on other stuff today, so i won't be needing to build off this in the next couple days) so take all the time you need to make sure your dashboards work+look how they need to.

src/style/cell-styles.css
```diff
@@ -0,0 +1,184 @@
...
+
+div.HIDDEN div.collapse-button {
+    width: calc(100%);
+    /* when collapsed, make the button visible*/
+  background: #fdfdfd;
+  outline: 1px solid #ddd;
+  cursor: pointer;
+}
+
+div.main-component {
```
hamilton: This style selection can also be accompanied with the following presentation mode override in the presentation stylesheet, toward the top:

```css
/* in the editor view, there is 24px less room for the collapse-button. Let's get it back. */
div#cells.presentation div.main-component {
  width: 100%;
}
```

To make it so the `main-component` width is 100%, not `calc(100%  - 24px)`. This will make it easy to center / change the presentation style for users who are mucking around with the width of the main components.

src/style/cell-styles.css
```diff
@@ -0,0 +1,184 @@
...
+  cursor: pointer;
+  padding: 0px 6px 0px 0px;
+}
+.cell-type-label:hover {
+  background: #fbfbfb;
+  outline: 1px solid #ddd;
+  color: hsl(275, 4%, 40%);
+}
+
+div.cell-row-container {
```
hamilton: This style can be accompanied with the following override in the presentation stylesheet:

```css
/* this resets the div.cell-row-container to not have a margin for collapse + cell label buttons,
which is an additional 50px.
*/
div#cells.presentation div.cell-row-container {
  width: 100%;
}
```

This is actually an issue right now in the presentation mode. The `div.cell-row-container` elements all have -50px in width, when I think users would expect them to be all 100% wide.

Easy fix though! Just put that declaration toward the top of the presentation mode css.

src/components/cell-menu-container.js
```diff
@@ -0,0 +1,66 @@
...
+  if (cell.cellType === 'code') {
+    label = cell.language
+  } else if (cell.cellType === 'markdown') {
+    label = 'md'
+  } else if (cell.cellType === 'css') {
+    label = 'css'
+  } else if (cell.cellType === 'plugin') {
+    label = 'plugin'
+  } else if (cell.cellType === 'external dependencies') {
+    label = 'resource'
```
hamilton: I notice that cells with "respource" overflow a bit. I bet we could do one of these two things:

1.) change `resource` to `res` - we are already using md, js, etc. and maybe `ext` or `dep` clear.
2.) change the font size defined for `cell-type-label` to `12px`. This is still legible, and everything eems to fit ok.

hamilton:
@bcolloran this is ready for your review. Here are the commits that I changed:

- b23f386 fixes the presentation css by overriding the widths of `div.cell-row-container` and `div.main-component` to both be 100%, instead of `calc(100% - some value)`
- 1f1fdea is a series of small tweaks based on our conversation - changed the width of the menu to be 64px to accomodate longer strings, changed the height of the button to be the same height as the editor when it is one line, nudged over the collapse button by 2px so the spacing between the menu, collapse, and editor are all equal now, fixed a small css issue regarding `:first-child` in the presentation view (so now you don't get a scroll bar in presentation by default)

PR #536, opened 2018-04-06
CSS margin fix for presentation mode for :first-child

hamilton:
I mis-selected the actual `:first-child`. This is a small fix that should patch things up for notebooks. @bcolloran I'm happy to just merge this if you don't think it is worth a review. It's a fairly minor css change.


bcolloran:
i'm sure it's aces. merged.

PR #537, opened 2018-04-06
Reorganizing files

bcolloran:
our biggest PR ever.

i think the great refilening is complete. tests are green, i've poked at it with many example NBs and many (but not all) kbd and menu actions, but since so many things are being moved around here, this deserves some thorough QA, so please poke at it a bit and merge whenever you're ready.

probs best to get to this one as soon as you can so that other folks don't fall out of sync too badly...


bcolloran:
fixes #395

Issue #521, opened 2018-03-27
[META] "pinned cells" side pane


hamilton:
Large notebooks are difficult to navigate. Sometimes, for example, a helper function defined in a cell toward the top of a notebook needs one small edit so that a cell much farther down the notebook can support another argument. Right now I have to scroll to find the cell Another example: CSS cells end up having to be made near the code block that draws a viz, but this makes it difficult to generalize the css for one viz for others, since you end up having to go hunt for another CSS cell for a separate part of the presentation. If I commit to making my CSS cell only at the top of the document, I would have to scroll all the way up to the top of the notebook to make a small tweak, then scroll all the way back down to my code block to continue work. 

Pinned cells in a side pane would solve these types of issues gracefully while maintaining the cell-style notebook workflow.

The work looks like this:

- [x] migrate to side panes that do not overlay cells (#520)
- [ ] fix header element in a side pane to not scroll with the rest of the side pane content, so the close button / title is always visible (#521)
- [x] move the close button (x) on the side pane to the right instead of left, to prevent accidental side pane closure when clicking on the main cell view (#522)
- [ ] change the state prototypes for cells to include a "pinned" key-value pair (maybe this is a pin id so we can preserve the pin order - @bcolloran we can either do this or just a boolean, any thoughts here? If we don't preserve pin order, we can preserve actual cell order) (#525)
- [ ] create a new action to pin a cell using the cell state addition above (#526)
- [ ] create a new side pane called PinnedCellsSidepane (#527)
- [ ] display the cells w/ the a restricted set of actions - (#527)
   - execute cell
   - unpin cell
   - change cell type
- [ ] add a button per cell that allows a user to unpin a cell in the side pane (#527)
- [ ] add a menu item for 'pin cell' (#528)
- [ ] add a keybinding for pinning a cell using the `UserTask` class (`meta+option+p` may be a useful one, but perhaps @wlach can guide us on this so as not to clash with editors) (#528)
- [ ] __BONUS__ include a way to scroll to the pinned cell in the main cell content (outside of the pane)

djbarnwal:
Hi @hamilton, I would like to work on this issue.

hamilton:
Hey @djbarnwal - that'd be great. It might be good to tackle the issues in order w/ a PR, but I'll leave it to you to appropriately break up the work so PRs are reviewable.

Feel free to ping me or @bcolloran if you have any questions while working on it.

bcolloran:
[...]

PR #541, opened 2018-04-07
change state-prototypes for pinning cells

### LINKED ISSUE #525: change the state prototypes for cells to include a "pinned" key-value pair
Part of #521. Maybe this is a pin id so we can preserve the pin order - @bcolloran we can either do this or just a boolean, any thoughts here? If we don't preserve pin order, we can preserve actual cell order.

pradeepgangwar:
I am not sure if this is the correct fix. Please let me know if it needs some changes.

Fixes #525 
Part of #521

### DIFF
--- a/src/actions/task-definitions.js
+++ b/src/actions/task-definitions.js
@@ -89,6 +89,13 @@ tasks.moveCellDown = new UserTask({
 })
 
 
+tasks.pinCell = new UserTask({
+  title: 'Pin this cell',
+  callback() {
+    dispatcher.pinCell()
+  },
+})
+
 tasks.selectUp = new UserTask({
   title: 'Select Cell Above',
   displayKeybinding: 'Up', // \u2191',
--- a/src/state-prototypes.js
+++ b/src/state-prototypes.js
@@ -37,6 +37,7 @@ const cellSchema = {
       type: 'string',
       enum: cellTypeEnum.values(),
     },
+    pinned: { type: 'boolean' },
     value: {}, // empty schema, `value` can be anything
     rendered: { type: 'boolean' },
     selected: { type: 'boolean' },
--- a/src/state-prototypes.js
+++ b/src/state-prototypes.js
@@ -37,8 +37,8 @@ const cellSchema = {
       type: 'string',
       enum: cellTypeEnum.values(),
     },
-    pinned: { type: 'boolean' },
     value: {}, // empty schema, `value` can be anything
+    pinned: { type: 'boolean' },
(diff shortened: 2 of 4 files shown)


src/state-prototypes.js
```diff
@@ -73,6 +74,10 @@ const stateSchema = {
       type: 'array',
       items: cellSchema,
     },
+    pinnedCells: {
+      type: 'array',
+      items: cellSchema,
```
djbarnwal: I think this would create redundancy. We would be storing information about pinned cells at 2 places (`cells` and `pinnedCells`). The idea was to either store a boolean in each cell or make a pinnedCell key value pair where we just store the cell ids of those cells which are pinned.

pradeepgangwar: Yes, I was also thinking this, but I added both. 
I think @hamilton can guide us on this.

pradeepgangwar:
@bcolloran I made the changes. Should I add more changes to this pr (as this is a small pr)? or get this merged and try with new PRs. :sweat_smile:

djbarnwal:
@pradeepgangwar You would also need to update the `newCell` function to include the default value of `false` when a new cell is created.

bcolloran:
+1 @djbarnwal's comment

I think you're right that this is too small of a unit of code to merge all at once. Let's try to wrap this into one PR that has the full functionality of the feature we're trying to implement.

If you'd like to push your feature branch to the main repo, perhaps other people will be able to help you implement some of that functionality.

pradeepgangwar:
@bcolloran I have added basic code to action.js for pin cell action. Please let me know if I am in the right direction so that I can move forward with implementation.

src/actions/task-definitions.js
```diff
@@ -89,6 +89,13 @@ tasks.moveCellDown = new UserTask({
 })
 
 
+tasks.pinCell = new UserTask({
+  title: 'Pin this cell',
```
bcolloran: maybe change the title to "pin this editor"

hamilton: I think "pin this editor" could be confusing too, tbh, since all the cells in sum are the editor. "pin this cell editor" is probably what we want.

hamilton:
Realizing I ran into a different bug before I could get into this PR - will review over the weekend!

bcolloran:
@pradeepgangwar how's it going on this? let us know if you need any help.

pradeepgangwar:
@bcolloran I got busy with my examination here :disappointed: . I will update as soon as I get time.

PR #547, opened 2018-04-10
Cell menu

bcolloran:
prereq for run-all flow controls. also includes some css cleanup (eliminating dead styles) and a couple other very minor touch ups.


bcolloran:
ooh, nice catch. i've disabled the menu option to switch to the current cell type. merging.

PR #550, opened 2018-04-11
Skip during run all

bcolloran:
@hamilton, this implements the UI and state bits for the skipping cells during run-all, so you should be able to plug this into the stuff you're doing, it should just be a simple check to see if the cell to be evaled has `includeInRunAll === true`

regarding the UI: i tried a number of options to indicate that a cell will be skipped during run all (a couple icons, plain grey text (italics and not), and a couple variants of the little bootstrap warning-ish pill that i ended up settling on. icons looked awkard and confusing, and this seemed to provide the best and clearest visual warning that _hey there is something weird going on with this cell that you need to be aware of_ without being too distracting and cluttery. (little icons were actually worse).

in the process, i also remember why i'd set the hover box height of the cell-label to 18px-- that matches the height of the collapsed cell rows. i needed to recover a bit of that space for the "skip" label, so i set it back to that. i think this is probably ok -- i was with you on wanting to make the cell label the same height as an empty editor, but i think that it will typically be the case that editors don't remain empty for long, and the shorter button has better visual balance with the collapse states and the `skip` warning

i punted on a kbd shortcut for for toggling skip state. this action may not need it. i haven't yet implemented "skip all cells above", but i'd imagine an action like that will be how this functionality is normally accessed.

once this is merged, i'll be ready to get started on freezing state :tada:

### DIFF
--- a/src/actions/task-definitions.js
+++ b/src/actions/task-definitions.js
@@ -198,20 +198,12 @@ tasks.changeToPluginCell = new UserTask({
   callback() { dispatcher.changeCellType('plugin') },
 })
 
-tasks.includeCellInRunAll = new UserTask({
-  title: 'Include Cell During Run All',
-  // keybindings: ['l'],
-  // displayKeybinding: 'L',
-  // keybindingPrecondition: isCommandMode,
-  callback() { dispatcher.setCellIncludeInRunAll(true) },
-})
-
-tasks.skipCellInRunAll = new UserTask({
-  title: 'Skip Cell During Run All',
-  // keybindings: ['l'],
-  // displayKeybinding: 'L',
-  // keybindingPrecondition: isCommandMode,
-  callback() { dispatcher.setCellIncludeInRunAll(false) },
+tasks.toggleSkipCellInRunAll = new UserTask({
+  title: 'Toggle Skipping Cell in Run All',
+  keybindings: ['s'],
+  displayKeybinding: 'S',
+  keybindingPrecondition: isCommandMode,
+  callback() { dispatcher.setCellSkipInRunAll() },
 })
 
 tasks.changeMode = new UserTask({
--- a/src/actions/task-definitions.js
+++ b/src/actions/task-definitions.js
@@ -29,7 +29,7 @@ const commandKey = () => (OSName === 'MacOS' ? '⌘' : 'Ctrl')
 const tasks = {}
 
 tasks.evaluateCell = new UserTask({
-  title: 'Evaluate Cell',
+  title: 'Run Cell',
   keybindings: ['mod+enter'],
 
   callback() {
@@ -40,8 +40,8 @@ tasks.evaluateCell = new UserTask({
 })
 
 tasks.evaluateAllCells = new UserTask({
-  title: 'Evaluate All Cells',
-  menuTitle: 'Evaluate All Cells',
+  title: 'Run All Cells',
+  menuTitle: 'Run All Cells',
   callback() {
     dispatcher.saveNotebook(true)
     evaluateAllCells(getCells(), store)
@@ -196,6 +196,22 @@ tasks.changeToPluginCell = new UserTask({
   displayKeybinding: 'L',
   keybindingPrecondition: isCommandMode,
   callback() { dispatcher.changeCellType('plugin') },
+})
+
+tasks.includeCellInRunAll = new UserTask({
+  title: 'Include Cell During Run All',
+  // keybindings: ['l'],
+  // displayKeybinding: 'L',
+  // keybindingPrecondition: isCommandMode,
+  callback() { dispatcher.setCellIncludeInRunAll(true) },
+})
+
+tasks.skipCellInRunAll = new UserTask({
+  title: 'Skip Cell During Run All',
+  // keybindings: ['l'],
+  // displayKeybinding: 'L',
+  // keybindingPrecondition: isCommandMode,
+  callback() { dispatcher.setCellIncludeInRunAll(false) },
 })
 
 tasks.changeMode = new UserTask({
--- a/src/components/cells/cell-menu-container.js
+++ b/src/components/cells/cell-menu-container.js
@@ -4,7 +4,6 @@ import { bindActionCreators } from 'redux'
 import PropTypes from 'prop-types'
 import Tooltip from 'material-ui/Tooltip'
 import Menu from 'material-ui/Menu'
-// import LowPriority from 'material-ui-icons/LowPriority'
 
 import CellMenu from './cell-menu'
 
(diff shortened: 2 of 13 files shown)


src/actions/task-definitions.js
```diff
@@ -29,7 +29,7 @@ const commandKey = () => (OSName === 'MacOS' ? '⌘' : 'Ctrl')
 const tasks = {}
 
 tasks.evaluateCell = new UserTask({
-  title: 'Evaluate Cell',
```
hamilton: is there a reason we're moving from "evaluate" to "run" here?

src/actions/task-definitions.js
```diff
@@ -198,6 +198,22 @@ tasks.changeToPluginCell = new UserTask({
   callback() { dispatcher.changeCellType('plugin') },
 })
 
+tasks.includeCellInRunAll = new UserTask({
```
hamilton: I do like that these are explicitly two tasks, but another possibility is to simply have this toggleable, which would make possible the keybinding `L` or something like that (maybe `S` for "skip" is a good key, but I'm worried about folks who were taught to keep their left hands on ASDF (like me) would accidentally hit it, but whatever)

src/components/cells/cell-menu-container.js
```diff
@@ -35,6 +38,23 @@ export class CellMenuContainerUnconnected extends React.Component {
 
   render() {
     const { anchorElement } = this.state
+    const includeInRunAllIndicator = (
```
hamilton: I think this is cool, and I might propose some pretty minor style changes to tighten it up, but I think this is the right direction. re: the cell menu, totally cool for it to not be that height, so long as the rest of it aligns well (border vs. outline, etc) which it does now. I agree that cells don't stay empty for long, but we do have a first impression to make.

src/actions/task-definitions.js
```diff
@@ -198,6 +198,22 @@ tasks.changeToPluginCell = new UserTask({
...
+  title: 'Include Cell During Run All',
+  // keybindings: ['l'],
+  // displayKeybinding: 'L',
+  // keybindingPrecondition: isCommandMode,
+  callback() { dispatcher.setCellIncludeInRunAll(true) },
+})
+
+tasks.skipCellInRunAll = new UserTask({
```
hamilton: philosophical question - should this be `skipCellInRunAll` or should it be `skipCellDuringEvaluation`? If I have selected a cell I'm skipping and attempt to evaluate it, should it evaluate? I'd argue probably not. If I want to not have a cell evaluate when I click "run all cells", I don't want to be in the position to accidentally evaluate it, either.

OR, should this be `disableCell`, which makes it 1. uneditable, and 2. doesn't run in any context. This option feels the most consistent and should be clear to users that a disabled cell doesn't run in any context, nor is it editable until it is enabled.

bcolloran:
thanks for the review @hamilton 

"""I am noticing that the cell execution doesn't skip when I click "run all cells" - was this PR in-place to implement that as well?"""

Nope, i haven't touched that cuz i know you're in the middle of working on the run/run all code, so i don't want to introduce merge conflicts

"""I'm also noticing that the skip-cell state isn't saved when I export my notebook. Is that intentional / is that work to be tackled later?""

oops, that was an oversight. i'll add that now. But i am going to make a big semantic change, in part to keep the JSMD tight-- i'm going to change the cell property from `includeInRunAll` to `skip`, which will be easier to type in the direct jsmd editor universe, and which will be the deviation from the default behavior (jsmd only serializes changes from default). that key is a little less explicit, but i think it'll be ok.

"""I have an inline comment about whether we should be disabling cells vs. the "skip all" nomenclature, and what that actually means. I am in favor of disabling cells because I think that is a much clearer paradigm than marking a cell as "skip during run all". Disabled cells can't be run. Open to discussing this a bit more, but I feel it makes sense to get this part right so it would be worth talking. We can also make the UI a bit clearer with disabled cells by lightly changing the opacity a bit and using a disabled label if we took that approach.
"""

i think this feature is basically targeted at the publication use case, and primarily to be used in concert with the "environment" cell that will be used for freezing state. i don't actually think that disabling a cell is the right thing-- there are good reasons that that a user would want to execute a cell by hand while experimenting in a notebook, but would want to mark the cell as a "skip" for publication/run all, and asking the user to toggle back and forth seems like a headache and even maybe a source of error (if you forget to re-toggle the cell to "skip")

""" re: collapsed cells and visual design, I think that visually another approach is to have a collapsed cell be the same height as a single-line cell, for visual balance, but this is something we can talk about in-person and hash it out, and at any rate I think your approach works. As someone with a visual brain there are probably some small UI tweaks we can make down the line to make sure everything feels a bit more consistent and we are utilizing the space well. That all said, we don't need to bikeshed over a UI that does what we need it to do right now :)
"""
ya i played with is some -- that actually makes collapsed cells look awkwardly big and chunky (obvi the point of collapsing a cell is to make it take up very little space).

making collapsed cells 1/2 the height of a single-line editor could work, but then to balance with the blocks of equal height for the cell label hover and the skip indicator, we'd have to shrink those down to the point where they wouldn't be readable (we're already on the edge of that threshold).

overall, i think the constraint to relax is the one around the single-line editor, but we can play with it more. don't fret, i have a visual brain too, and it's been turning on this stuff also :-)

"""You always seem to touch exactly 9 files in your recent PRs!"""

my lucky number! :-)

src/actions/task-definitions.js
bcolloran: couple reasons:
- jupyter uses the "run" terminology
- "run" fits better in menus and tooltip
- it's a little less wordy

fwiw: i think this is mostly for user-facing stings, and that using "evaluate" within the iodide code base is good

src/actions/task-definitions.js
bcolloran: sure, i can make this task a toggle

"S" is fine with me -- i can add that if you wish. I'm not sure about the balance between the value of adding shortcuts and the cost of making the kbd a minefield of accidental actions (as you say). we may not be ther yet, so perhaps "S" is fine? i don't have a strong intuition on this

bcolloran:
re the JSMD -- actually, this is a more general issue. Maybe i'll just add a mapping that translates from "skip" to the full state property name. That machinery will be good to have in place.

src/actions/task-definitions.js
bcolloran: actually, i want to have the two separate tasks for use in the menus, but I can add another "toggle skip" task to handle the keybinding.

bcolloran: rechanged my mind, one toggle task is best :-)

bcolloran:
again re the jsmd -- actually, the mapping is out of scope. i can worry about making the jsmd pretty later. for now, i'll just add this to included cell properties

src/actions/task-definitions.js
hamilton: Yep, totally fine skipping the keyboard shortcut for now, unless it ends up being very obviously needed.

bcolloran: oh i actually went ahead and added it :-) we can remove if we don't like it

bcolloran:
thanks @hamilton  :-)

PR #553, opened 2018-04-12
Fix #495: Dynamically load codemirror syntax highlighters

### LINKED ISSUE #495: Add support for dynamically loading CodeMirror syntax highlighters
When a new language is loaded via the `addLanguage` API (as being developed in #489), we'll need to dynamically load the syntax highlighting definition for that language over the network.  There is a [codemirror addon](https://codemirror.net/demo/loadmode.html) for this that hopefully will do the trick.

mdboom:


bcolloran:
CodeMirror is definitely a weird little snowflake in our code. I admit I don't full understand how all the plugins work -- seems like they all mutate the CodeMirror object  in `cell-editor.js` directly when they are imported, which is guess is why we have all those `// eslint-disable-line no-unused-vars` in there, and why we have to load the other dependencies in this file rather than in a helper file? Hmm, I wonder if there is a way we can pull this out-- maybe manage all of our codemirror stuff in one file (codemirror-iodide.js) that exports the configured version of codemirror with all the plugins we need and that has your helper functions for loading new codemirror plugins as needed. Just trying to think whether there's a way of keeping the cell-editor file focused on that component.

i'll do the requisite QA tomorrow, at which point i'll be happy to merge this cuz i don't want to block the progress, and we can file a separate issue around making the codemirror stuff tidier.

I guess it's also worth pinging @hamilton to peek at this since he's been working on the plugin loader code. Hamilton, do you see any quick wins in terms of moving any of this functionality into your revised plugin loader stuff?

bcolloran:
CodeMirror is definitely a weird little snowflake in our code. I admit I don't full understand how all the plugins work -- seems like they all mutate the CodeMirror object  in `cell-editor.js` directly when they are imported, which is guess is why we have all those `// eslint-disable-line no-unused-vars` in there, and why we have to load the other dependencies in this file rather than in a helper file? Hmm, I wonder if there is a way we can pull this out-- maybe manage all of our codemirror stuff in one file (codemirror-iodide.js) that exports the configured version of codemirror with all the plugins we need and that has your helper functions for loading new codemirror plugins as needed. Just trying to think whether there's a way of keeping the cell-editor file focused on that component.

i'll do the requisite QA today/tomorrow, at which point i'll be happy to merge this cuz i don't want to block the progress, and we can file a separate issue around making the codemirror stuff tidier.

I guess it's also worth pinging @hamilton to peek at this since he's been working on the plugin loader code. Hamilton, do you see any quick wins in terms of moving any of this functionality into your revised plugin loader stuff?

Issue #546, opened 2018-04-10
plugin helper scripts should return a promise, plugin cell should update based on promise


bcolloran:
@mdboom  @hamilton 
any more thoughts on details of how this should work?

hamilton:
I will have a PR in the next day or so demonstrating how we might handle making the cell execution w/ plugin cells feel synchronous. This will require some tradeoffs I'll bring up in the PR.

hamilton:
Closing this due to the merging of #554

PR #545, opened 2018-04-10
when using ?url= notebooks that were exported in presentation mode now evaluateAllCells

imshubhamsingh:
I've created a new action `importFromURL` which returns a promise so that once state is loaded and if `state.viewMode = 'presentation'`, it will run `evaluateAllCells`

PR #554, opened 2018-04-12
evaluateAllCells can handle Promise-based actions, evaluateLanguagePluginCell is now a Promise-based action

hamilton:
This is work toward making #546 possible. Given that the compilation of the wasm file is done asynchronously, we need a way to wait to evaluate further cells until after the compilation is complete and the resulting namespace is available.

The main tweaks here are:
- `evaluateAllCells` now chains together Promises that evaluate the next cell. So if any of those dispatches returns promises, they must resolve before we move on. I moved that action into `actions.js` since `redux-thunk` provides an easy way of formulating the action.
- `evaluateLanguagePluginCell` now returns a Promise that awaits the resolution of the helper script XHR request & subsequent evaluation of it. Critical to this - the helper script found at `url` is now `eval`'d. If the output of that `eval` is a Promise, then we wait for the promise to resolve. If the output of the `eval` is NOT a Promise (the `pyodide.js` helper is not as of this filing) then it will do what it currently does - evals the helper and moves on.
- I will be filing a PR to tweak `pyodide.js` and turn it into a Promise based on the [GH Gist](https://gist.github.com/hamilton/8498d5eea350dce94d37d7bbcf5dc25e) that I'm using in the notebook linked to below. It functionally does the exact same thing, but now we can wait for the compilation to complete before resolving the promise and moving on.

This will set the stage for `iodide.requireResolution()` and `iodide.resolve()` as a way of preventing succeeding cells from being run until we resolve explicitly. I will tackle this in the next PR, but wanted to set the stage for this here first.

So! To test this PR, you can use `?url=https://gist.githubusercontent.com/hamilton/578c3ac2b87d99735b9b766145d01282/raw/python-test.html` to load a shorter version of @mdboom's notebook, and follow along.

I am not sure this is the right approach, but it seems like the easiest way to minimally achieve what we need.


hamilton:
@bcolloran I brought in changes from the #545 merge into master. I'll leave it to you to merge this at your discretion, unless there are any other changes you'd like to see here.

hamilton:
@mdboom thanks for the review. I'm not sure if this will be the right way going forward. Frankly I'd really like to also have another language via WebAssembly that's a bit lighter. I think Lua is a good candidate since it's designed to be lightweight / embeddable, which I've brought up to @bcolloran before. This approach is probably the most flexible for now, and until we get another language here, I think we should be fine with the API for adding languages to change a bit down the line.

mdboom:
We could also support transpiled languages through this interface (Typescript, Coffeescript, Clojure etc.) which would be even lighter...

PR #556, opened 2018-04-12
update to react 16 + enzyme-react-adapter-16

### LINKED ISSUE #404: Upgrade to React 16
[React 16](https://auth0.com/blog/whats-new-in-react16/) has a number of big updates, particularly Fiber, which could have big (positive) performance implications. It also has support for fragments, which would be immediately useful for a PR I am working on now, and would be generally useful for the future. The API changes seem fairly minimal and I don't believe would break our app.

Not pressing, but worth bringing up.

hamilton:
Fixes #404. This PR is a test of moving to [React 16](https://reactjs.org/blog/2017/09/26/react-v16.0.html), which has many very nice features (especially fragments) and a more efficient rendering method.

There are no API-breaking changes in React 16 that affect us. All tests pass, and example notebooks appear to work.

This is both non-breaking and very likely to not cause any merge conflicts in the next few weeks, so I'd suggest a review simply goes through example notebooks and makes sure things actually do all run as we expect.


bcolloran:
works great, thank you for tackling this one!

PR #557, opened 2018-04-12
decomposed all the handlers into individual files

### LINKED ISSUE #276: break out handlers into separate files
The handlers appear to all be in `output.jsx` out the moment, which could grow unwieldy as we add more handlers. Doing this would make it easier to just let `CellOutput` be a vessel for the handlers defined in other files.

Unless I'm missing something. `ExternalResourceCell` has a special type of output that isn't really applicable in other cells, and I think that's probably a good thing, but just trying to figure out how to square this with the current approach.

I could:

a. make a special `externalOutputHandler` and square it with the current `CellOutput` implementation (specifically around checking for certain structure to the value) before it is selected, or
b. make a whole other custom component type for the `externalOutputHandler` and leave the `CellOutput` as-is. This is the approach taken in `MarkdownCell, roughly`. If we go with this direction (which I'm not averse to) I would maybe suggest we rename `output.jsx` to reflect that it is primarily for javascript cells.

For #272 I'm going to pursue b given the example in `MarkdownCell`. All the same, as we add more output handlers into the core, it still might make sense to decompose that file.

hamilton:
This closes #276 by separating out all the different handlers in `value-renderer.jsx` into their own files.

A few questions:

- is 'handler' the term we want to go with here, or 'rep'?
- ~~should I add some basic tests to each one of these to determine if `shouldHandle` catches correctly?~~ (this one mostly done now)

### DIFF
--- a/src/components/reps/__tests__/array-handler.test.js
+++ b/src/components/reps/__tests__/array-handler.test.js
@@ -1,0 +1,18 @@
+import arrayHandler from '../array-handler'
+
+describe('dateHandler shouldHandle', () => {
+  it('rejects non-scalar values', () => {
+    expect(arrayHandler.shouldHandle(undefined)).toBe(false)
+    expect(arrayHandler.shouldHandle({})).toBe(false)
+    expect(arrayHandler.shouldHandle(new Promise(() => {}))).toBe(false)
+    expect(arrayHandler.shouldHandle('string')).toBe(false)
+    expect(arrayHandler.shouldHandle(4000)).toBe(false)
+  })
+  it('accepts array objects', () => {
+    expect(arrayHandler.shouldHandle([])).toBe(true)
+    expect(arrayHandler.shouldHandle(new Array(10000))).toBe(true)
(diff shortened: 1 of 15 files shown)


bcolloran:
looks good to me, veeery nice to have some tests. i have no strong feelings about "reps" v "handler" either.

Not to be addressed here, but seeing the repetition in the tests re whether a handler should be used or not, it makes me wonder whether about the pros and cons of having a single "repChooser" object that we can test in one place rather than the shouldHandle methods attached to the individual handlers. i the main drawback i can think of off the top of my head would be some additional complexity around injecting new shouldHandle functions into that object.

@mdboom i'm sure you have thoughts on this -- iPython does it similarly to the shouldHandle approach? 

(Just brainstorming at this point, I haven't really put much thought into this, but maybe worth looking into whenever we get around to consolidating handlers with the redux store)

mdboom:
The main reason for having `shouldHandle` is that it's extensible: you can add new output handlers to the list (as our example notebook does) without having to edit a singleton object.

PR #559, opened 2018-04-13
Freeze state

### LINKED ISSUE #551: proposals re: freezing environment variables
Objective: to enable users to "freeze" some user-defined variables from their global js execution scope so that expensive computations can skipped when sharing notebooks.

#### writing and export
- During exploration phase, a user runs a bunch of cells, including some slow computations.
- As the narrative that the user wants to tell begins to crystallize, the user realizes that only some of the outputs from these computations will be needed to create their final report.
- the user marks the cells with expensive computations as "skip in run all"
- The user creates an "environment cell" that includes the names of several variables from the global scope.
- Upon running this cell, the indicated variables are saved _in their current form_ to the iodide store
    - running it multiple time will allow updating of the frozen environment vars
- At export, the frozen env variables are added to the JSMD

#### reading and import
- At next import, the saved env vars are immediately dropped into the global namespace (before the "run all" action is triggered).
- since the author marked a bunch of cells as "skip", the non-skipped cells will begin running right away with access to the frozen values of the intermediate computations.
- the user immediately see the report the author intended

#### server-side execution
- for the server-side computation we've been thinking of, the default might be to ignore "skip in run all", or to have a flag that allows this.
- this would mean that the server-side runner would execute the NB cells in order, attach the specified  vars to the iodide store when it hits the env cell, keep executing the NB until the end (including any additional env cells) and at the end the server-side runner script would emit a new notebook with jsmd that includes the frozen values of these variable
- then when a user loads the NB in the browser in Report view, the env vars will be loaded immediately and the "skips" will be honored, dropping the reader right into the report as the author intends

#### wrinkles
- i need to look more for a good library that can "pickle" objects, but it may be the case that serializing functions (and hence js classes / class instances?) is not a straightforward thing to do, so we might need to forbid that. in that case, we'd want to throw obvious errors in the UI.

bcolloran:
fixes #551 (at least until we have experimented enough to know what use cases are missing)

to test this load:
`iodide.dev.html?gist=bcolloran/14a1eceeefc2cb1cbf65d38cea4112fd`
in your local /dev directory.

i've added several methods to `iodide.environment` that allow saving of JS objects to `state.savedEnvironment` in a few ways:
- `save` converts to json and compresses to a base64 encoded string
- `saveStringCompressed` compresses a string before adding to the env (maybe for big text chunks (csvs or strings that are already json))
-  `saveJson` converts to json but does not compress (in case you want human-readable json in the jsmd)
- `saveStringRaw` justs adds a raw string to the env (and hence the jsmd)

there are corresponding `load` methods for each type of save.

(ps: i've done these as separate methods rather than options to make them more findable via tab completion. i'm not wedded to this if it is for some reason powerfully odious, but i think findable-via-autocomplete is pretty nice.)


bcolloran:
thanks @hamilton, great comments/questions/feedback!

- re save and append: good point, agree that both should be available. I'll add
`environment.append.`
- i'll look at that bug re the string save/load
- good point re the getter. I'll add a method `get` that adds that behavior
- i'll fix loadStringRaw, saveRawString etc to have consistent names (hmm, or maybe better yet, just add some info to the export that makes thawing automatic)
- on the other name stuff -- ya i thought freeze/thaw was a little weird. maybe pack/unpack is better, or bundle/unbundle, or stash/retrieve, or ... ? hmm, it's a bit more verbose, but i guess like "bundle" ok. any preferences? @mdboom any thoughts?
- i'll add some tests before merging.

bcolloran:
i'm going to remove some of these options. i think it's sufficient for `bundle(obj)` to jsonify and compress and for `bundleRaw(obj)` to store raw strings. if for some reason someone wants to store a json string in their jsmd, they can use `bundleRaw({key:JSON.stringify(obj)})`

PR #567, opened 2018-04-17
Promise output handler / rep

### LINKED ISSUE #560: Promise output handler
The Promise output handler should display a the resolved value, and have states for resolved (the value pushed through the output handler handler), rejected (some special error / rejected output handler), and pending (a spinning thing that says "pending")

hamilton:
This PR creates a new handler for Promises (and thus closes #560), which will display the status of the Promise, as well as the eventual return value.

apr-17-2018 13-55-15

PR includes:
- new output handler + React Component (`PromiseRep`)
- tests for `shouldHandle` as well as verifying the `PromiseRep` does things as expected

To test, check out this PR and use `iodide.dev.html?url=https://gist.githubusercontent.com/hamilton/d49104af9f52a6c159444209bac83eec/raw/4952d79b5072a4a5e7f96de93e8c2982affb1247/promise-output-handlers.html`.


src/components/reps/promise-handler.jsx
```diff
@@ -0,0 +1,70 @@
...
+    )
+  }
+}
+
+export default {
+  // based off of https://stackoverflow.com/questions/27746304/how-do-i-tell-if-an-object-is-a-promise
+  // It is unfortunate that this seems to be the main perspective for how to determine if
+  // something is Promise-ful. Ideally we would check only for native Promises,
+  // but at this point I don't know if others use Promise-based libraries.
+  shouldHandle: val => typeof val === 'object' && typeof val.then === 'function',
```
bcolloran: as discussed-- let's only support built-ins
in that case, i think we should have:
`Promise.resolve().toString()` -> `"[object Promise]"`
(this is the case in Chrome, i haven't checked FF)

hamilton:
Chatted with @bcolloran - changes are as expected, so I'm going to merge.

Issue #565, opened 2018-04-17
Warning: Failed prop type: Invalid prop `data` supplied to `JSONTree`.


hamilton:
This warning has been in the console for a minute, but we should make sure to catch this and deal with it.

bcolloran:
i'm not

hamilton:
@bcolloran was there more to your comment?

PR #571, opened 2018-04-18
WIP: myriad output handler changes

hamilton:
This is not complete yet - will tag a reviewer when it is. The goal is to clean up the output handlers a bit:

- [x] fix the warning made by the recent changes to the declared variables pane (#565)
- [x] add typed arrays to the output handlers
- [x] finish the test suite associated with reps / handlers
- [x] DOM elements now output the dom element
- [x] ~DOM elements will render as a DOM inspector, akin to a browser DOM inspector~ __we decided that this is less useful right now that the raw DOM element, so we are going to try that out and see what breaks__

To test a DOM element, try this in a cell:

```javascript
var test = document.createElement('h1')
test.innerHTML = 'hiya!'
test
```

To test the new typed array interface, copy this into a cell and run:

```javascript
var typedArray = new Uint32Array([10000, 2, 3, 4])
typedArray
```

And you'll see it returns a normal `arrayHandler` rendered component, appropriately labeled for now.


hamilton:
It's not 100% clear that `domElementHandler` is a good idea, though it feels like it is for now. It will, however, end up with cases like this one:

screen shot 2018-04-17 at 10 43 43 pm

This could lead to some weird side-effects if I select a heavy or visually wide element.

hamilton:
Having a special `domElementHandler` is emphatically __not__ what we want to support. Here's a pretty concise example of why:

apr-17-2018 22-48-27

Yes, I'm selecting the entire react app and attempting to output it using the `domElementHandler` I just made.

I am way more in favor of having the DOM inspector elem that @mdboom found than this, obviously.

hamilton:
Here's `domElementHandler` using `DOMInspector`, as per @mdboom's earlier work.

apr-18-2018 08-56-15

src/components/panes/declared-variables-pane.jsx
```diff
@@ -8,6 +8,8 @@ import SidePane from './side-pane'
 import { DeclaredVariable } from './declared-variable'
 import tasks from '../../actions/task-definitions'
 
+const doNotShow = new Set(['CodeMirror'])
```
bcolloran: did you know that you've already created equivalent functionality elsewhere in the code? see `initialVariables` in `notebook-reducer.js`. ;-)

src/components/panes/declared-variables-pane.jsx
```diff
@@ -21,7 +23,9 @@ export class DeclaredVariablesPaneUnconnected extends React.Component {
     return (
       <SidePane task={tasks.toggleDeclaredVariablesPane} title="Declared Variables" openOnMode="declared variables">
         <div className="declared-variables-list">
-          {this.props.userDefinedVarNames.map(varName =>
+          {this.props.userDefinedVarNames
+            .filter(varName => !doNotShow.has(varName))
+          .map(varName =>
```
bcolloran: see other comment -- we filter out `initialVariables` in notebook-reducer, so i think maybe no changes are needed in this file?

src/components/reps/value-renderer.jsx
```diff
@@ -13,15 +12,17 @@ import arrayHandler from './array-handler'
...
 
 export function renderValue(value, inContainer = false) {
   for (const handler of handlers) {
     if (handler.shouldHandle(value, inContainer)) {
       const resultElem = handler.render(value, inContainer)
-
       if (typeof resultElem === 'string') {
         return (<div>{ resultElem }</div>)
       } else if (resultElem.tagName !== undefined) {
+        // what is this even for?
```
bcolloran: @mdboom @hamilton mike do you recall what this code block is for? if we don't know how it's used, let's cut it, no?

src/components/panes/declared-variables-pane.jsx
hamilton: I am half-aware, but figured it somehow got lost during one of the refactors since the declared variables pane doesn't functionally use any of it. I'll take a second look here and remove this if not.

src/components/panes/declared-variables-pane.jsx
hamilton: If this were true, `CodeMirror` would not be in the declared variables pane by default, right? Perhaps the available globals are getting checked before we instantiate CodeMirror though.

src/components/reps/value-renderer.jsx
hamilton: I think this was for DOM elements before, but I believe `resultElem` almost never returns a vanilla DOM element at this point, right?

hamilton: I'll remove it with that in mind.

hamilton:
Got it :)

>

src/components/reps/value-renderer.jsx
mdboom: This detects a DOM node and displays it verbatim.  Some of the examples in the output handling notebook use this to return custom representations of objects.  This is distinct from the cell returning a DOM element, in which case we should get a DOM inspector.  If an output handler returns a DOM element, it should be displayed verbatim (and I think this is what this does).

If we want to be careful about the recursive DOM element problem you outline above (where the user inserts a DOM element included elsewhere on the page), I think it should be possible to check if the DOM element already has a document assigned, and only do this if there isn't a document.

hamilton:
@mdboom yep I'm going to uncomment out the original part of displaying DOM elements and remove the inspector for now, since I don't think it adds much at the moment. It would be easy to perform that check that you suggested later, if it seems like this is a bad idea. So I am cool with providing just enough rope, as I've demonstrated before :) 

apr-19-2018 10-03-39

src/components/reps/value-renderer.jsx
hamilton: Got it, will keep it in. Thanks for the clarification!

src/components/panes/declared-variables-pane.jsx
hamilton: fixed - thx for feedback!

hamilton:
One thing worth thinking about - if the raw DOM element overflows in X, it might be smart to make the container scroll-x for now, since I do think that blowing up the page CSS in that instance is not useful, and at any rate I bet this will be common for users tweaking the the sizing of their DOM element.

hamilton:
At this point all the review points have been addressed. Let me know if there's anything else that should get tweaked here.

apr-19-2018 10-44-10

PR #575, opened 2018-04-21
Make side panes resizable

djbarnwal:
Side panes are now resizable. Looking for feedback on the resizing bar. I have made it black and thick so that it would blend with the current theme.

### DIFF
--- a/src/components/page.jsx
+++ b/src/components/page.jsx
@@ -82,7 +82,13 @@ class Page extends React.Component {
         className={this.props.viewMode === 'presentation' ? 'presentation-mode' : ''}
       >
         <NotebookHeader />
-        <div id="cells" className={this.props.viewMode + (this.props.sidePane ? ' pane-open' : '')}>
+        <div
+          id="cells"
+          className={this.props.viewMode}
+          style={{
+            width: this.props.sidePane ? `calc(100% - ${this.props.sidePaneWidth + 17}px)` : '100%',
+          }}
+        >
           {bodyContent}
         </div>
       </div>
@@ -97,6 +103,7 @@ function mapStateToProps(state) {
     viewMode: state.viewMode,
     title: state.title,
     sidePane: state.sidePaneMode,
+    sidePaneWidth: state.sidePaneWidth,
   }
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -365,7 +365,7 @@
     "arr-flatten": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/arr-flatten/-/arr-flatten-1.1.0.tgz",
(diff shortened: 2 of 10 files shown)


src/components/page.jsx
```diff
@@ -82,7 +82,13 @@ class Page extends React.Component {
         className={this.props.viewMode === 'presentation' ? 'presentation-mode' : ''}
       >
         <NotebookHeader />
-        <div id="cells" className={this.props.viewMode + (this.props.sidePane ? ' pane-open' : '')}>
+        <div
+          id="cells"
+          className={this.props.viewMode}
+          style={{
+            width: this.props.sidePane ? `calc(100% - ${this.props.sidePaneWidth + 17}px)` : '100%',
```
hamilton: this actually breaks quite a bit of css for presentation view. Can make sure this width only happens when we are in editor mode as well? Setting `div#cells` to `100%` for presentation view will counteract some internal reports, custom styling, etc. that others are relying on for their own presentation style sheets, which will set this `div` to be another width in some cases. (I'm noticing this because of an internal Mozilla notebook I test things against is broken with the change).

One approach would be to set the `width` to `undefined` in this operator if `this.props.sidePane === false` and `this.props.viewMode !== 'editor`.

hamilton:
Forgot to mention - I'm also noticing the history pane is a bit broken as a result of these changes - would you mind looking into that?

screen shot 2018-04-24 at 12 27 03 pm

djbarnwal:
@hamilton I am unable to recreate the history pane issue. Can you give me a little more insight into the problem?

hamilton:
@djbarnwal just checked on master - it looks like the history pane is buggy there as well, so I'll file a separate issue.

src/components/page.jsx
```diff
@@ -46,12 +46,21 @@ class Page extends React.Component {
...
+    this.getPageWidth = this.getPageWidth.bind(this)
   }
 
   shouldComponentUpdate(nextProps) {
     return !deepEqual(this.props, nextProps)
   }
 
+  getPageWidth() {
+    let width = '100%'
+    if (this.props.viewMode === 'presentation') width = 'unset'
```
hamilton: Oddly, the `unset` keyword actually keeps the width at `100%` in presentation mode. Could you replace this with `undefined` which will actually skip doing anything with the `width` property for this element?

djbarnwal: Strange! `undefined` was not working as expected while I was working on this issue whereas `unset` was. I'll just check if there is something loose on my end and send a fix.

djbarnwal:
@hamilton Turns out it was a mix-up on my end. My latest commit solves the issue :)

bcolloran:
nice job on this one @djbarnwal and @hamilton :+1:

PR #578, opened 2018-04-25
fixes Promise state change invisibility issue

### LINKED ISSUE #576: If a promise errors, it isn't possible to correct and re-run the cell
### What I Did

Create a promise-returning cell with a typo:

```
fetch('iodide.dev.css').then(response => response.txt())
```

Running the cell resolves the promise and displays the exception / traceback.

However, when I corrected the typo (`txt` -> `text`) and reran the cell, the same exception was displayed again.

hamilton:
Should fix #576 

The `PromiseRep` takes only one prop - the Promise. Because there is no real state that can be gleaned from the Promise (sadly), the rep won't update with a new (or identical) promise. So I propose we use `key` every time the handler `render` is called to just automatically update the component.

Promises that have already resolved will return the exact same thing.

### DIFF
--- a/src/components/reps/promise-handler.jsx
+++ b/src/components/reps/promise-handler.jsx
@@ -60,8 +60,10 @@ export default {
   shouldHandle: value => Object.prototype.toString.call(value) === '[object Promise]',
 
   render: (value) => {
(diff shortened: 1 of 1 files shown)


hamilton:
🎉🎉🎉

Issue #583, opened 2018-05-01
history pane view is broken


hamilton:
Something down the line broke the CSS for the history item.

<img width="678" alt="screen shot 2018-05-01 at 10 38 16 am" src="https://user-images.githubusercontent.com/95735/39484847-d61dfe50-4d2b-11e8-8796-0e6675c376bf.png">

PR #584, opened 2018-05-01
history item styling fixes

hamilton:
This PR addresses #583 -

- the history item width has been fixed and scales to the size of the parent container
- the outline now matches the unselected cell editor


hamilton:
I am realizing the overflow issues are still present, so I'm going to fix those first.

bcolloran:
lgtm, thanks @hamilton!

Issue #499, opened 2018-03-18
implement proper "prod" builds


bcolloran:
I haven't looked into this in detail yet, but it seems like we're at the poing where our env=prod build environment would benefit from all the standard JS build best practices-- minification, uglification, tree-shaking...

@wlach @hamilton do you have any thoughts on this? I've never really don't it before. I have the sense that this might be a good first project for someone that know JS build stuff well, but i'm not really sure how complex it is in general, so i don't really know

hamilton:
Couldn't agree more - let's put it on the agenda for the internal meeting this week. I don't think this is very hard to put together with webpack, but there may be some small quirks we'll have to figure out.

hamilton:
We've done a lot of this already, I think - using uglify has cut our prod bundle size down, and there is more savings to be had that we should pursue for both this and the server bundle.

bcolloran:
yup. we also need to set `NODE_ENV=production` for our prod builds to clear some kind of redux warning

bcolloran:
i think our current set up satisfies this in many way, and this is very stale. can file more specific issues if needed.

PR #585, opened 2018-05-01
simple uglify renders the prod bundle from 4.3mb -> 1.4mb

hamilton:
I added the `uglifyjs-webpack-plugin` to the production build to see the savings. That alone reduces the bundle size pretty considerably. (addressing #499)

This does not do any tree-shaking, as far as I can tell, though it might be. It's hard to tell how much work `uglifyjs-webpack-plugin` is doing. Webpack 4 makes tree-shaking easier as well.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,5 @@
 {
-  "presets" : ['es2016', "react"],
+  "presets" : ["es2016", "react"],
   "plugins" : ["transform-class-properties"],
   "env": {
     "test": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -335,6 +335,12 @@
         "default-require-extensions": "1.0.0"
       }
     },
+    "aproba": {
+      "version": "1.2.0",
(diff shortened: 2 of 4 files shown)


bcolloran:
looks good to me @hamilton.

small thing -- i was getting this in the console:
```
You are currently using minified code outside of NODE_ENV === 'production'. This means that you are running a slower development build of Redux. You can use loose-envify (https://github.com/zertosh/loose-envify) for browserify or DefinePlugin for webpack (http://stackoverflow.com/questions/30030031) to ensure you have the correct code for your production build.
```
so based on this --
https://stackoverflow.com/questions/39130860/minified-code-outside-of-node-env-production-this-means-slower-developmen
-- i made a couple small tweaks that cleared the warning.

please review/QA and verify it still works as you intended, then feel free to merge if you are happy (or of course you can ping back to me if you need to make additional revisions)

hamilton:
@bcolloran - looks good to me. Thanks for the QA! Going to merge.

PR #589, opened 2018-05-03
WIP - material-ui refinements and menu bug fixes

hamilton:
This seeks to fix most of the menu issues we've encountered with the enigmatic `material-ui@next` library.

- [x] menus now close when you click away
- [x] submenus now close all menus
- [x] add carat to cell menu so it is more obviously clickable
- [ ] you can easily move between submenus


hamilton:
@bcolloran I have fixed some % of the bugs associated with the menus. There is still one that is vexing me, but I'd rather not delay these positive changes too much more and will file an issue accordingly. For now, at least we can click away from menus, and the cell menu has a carat.

djbarnwal:
@hamilton Looks good to me :) Maybe we can address 'switching between submenus' issue in another PR.

hamilton:
Great, thanks for the review @djbarnwal :) I agree - let's worry about the unfixed bugs mentioned in my initial comment in another PR. Merging now.

Issue #196, opened 2017-12-08
[META] aim for high test coverage


hamilton:
As of this filing, we've mostly focused on testing the reducers. We still need to:

- [ ] complete the reducer tests.
- [ ] decompose the reducers further, and test some of the more atomic actions
- [ ] write the appropriate tests for components - cells, header, etc.

rvwatch:
Hello Iodide friends! I'm a first-time contributor and figured I'd knock out some reps for basic actions tests. I've created an actions.test.js file and added 31 basic tests to help get started. Would that be something you'd like me to create a PR for?

bcolloran:
Hi @rvwatch! Yeah, a PR for that sounds great, thanks! :-)

bcolloran:
stale; need to think more about tests, but will need more specific issues/spints

PR #591, opened 2018-05-08
Creates a actions.test.js file and 31 tests for basic action creators

rvwatch:
This PR touches on issue #196.  It helps with some test coverage for the ```actions.js``` file. It creates a ```actions.test.js``` file and ensures that all of the current basic action creators are tested. It does not test any of the actions that utilize middleware.


bcolloran:
thanks @rvwatch!

Issue #605, opened 2018-05-16
Firefox bookmark hotkey not functional on iodide page


jtg567:
MASTER

### What I Did

tried to bookmark the page by pressing CTRL+D in Firefox

### What I Expected

the bookmark dialog to come up and the star in the awesome bar to fill in

### What Happened 

nothing

### (OPTIONAL) notebook that reproduces bug 

https://strategy-and-insights.mozilla.com/shield-studies/index.html

hamilton:
You are the king of shortcuts my friend! We have taken `ctrl+d` at the moment for opening the "declared variables" pane, though clearly this is colliding in a way that isn't attractive. We are using Mousetrap to do this, but clearly we need to rethink how we're using the keybindings.

Were you in editor mode when you used `ctrl+d`, or just the top-level presentation view?

hamilton:
@bcolloran we could approach this in two ways:

- let's just get rid of the declared variables / history pane keybinding
- let's disable ALL iodide keybindings in presentation mode

Thoughts?

jtg567:
@hamilton I was in presentation mode

hamilton:
Great - easy fix. Thanks for the issue!

bcolloran:
thanks @jtg567 for the issue, thanks @hamilton for the fix!

PR #606, opened 2018-05-16
re-enable some default keybindings in presentation view

hamilton:
#605 revealed a small problem that we should just in general be cautious of - if someone goes to an iodide notebook that opens in presentation view, we should try to have as few keybindings as possible fire in that circumstance. This PR retains a number of keybindings one might expect if they were just looking at a webpage in presentation view.

Added 2 folk for review - not pressing, easy fix, take a look at your leisure.


bcolloran:
this looks reasonable and correct to me; i have not done manual QA, but trust you to do so. R+ as far as i'm concerned, merge when you are ready

Issue #599, opened 2018-05-10
show which variables are frozen in the declared variables pane


hamilton:

hamilton:
This was fixed w/ the PR mentioned above.

PR #610, opened 2018-05-17
added saved environment section to declared variables pane

hamilton:
screen shot 2018-05-17 at 4 33 01 pm

### DIFF
--- a/src/components/panes/declared-variable.jsx
+++ b/src/components/panes/declared-variable.jsx
@@ -5,7 +5,8 @@ import { ValueRenderer } from '../reps/value-renderer'
 
 export class DeclaredVariable extends React.Component {
   static propTypes = {
-    value: PropTypes.string,
+    value: PropTypes.any,
+    varName: PropTypes.string,
   }
   render() {
     const r = true
--- a/src/components/panes/declared-variable.jsx
+++ b/src/components/panes/declared-variable.jsx
@@ -5,7 +5,7 @@ import { ValueRenderer } from '../reps/value-renderer'
 
 export class DeclaredVariable extends React.Component {
   static propTypes = {
-    varName: PropTypes.string,
+    value: PropTypes.string,
   }
   render() {
     const r = true
@@ -13,7 +13,7 @@ export class DeclaredVariable extends React.Component {
       <div className="declared-variable">
         <div className="declared-variable-name">{this.props.varName} = </div>
         <div className="declared-variable-value">
-          <ValueRenderer render={r} valueToRender={window[this.props.varName]} />
+          <ValueRenderer render={r} valueToRender={this.props.value} />
         </div>
       </div>)
   }
--- a/src/components/panes/declared-variables-pane.jsx
+++ b/src/components/panes/declared-variables-pane.jsx
@@ -20,7 +20,7 @@ export class DeclaredVariablesPaneUnconnected extends React.Component {
   }
 
   render() {
-    const environmentalDVs = this.props.environmentVariables !== undefined ?
+    const environmentalDVs = Object.keys(this.props.environmentVariables).length ?
       Object.keys(this.props.environmentVariables).map(varName =>
         (<DeclaredVariable
           value={this.props.environmentVariables[varName][0]}
--- a/src/components/panes/declared-variables-pane.jsx
+++ b/src/components/panes/declared-variables-pane.jsx
@@ -13,7 +13,7 @@ import tasks from '../../actions/task-definitions'
 export class DeclaredVariablesPaneUnconnected extends React.Component {
   static propTypes = {
     userDefinedVarNames: PropTypes.arrayOf(PropTypes.string),
-    environmentVariables: PropTypes.arrayOf(PropTypes.string),
+    environmentVariables: PropTypes.object,
   }
 
   shouldComponentUpdate(nextProps) {
@@ -21,7 +21,6 @@ export class DeclaredVariablesPaneUnconnected extends React.Component {
   }
 
   render() {
-    console.log(this.props.environmentVariables)
     const edvElem = Object.keys(this.props.environmentVariables).length !== undefined ? (
       <div className="declared-variables-list">
         <h3>Saved Environment</h3>
--- a/src/components/panes/declared-variables-pane.jsx
+++ b/src/components/panes/declared-variables-pane.jsx
@@ -12,6 +12,7 @@ import tasks from '../../actions/task-definitions'
 export class DeclaredVariablesPaneUnconnected extends React.Component {
   static propTypes = {
     userDefinedVarNames: PropTypes.arrayOf(PropTypes.string),
+    environmentVariables: PropTypes.arrayOf(PropTypes.string),
   }
 
   shouldComponentUpdate(nextProps) {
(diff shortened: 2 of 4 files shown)


src/components/panes/declared-variable.jsx
```diff
@@ -5,15 +5,15 @@ import { ValueRenderer } from '../reps/value-renderer'
 
 export class DeclaredVariable extends React.Component {
   static propTypes = {
-    varName: PropTypes.string,
+    value: PropTypes.string,
```
bcolloran: why did you remove the varName proptype? we should check that as well s the value, no?

src/components/panes/declared-variables-pane.jsx
```diff
@@ -12,28 +12,50 @@ import tasks from '../../actions/task-definitions'
...
+          varName={varName}
+          key={varName}
+        />)) :
+      undefined
+    const edvElem = environmentalDVs !== undefined ? (
+      <div className="declared-variables-list">
+        <h3>Saved Environment</h3>
+        {environmentalDVs}
+      </div>
+    ) : undefined
```
bcolloran: maybe simplify this into a single if/else block rather than repeating the ` ? : ` pattern twice?

bcolloran:
hmmm, i'm not sure this is very useful in it's current state. that's not a criticism of this PR hamilton -- rather, i reckon you've exposed a limitation in the implementation of saving vars to a notebook. let's meet back in #599 to discuss what we want to get out of this. it might be that we can's show the values, then we shold just just need to list the env keys...

src/components/panes/declared-variables-pane.jsx
hamilton: sure will do

src/components/panes/declared-variable.jsx
hamilton: I removed the `varName` because this makes the `DeclaredVariable` more flexible to deal with other values, not just stuff that's in `window`, but happy to change this back / make a different component.

hamilton:
Sure, curious what you see as the limitation here. I can see this also just being a list of frozen variables as well.

bcolloran:
ah yes, the limitation i'm thinking of is that i set it up so that the variables are saved to the JSMD environment as compressed strings, so we'd need to decompress them to show the values (which i assume is what we really want / what would be most useful to the user?). i guess we could just go ahead and do that. perhaps one could imagine a user / notebook author not wanting that to happen? although, one assumes that the values will almost always be decompressed (why save them to the NB env if you aren't going to use them?...)

hrm, maybe we _should_ just decompress and show the values of the env variables...

src/components/panes/declared-variable.jsx
bcolloran: i think what you did re `window` makes sense, we just want the proptypes check for both props --
`varName` and `values` are both passed in as props, no? (varName is still used on line 14: `<div className="declared-variable-name">{this.props.varName} = </div>`)

hamilton: yep good call :)

hamilton:
Ah, yes. Had thought about this.

Hard to know for sure until we try using `iodide.environment` on more stuff, but when in doubt, I think we should aim for more flexibility, not less, and if the additional flexibility seems obviously a pain upon some usage, then we can change accordingly. I can think of lots of reasons why a user might not use a saved environment variable but still want it saved. Maybe they save a bunch of really big heavy models so they have them available, and only wants to use one of them for something, or later forks their NB to work on a tangent with a different one. We are in uncharted territories, and I don't think we gain too much from forcing users down a path with this feature.

That all said, if the decompression isn't THAT slow, and the compression doesn't save much in terms of size, might as well unfreeze all the saved variables. But if the decompression actually radically reduces the size of the thing (which might be the case in a big data set, if I have an array of objects with lots and lots of copied keys) and those space constraints really matter to the amount of memory being used, then I think users would probably prefer we don't unfreeze their variables for them. Not totally sure of the mechanics here.

bcolloran:
i'm really open to any alternative. If we think it buys more flexibility for the user, we need  not decompress automatically.

In that case, perhaps the thing to do is show the user the compressed size of the variable (at least in terms of char length, which is easy to do). (FWIW, showing the JSMD "type" probably doesn't matter from a user perspective -- `iodide.environment.get` uses that to automatically apply the correct decompression method, so a user shouldn't need to know the type)

Decompressing should be quite fast in general, but of course that will always depend on the size of the data. Perhaps there is a heuristic in the middle -- automatically decompress anything less than X chars in length, and require user action for anything bigger than that? We can chat about it thursday.

hamilton:
@bcolloran the environmental var PR is ready for another review. Let me know what you think of this direction / any changes you'd like to see with this.

src/components/panes/declared-variables-pane.jsx
```diff
@@ -6,34 +6,57 @@ import deepEqual from 'deep-equal'
...
     userDefinedVarNames: PropTypes.arrayOf(PropTypes.string),
+    environmentVariables: PropTypes.arrayOf(PropTypes.string),
   }
 
   shouldComponentUpdate(nextProps) {
     return !deepEqual(this.props, nextProps)
   }
 
   render() {
+    console.log(this.props.environmentVariables)
```
bcolloran: this log can  probably go :-)

bcolloran:
a few comment below, but overall i think this is a very nice improvement in usability! thanks for driving it forward! :-)

--

when i load `file:///home/bcolloran/iodide/iodide/dev/iodide.dev.html` (has no frozen env vars) i get

```
Warning: Failed prop type: Invalid prop `environmentVariables` of type `object` supplied to `DeclaredVariablesPaneUnconnected`, expected an array.
    in DeclaredVariablesPaneUnconnected (created by Connect(DeclaredVariablesPaneUnconnected))
    in Connect(DeclaredVariablesPaneUnconnected) (created by ViewControls)
    in div (created by ViewControls)
    in ViewControls (created by EditorModeToolbarUnconnected)
    in div (created by Toolbar)
    in Toolbar (created by WithStyles(Toolbar))
    in WithStyles(Toolbar) (created by EditorModeToolbarUnconnected)
    in div (created by EditorModeToolbarUnconnected)
    in EditorModeToolbarUnconnected (created by Connect(EditorModeToolbarUnconnected))
    in Connect(EditorModeToolbarUnconnected) (created by NotebookHeader)
    in MuiThemeProvider (created by NotebookHeader)
    in div (created by NotebookHeader)
    in NotebookHeader (created by Page)
    in div (created by Page)
    in Page (created by Connect(Page))
    in Connect(Page)
    in Provider
```


--

when i load `iodide/dev/iodide.dev.html?gist=bcolloran/215d760702429b6ddcf9a21e43485841` (which does have a couple frozen env vars) i get 
```
warning.js:33 Warning: Failed prop type: Invalid prop `environmentVariables` of type `object` supplied to `DeclaredVariablesPaneUnconnected`, expected an array.
    in DeclaredVariablesPaneUnconnected (created by Connect(DeclaredVariablesPaneUnconnected))
    in Connect(DeclaredVariablesPaneUnconnected) (created by ViewControls)
    in div (created by ViewControls)
    in ViewControls (created by EditorModeToolbarUnconnected)
    in div (created by Toolbar)
    in Toolbar (created by WithStyles(Toolbar))
    in WithStyles(Toolbar) (created by EditorModeToolbarUnconnected)
    in div (created by EditorModeToolbarUnconnected)
    in EditorModeToolbarUnconnected (created by Connect(EditorModeToolbarUnconnected))
    in Connect(EditorModeToolbarUnconnected) (created by NotebookHeader)
    in MuiThemeProvider (created by NotebookHeader)
    in div (created by NotebookHeader)
    in NotebookHeader (created by Page)
    in div (created by Page)
    in Page (created by Connect(Page))
    in Connect(Page)
    in Provider
```

and a bit below that

```
warning.js:33 Warning: Failed prop type: FrozenVariable: prop type `byteLength` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.
    in FrozenVariable (created by DeclaredVariablesPaneUnconnected)
    in DeclaredVariablesPaneUnconnected (created by Connect(DeclaredVariablesPaneUnconnected))
    in Connect(DeclaredVariablesPaneUnconnected) (created by ViewControls)
    in div (created by ViewControls)
    in ViewControls (created by EditorModeToolbarUnconnected)
    in div (created by Toolbar)
    in Toolbar (created by WithStyles(Toolbar))
    in WithStyles(Toolbar) (created by EditorModeToolbarUnconnected)
    in div (created by EditorModeToolbarUnconnected)
    in EditorModeToolbarUnconnected (created by Connect(EditorModeToolbarUnconnected))
    in Connect(EditorModeToolbarUnconnected) (created by NotebookHeader)
    in MuiThemeProvider (created by NotebookHeader)
    in div (created by NotebookHeader)
    in NotebookHeader (created by Page)
    in div (created by Page)
    in Page (created by Connect(Page))
    in Connect(Page)
    in Provider
```


--

Design is definitely on the right track, but perhaps the table could be a little tighter? there is a lot of space between the var names and the data size:

image

perhaps readability could be improved if the varname table column autosized to the content? (with some padding of course).

hamilton:
@bcolloran I don't personally mind the whole width spanning element actually - it reads pretty nicely and legibly to me, and adjusts nicely if folks change the pane. All the same I went ahead and reset the css for that so that it will scale a bit more intelligently.

Thanks for reviewing! Let me know if there's anything else with this PR that you'd like to see changed.

PR #611, opened 2018-05-18
typo

Fil:


hamilton:
Thanks for the PR @Fil :)

Fil:
Cool. I was not able to run the python example, maybe it's not ready for consumption yet?

hamilton:
Do you have ad blockers on both browsers?

Fil:
Yes!
Now (on FF) it runs a bit further, and then I get

AttributeError: module 'pandas' has no attribute 'read_csv'


-- Fil

mdboom:
Thanks for trying this out, @Fil!  Why don't we move this discussion over to its own issue in iodide-project/pyodide?  If there's any error messages in your Javascript console, those might be helpful...

mdboom:
Also, try turning off any ad blockers / tracking blockers if you have them -- that's been causing issues for other folks.

PR #614, opened 2018-05-22
Test for Cell components

djbarnwal:
Unit test for cell components.

@hamilton @bcolloran Do you think the unit test for `cell-menu` needs to be more rigorous? Though code coverage for that component is 100%, I am not checking for the individual props for all the cell options.


hamilton:
This looks great so far @djbarnwal! I think that the `cell-menu` tests that you've added are good for now, and this set of tests is really nice and helps with coverage quite a bit.

Let me know when you're ready for a full review (or if you want to add more tests for `cell-menu` that you have in mind, go for it). I can review this today or tomorrow when you give the thumbs up.

hamilton:
Thanks for the PR @djbarnwal!

Issue #597, opened 2018-05-10
if Promise is returned element, block further cell evaluation on further Promise resolution / rejection / error


hamilton:

hamilton:
After further thinking, I don't think this will necessarily solve the Pyodide issue of loading packages, and in fact will create new problems elsewhere.

- (@mdboom this is relevant for pyodide) for other language integration that requires something async (via WebAssembly) this will not solve our fundamental issues of blocking further cell execution upon resolution. It might make  something like this work, if `loadPackage` is a Promise:

```python
pyodide.loadPackage('pandas')
```

But this will not:

```python
import pyodide
pyodide.loadPackage("pandas")
import pandas as pd

df = pd.read_csv(pyodide.open_url("music-reviews.csv"))
df.describe(include='all')
```

- there are compelling reasons _not_ to force a `fetch` or some other Promise to block further cell evaluation upon resolution, and by forcing this on users we are limiting the utility of Promises / iodide generally.
- if users need a way of pulling in their data sets / external data via a `fetch`-like interface (but don't really want to learn how to use the `iodide.evalQueue` api or Promises generally) then perhaps we should actually provide a simple stdlib async data pull that does this as a convenience.

So I think to solve the pyodide issue we should think about whether there is something we can call or use to make this work. If pyodide have access to the iodide API, then perhaps we can wrap `loadPackage` to also use `iodide.evalQueue.requireExlicitContinuation()` and `iodide.evalQueue.continue()`.

mdboom:
To get your second example to work, one would have to do something like:

```python
def do_rest():
    import pandas as pd

    df = pd.read_csv(pyodide.open_url("music-reviews.csv"))
    df.describe(include='all')

import pyodide
pyodide.loadPackage("pandas").then(do_rest)
```

Which obviously is worse than the Javascript equivalent since Python doesn't have good anonymous functions etc.  So I sort of saw the "block on returned promises" as syntactic sugar around this problem.  Note also, that `loadPackage` is a Javascript function, not a Python one, since I haven't yet figured out a way to integrate async Javascript with Python.

But I see your point that we are forcing one way of thinking about promises on the user that may not be the right one in all cases.  I think I agree with @bcolloran that it's reasonable that block-on-promise is the default.  However, it also breaks the concept that, at least logically, if you took all of the code cells and put then in a source file and ran it, you would have the same result -- i.e. there's implicit "wait and then" in there.

[...]

PR #618, opened 2018-05-23
added evalQueue.await to the iodide api

hamilton:
This PR adds another feature to the iodide api - `iodide.evalQueue.await`. Motivated by the discussion in #597, we want an easy way to prevent further cell evaluation until certain Promises are resolved, specifically any using the `fetch` api. So `iodide.evalQueue.await` works exactly as `Promise.all` - you pass in an array of promises (like `fetch` calls) and it waits for all of the Promises to resolve / reject before moving on.

I would imagine that the majority of data scientists / our users would prefer to just use this function instead of `iodide.evalQueue.requireExplicitContinuation` / `iodide.evalQueue.continue`, but those pieces are functionality will be vital in non-Promise situations, and at any rate I use both of them within `iodide.evalQueue.await`.

Use `?gist=hamilton/50e813eb8169f0486b48d55d5d206df8` for an example.  Below is a code block that demonstrates the functionality:

```javascript
let url1 = 'https://normandy.cdn.mozilla.net/api/v2/recipe/?format=json&ordering=-last_updated'
let url2 = 'https://normandy.cdn.mozilla.net/api/v2/recipe/?format=json&ordering=-last_updated&page=2'
var data1 =[]
var data2 = []

let timer = new Promise((go)=>{
  setTimeout(()=>{
    go('setTimeout Resolved.')
  }, 4000)
})

iodide.evalQueue.await([
  timer,
  fetch(url1).then(d=>d.json()).then(d=>{data1 = d; return d}),
  fetch(url2).then(d=>d.json()).then(d=>{data2 = d; return d})
])

```

### DIFF
--- a/src/iodide-api/evalQueue.js
+++ b/src/iodide-api/evalQueue.js
@@ -34,7 +34,18 @@ export const waitForExplicitContinuationStatusResolution = () => new Promise((re
   }
 })
 
+const awaitPromises = promises =>
+  // is functionally identical to Promise.all.
+  Promise.resolve()
+    .then(() => setRunningCellEvalStatus('ASYNC_PENDING'))
+    .then(() => Promise.all(promises))
+    .then((resolutions) => {
+      setRunningCellEvalStatus('SUCCESS')
+      return resolutions
+    })
+
 export const evalQueue = {
   requireExplicitContinuation: () => { setRunningCellEvalStatus('ASYNC_PENDING') },
   continue: () => { setRunningCellEvalStatus('SUCCESS') },
+  await: awaitPromises,
 }
--- a/src/iodide-api/evalQueue.js
+++ b/src/iodide-api/evalQueue.js
@@ -59,14 +59,14 @@ export const waitForExplicitContinuationStatusResolution = () => new Promise((re
   }
 })
 
-const awaitPromises = promises =>
-  // is functionally identical to Promise.all.
-  Promise.resolve()
-    .then(() => {
-      setRunningCellEvalStatus('ASYNC_PENDING')
-      incrementAsyncProcessCount()
-    })
-    .then(() => Promise.all(promises).catch((err) => { throw Error(err) }))
+const awaitPromises = (promises) => {
+  setRunningCellEvalStatus('ASYNC_PENDING')
+  incrementAsyncProcessCount()
+  return Promise.resolve().then(() => Promise.all(promises).catch((err) => {
+    resetAsyncProcessCount()
+    setRunningCellEvalStatus('ERROR')
+    throw Error(err)
+  }))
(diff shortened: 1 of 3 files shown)


mdboom:
Cool.  This looks like it will help the existing `pyodide.loadPackage` functionality work with "run all".  

What happens if this is called multiple times in the same cell?  It might be called from within a called function, for example, and not actually be present in the cell content, therefore it might be easy for the user to accidentally call it twice.

bcolloran:
looks very nice. ping when you're ready for review

hamilton:
@mdboom + @bcolloran I've been playing with some ideas around how we want to handle this. I've migrated over to a version in which the api calls increment + decrement a new cell field `asyncProcessCount`. The polling then watches that counter, and once it hits zero, resolves. This allows multiple / nested calls for `iodide.evalQueue.await`, `iodide.evalQueue.requireExplicitContinuation` (which increments) and `iodide.evalQueue.continue` (which decrements).

This feels like the right way to move forward with this, api-wise. Let me know what y'all think.

bcolloran:
needs some error handling. for example, this prevents execution of further js cells:

```
let url1 = 'https://normandy.cdn.mozilla.net/api/badrequest'
let url2 = 'https://normandy.cdn.mozilla.net/api/v2/recipe/?format=json&ordering=-last_updated&page=2'
var data1 =[]
var data2 = []

let timer = new Promise((go)=>{
  setTimeout(()=>{
    go('setTimeout Resolved.')
  }, 4000)
})

iodide.evalQueue.await([
  timer,
  fetch(url1).then(d=>d.json()).then(d=>{data1 = d; return d}),
  fetch(url2).then(d=>d.json()).then(d=>{data2 = d; return d})
])
```

api looks great.

code looks ok to the extent i can follow it, but i don't know promise stuff that well, so requesting code review from @wlach

hamilton:
@bcolloran thanks for that catch. I've added more error handling around `evalQueue.await`. Any rejection or error will now be caught in a `.catch` in the api, so we can surface the error message and free up further cell evaluation (we can revisit THAT part down the line)

src/iodide-api/evalQueue.js
```diff
@@ -34,7 +59,29 @@ export const waitForExplicitContinuationStatusResolution = () => new Promise((re
   }
 })
 
+const awaitPromises = promises =>
+  // is functionally identical to Promise.all.
+  Promise.resolve()
```
wlach: Is there a reason why you're creating a problem that is immediately resolved here? Could we not just key this off of Promises.all() and do whatever state setting we want to do immediately?

hamilton: yep this is madness, fixing

PR #624, opened 2018-05-26
601 fix collapse button

bcolloran:
@hamilton small pr, mostly css changes to fix the collapse button, and i also changed out the cell menu carrot icon for something a bit smaller. should be a two minute review.


hamilton:
This looks good to me :) will probably useful in the meantime to have this since we are not implementing #521. Feel free to merge when you're in front of your keyboard.

Issue #433, opened 2018-03-05
add ability to "highlight" multiple cells (for "run selected", cell cut and paste)


bcolloran:
in order to facilitate "cell cut and paste" #65  and "run selected" #432 , we'd need to add the ability to have multiple cells selected to the state schema

pradeepgangwar:
@bcolloran I would like to give it a try.

bcolloran:
Cool, thanks Pradeep. :-) You might want to off on this one for a couple days, i have some changes coming around how state is managed and validated. Maybe we can check in around Wednesday? Thanks!

Issue #65, opened 2017-09-18
cell cut and paste


bcolloran:

djbarnwal:
Hey @bcolloran I would like to give this issue a try.

bcolloran:
Go for it @djbarnwal. Are you familiar with how this works in Jupyter?  We want to achieve similar functionality, but maybe we can do a little better.

all of the following will only work in command mode. I think the objectives for now will be something like:
- `ctrl+x` should "cut" the currently selected cell
- `ctrl+c` should "copy" the currently selected cell
- `ctrl+v` should "paste" the cell in the clipboard _below_ the currently selected cell

Juypter also has the possibility of highlighting multiple cells and cut/copy/pasting them. You can work on that too if you want, or just try to achieve the above objectives, but even if you want to stick with the smaller case, try to see if you can write code that will be flexible enough to accommodate the possibility of handling multiple cells.

Please let me know if you have any questions or want advice about how you might implement this. I have some ideas that you I could offer if you get stuck.

djbarnwal:
@bcolloran Yes, I am familiar with Jupyter's working. I will try to incorporate a similar behaviour in iodide with improvements wherever possible. I'll ping you if I stuck somewhere in between :-)

djbarnwal:
@bcolloran I have successfully implemented a basic copy and paste feature on a single cell which can be extended to cut-paste too. I am about to implement this to multiple cells. But before I start, I need to make sure that having multiple `selected` cells does not break anything. I know that `cellUp` and `cellDown` feature depends upon the currently selected cell (and it looks more logical to support multiple cell shift as Jupyter does). But a more thorough check would be required. 

Would it be OK if I send a PR for selecting multiple cells (with the existing features working) first and then follow it up by sending a PR for Cut-Copy-Paste? Moreover, do you want me to add some additional features to any part of this apart from what Jupyter offers?

bcolloran:
@djbarnwal that's great!!

i think that for multiple cell selection, we should actually introduce a new property in the state for each cell, perhaps we could call it "highlighted" or something like that. I'm quite sure that adding the possibility of having more than one cell in the `selected` state will introduce problems -- the entire app is built around the idea that only one cell is in `selected` state at a time.

See `newCell` in `state-prototypes.js` -- that is where we would add the new "highlighted" property. Lmk if you want any more advice, and thanks! :-)

PR #629, opened 2018-05-31
Multiple Highlight, Cut Copy and Paste

djbarnwal:
This PR adds two new features - Multiple selection and Copy-Paste. These features have been discussed at Issue #65 and Issue #433 

To select multiple cells hold down `Ctrl` or `⌘` key and click on the cells. Copy-Paste works with the standard key bindings.

### DIFF
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -565,6 +565,12 @@ export function unHighlightCells() {
   }
 }
 
+export function multipleCellHighlight(cellID) {
+  return {
+    type: 'MULTIPLE_CELL_HIGHLIGHT',
+    id: cellID,
+  }
+}
 
 export function cellCopy() {
   return {
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -551,10 +551,11 @@ export function selectCell(cellID, scrollToCell = false) {
   }
 }
 
-export function highlightCell(cellID) {
+export function highlightCell(cellID, revert = true) {
   return {
     type: 'HIGHLIGHT_CELL',
     id: cellID,
+    revert,
   }
 }
 
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -558,9 +558,22 @@ export function highlightCell(cellID) {
   }
 }
 
+export function unHighlightCells() {
+  return {
+    type: 'UNHIGHLIGHT_CELLS',
+  }
+}
+
+
 export function cellCopy() {
   return {
     type: 'CELL_COPY',
+  }
+}
+
+export function cellCut() {
+  return {
+    type: 'CELL_CUT',
   }
 }
 
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -558,6 +558,18 @@ export function highlightCell(cellID) {
   }
 }
 
+export function cellCopy() {
+  return {
+    type: 'CELL_COPY',
+  }
+}
+
+export function cellPaste() {
+  return {
+    type: 'CELL_PASTE',
+  }
+}
+
 export function deleteCell() {
   return {
     type: 'DELETE_CELL',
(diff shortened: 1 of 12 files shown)


bcolloran:
This is really beautiful work @djbarnwal, very nicely done! regarding @hamilton's comments, and a few other thoughts--
* hamilton, are you thinking that maybe we __should__ add buttons for cut/copy/paste? fwiw: jupyter does have button for these actions (but also fwiw: jupyter can look a bit cluttery). imo, visible buttons may not be needed, but i'm open to it
* to indicate highlighting, jupyter applies a blue background to the selected cell. i think it's a very clear visual cue, and in the absence of any firm reason to deviate, i think it seems quite reasonable to follow jupyter's lead on this one. However, i would suggest de-saturating the shade of blue a bit -- i think the alpha=0.3 is a bit dark; after experimenting a bit, i think alpha=0.1 is better, and is closer to jupyter (a shade lighter than their highlighting).
* +1 undo, that will be nice (some day we'll get there!... :-) )

other thoughts:
* i think we should follow jupyter's keybindings for this: in command mode, shift+up/down should do cell selection (like jupyter), and we should change the binding for moving cellsl up/down to be ctrl/cmd+up/down  (jupyter doesn't have a kbd shortcut for this action)

the code and overall approach are very straightforward and you've kept it clean and simple, so I'm happy merge once we change
(A) those keybindings and
(B) that alpha value.

There are a few other things that we can either address now or write up as follow-up issues (i'm happy to accept this PR now because it adds a lot even if we want to these follow up issues outstanding for a little while)

(1) in jupyter, in command mode if you use shift+mouse click, all of the cells from the currently selected cell to the one you click on will be highlighted. if we add this alongside with the ctrl+click behavior you have already implemented we will (i) match that feature from jupyter (ii) match the more general ctrl+click and shift+click behavior familiar from many situations in which you are presented with a list of items (file system browsers, spreadsheet cells, etc etc etc)

(2) it would be really neat to be able to use the cell menu on a whole bunch of highlighted cells, in particular, to be able to toggle "skip during run all" for all of the currently highlighted cells.

but yeah, i'm very happy with this PR, and I don't consider those ideas blockers at all, it's already very useful. so it's up to you -- if you just want to knock out (A) and (B) and merge we can do it (i lean toward merging sooner), but if you think that either or both of (1) and (2) are easy and you want to try to roll them into this PR while this section of the code is fresh in your mind, that's works too.

nice job! :-)

djbarnwal:
I'll like to give (1) and (2) a try. I am hoping that I would be able to sum it up by tonight. In case I don't I'll push (A) and (B) and save my work on (1) and (2) for later.

djbarnwal:
@bcolloran This PR is ready for another round of review. I have set the value of alpha to 0.15 as the value 0.1 wasn't contrasting enough on my screen and the highlight was barely visible.

bcolloran:
This is awesome @djbarnwal! I noticed a couple more things while doing some QA:

- if you are in command mode and you do e.g. `shift+down` a few times to highlight a few cells, the only way to un-highlight the cells seems to be to do a mouse click. We want the shift highlighting to work analogously to how it does in most text editors -- if you let go of shift and then do a key press without shift held, the highlight goes away, so let's make sure to do that.

- it's important that a cell always be selected, so in the case of doing a `ctrl-x` to cut cells, be sure to select the cell directly underneath the set of cells that were cut (if that cell exists; if there is no cell underneath, select the cell above; if _all_ cells are cut, a new blank cell should be added). (i'd recommend adding test around this to the cut/paste code; see also #670, though that is out of scope for this)

- in jupyter, the multiple highlight only appears when you click in the cell container, not when you click in the editor. Let's follow this behavior. this means that the changes in `cell-editor.jsx` should not be needed

- also following jupyter: when a cell is pasted, that cell is selected; when multiple cells are pasted, the bottom most of the pasted cells is selected. I think this is sensible behavior, lets follow it.

great work @djbarnwal -- this is a major and complicated feature, and you're doing an impressive job!

PR #617, opened 2018-05-22
creates a snackbar that appears when app messages are recorded

hamilton:
This PR creates a new app message notification system using `material-ui` snackbars. This will be useful for all sorts of app-level events, like saving to local storage, exporting to gist, and the like.

It's not particularly useful right now, and it users the probably-too-simplistic format of just an array of strings (so no time in which the app message happened, level, understanding of componentizing different types of messages, etc.)

I think one review would be enough, if any of y'all could take a look.

PR #631, opened 2018-06-01
Add support for importing ipynb files

mdboom:
To try this out, try appending to the URL:

`?ipynb=https://raw.githubusercontent.com/mdboom/strata-mpl-tutorial/master/Introduction%20to%20matplotlib.ipynb`

Eventually, we'll want a UI for this (and to send the errors somewhere other than the console), but I think this is good enough for demo purposes.

~~Note that the linked notebook revealed a bug in pyodide, in that the matplotlib figure isn't updated automatically when parameters are changed after it's created (see https://github.com/iodide-project/pyodide/issues/46), but if you force a refresh by panning and zooming you should be fine.  I think that bug will be easy to fix in short time anyway.~~

### DIFF
--- a/src/tools/handle-url-query.js
+++ b/src/tools/handle-url-query.js
@@ -3,6 +3,13 @@ import queryString from 'query-string'
 import { stateFromJsmd } from './jsmd-tools'
 import { store } from '../store'
 import { importFromURL, evaluateAllCells } from '../actions/actions'
+import { importIpynb } from './ipynb-import'
+
+function loadJsmd(jsmd) {
+  store.dispatch(importFromURL(stateFromJsmd(jsmd))).then(() => {
+    if (store.getState().viewMode === 'presentation') { store.dispatch(evaluateAllCells(store.getState().cells, store)) }
+  })
+}
 
 async function loadJsmdFromNotebookUrl(url) {
   try {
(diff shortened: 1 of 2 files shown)


src/tools/handle-url-query.js
```diff
@@ -11,21 +18,31 @@ async function loadJsmdFromNotebookUrl(url) {
...
 }
 
+async function loadIpynbFromUrl(url) {
+  try {
+    const response = await fetch(url)
+    const ipynbJson = await response.json()
+    const jsmd = importIpynb(url, ipynbJson)
+    loadJsmd(jsmd)
+  } catch (err) {
+    console.error('failed to load ipynb url', err)
```
hamilton: Any chance we could use `updateAppMessage` from `actions.jsx` here once the snackbar PR #617 lands? I'm assuming that PR might land before this one. Either that or this can be tackled in another issue easily.

hamilton:
This looks pretty good to me! Works quite smoothly. I am not sure if we should worry about the app message comment I made above too much before that PR lands.

And once `iodide.evalQueue.await` lands, we can add that to this template as well for the `pyodide.loadPackage` calls.

src/tools/handle-url-query.js
mdboom: Sure, I don't see why that wouldn't work.

hamilton:
We're going to change the app message piece once #617 lands. I'm going to merge this.

PR #636, opened 2018-06-03
language plugin load now applies syntax highlighting to all language cells

### LINKED ISSUE #563: plugin cells should update all editors when codemirror mode loader finishes
currently:
- when you open a notebook with python cells, at initial open the py cells don't have syntax highlighting
- when you run the py plugin cell, syntax highlight does not take effect immediately
- you have to change the cell mode to something other than py then back to py to get the highlighting.

obvi the highlighting should show up immediately, so the editors require some kind of refresh.

hamilton:
This fixes #563. The easiest way to check if a language has been added is to see if we have `window[languageModule]`. Saved notebooks with languages defined in the jsmd (like python) don't actually indicate if the language module is loaded yet, so you have to look for it. There is probably a smarter way of saving this information in the state so we don't have to do that. Either way, my approach changes a computed prop based on whether the module is present in `window`, which forces an update to the editor by changing the mode from `''` to `<language>`

### DIFF
--- a/src/components/cells/cell-editor.jsx
+++ b/src/components/cells/cell-editor.jsx
@@ -197,7 +197,9 @@ class CellEditor extends React.Component {
 function mapStateToProps(state, ownProps) {
   const { cellId } = ownProps
   const cell = getCellById(state.cells, cellId)
-  const languageModule = state.languages[cell.language].module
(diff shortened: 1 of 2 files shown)


mdboom:
Cool.  I've been annoyed by syntax highlighting not always working.

Issue #627, opened 2018-05-30
"export data" helper


hamilton:
Some Jupyter users (and I've been in this camp) will write a notebook simply to transform data and export it so they can use the data in some other system. It might be nice to provide some simple API functionality to export your data as a csv or json efficiently, so a user doesn't necessarily need to figure out how to do that w/ lower level system calls. The calls could look like:

```javascript
iodide.file.exportAsCSV(data, 'out.csv')
iodide.file.exportAsJSON(data, 'out.json')
iodide.file.exportAsBinary(data, 'out.whatever')
// the functions above serve as convenience functions for this one:
iodide.file.export(data, Array.isArray(data) ? : 'csv' : 'json', 'out.data')
iodide.file.export(imgData, 'binary', 'out.imgData')
```

Or something like that.

PR #637, opened 2018-06-03
iodide.file API

hamilton:
This adds API calls around exporting variables / data to files (#627). For now, it supports exporting as CSV, JSON, and as a general blob. CSV export really is the trickiest, because the JSON / raw representation relies on the JS apis for those, and CSV we have to make up as we go. With that said, the tests I've written are primarily about the file representation of various types of objects.

I have some open questions:

- should `exportCSV` work fine with single strings, numbers, & other scalars? I have it set up so it throws if these are accepted, but I don't really know / don't have a strong feeling about this right now.
- should `exportCSV` not use the `.toString` representation of functions?
- I have broken out the downloading function from `notebook-reducer` into its own exported function. For some reason, however, when I put this new function in `notebook-utils`, one of the tests breaks, but only one. And I can't figure out why that'd be the case, since other tests utilize the function that breaks. It's baffling. SO I've kept this downloader function in `notebook-reducer.js` because it doesn't break that test.

### DIFF
--- a/src/iodide-api/__test__/file.test.js
+++ b/src/iodide-api/__test__/file.test.js
@@ -19,12 +19,15 @@ const d9 = [
 const d10 = [
   { a: 10, b: { n: 1 } }, { a: 1, b: { n: 1000 } },
 ]
-
 const d11 = () => { console.log('this should break!') }
-
-// embedded fcns in array of objects?
 const d12 = [
   { a: 10, b: () => 100 }, { a: 20, b: () => 200 },
+]
+
+const d13 = undefined
+const d14 = null
+const d15 = [
+  { a: undefined, b: null }, { a: undefined, b: null },
 ]
 
 describe('toCSVString transformations', () => {
@@ -42,6 +45,9 @@ describe('toCSVString transformations', () => {
     expect(() => toCSVString(d11)).toThrow()
     // should functions be expressed w/ toString like this, or as [function Function] or whatver?
     expect(toCSVString(d12)).toBe(`a,b\r\n10,${d12[0].b.toString()}\r\n20,${d12[1].b.toString()}`)
+    expect(() => { toCSVString(d13) }).toThrow()
+    expect(() => { toCSVString(d14) }).toThrow()
+    expect(toCSVString(d15)).toBe('a,b\r\n,\r\n,') // is this right? should undefined / null be these empty things?
   })
 })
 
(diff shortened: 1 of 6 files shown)


bcolloran:
thinking about this somemore -- i'm not sure that we need to include the file formatters (csv) directly in core Iodide -- that should maybe be the responsibility of libraries that users load into their session. Perhaps the file API should really just be a wrapper around the `downloadResource` function, and users are responible for serializing the data themselves.

So that would be like--
```
iodide.file.downloadText(csvtools(data),"data.csv")
iodide.file.downloadText(JSON.stringify(data),"data.json")
iodide.file.downloadBlob(dataFrame.toArrow(),"dataframe.arrow")
```
--or whatever, and we can move the extra data handling libs to a separate repository.

this would be similar to how dealing with csvs and json and etc works in python -- the parsing/serializing libs don't handle the files directly.

the main thing is that i don't think we want to add a lot of data handlers into iodide core. i'm all for us building data handling libs, but as in python, those should be libs that you load, not built-ins.

hamilton:
hmm, good point. I also think this will keep us from having to write & test a really robust csv formatter, vs., say, using [`d3.csvFormat`](https://github.com/d3/d3-dsv/blob/master/README.md#dsvFormat). The json / blob versions are pretty light overall.

Though I'd say if we are going to take that approach, we should pick a winner for our examples, tutorials, and docs. Enthusiastic new users are going to get bogged down quickly without a good cut-and-paste example. I'd recommend we just use `d3` for the parsing / formatting.

Happy to `wontfix` / close this PR and we can think a bit more about the API / approach.

bcolloran:
to clarify: i think this PR is a great idea with some small tweaks (should be small since it's a light weight PR to begin with?), and that it's very sensible to have a file download api in our stdlib. So i don't think we should wontfix/close (unless you feel it's easier to start from scratch), just that we should strip out the specific data formatters and narrow the scope a bit.

i fully support picking the d3-csv module as our go-to example lib for dealing with csvs... it's what i always use anyways :-)

hamilton:
Cool. I will need to spend some time investigating why `downloadResource` caused one reducer test to fail for some reason when it was in `notebook-utils`, but I'm really into this API generally. The only downside I can think of, though is we're missing the mime type this way. Probably won't matter in practice, but figured I'd bring it up.

bcolloran:
hmm, could add mime as an optional arg for those who wish to specify it?

PR #620, opened 2018-05-25
WIP: Iodide Server

djbarnwal:
An approach towards Iodide Server and Export to Gist.
This PR has been deployed at [http://iodide-sv.herokuapp.com/](http://iodide-sv.herokuapp.com/)

To do - 
- [x] Attach Snackbar components on updates, give Feedback
- [x] Improve UI of App Messages
- [x] Persist login when a new notebook is loaded
- [ ] Persist Login for different sessions using LocalStorage (?)

PR #655, opened 2018-06-09
Export to Gist

djbarnwal:
This PR pushes the changes as planned in PR #620. 

There are a couple of things to note - 
1. Trying to log in from a domain other than `iodide.io` would give an App message of 'Unauthorized domain' for now. We plan to later funnel those users into iodide.io with their notebook intact. I think it would be fairly easy to do it using `window.open` method and passing the JSMD to it.

2. We are going to need a host to deploy our server. Heroku currently offers free dynos but they are not suitable for production. Do we plan to deploy the server after All hands?

### DIFF
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -2,6 +2,7 @@ import MarkdownIt from 'markdown-it'
 import MarkdownItKatex from 'markdown-it-katex'
 import MarkdownItAnchor from 'markdown-it-anchor'
 
+import { exportJsmdBundle, titleToHtmlFilename } from '../tools/jsmd-tools'
 import { getCellById, isCommandMode } from '../tools/notebook-utils'
 import {
   addExternalDependency,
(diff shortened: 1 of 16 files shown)


src/actions/actions.js
```diff
@@ -418,6 +419,111 @@ export function markCellNotRendered() {
...
+}
+
+function loginFailure() {
+  return (dispatch) => {
+    dispatch(updateAppMessages({ message: 'Login Failed' }))
+  }
+}
+
+export function login() {
+  if (window.location.hostname !== 'iodide.io') {
```
djbarnwal: Comment this out for testing the feature locally.

bcolloran:
@djbarnwal this code all looks pretty clean (to the extent i am able to follow it :-) ) but, um, sorry to be dense... what steps do i need to take to test this locally?

it looks like i need a .env file, and i think i probably need to register a new oauth app here: https://github.com/settings/applications/new
but i'm not sure what to enter for "Authorization callback URL", and i'm not sure how it all maps onto the thing i think i need to set in the `.env`, namely:
```
PORT
GITHUB_CLIENT_ID
GITHUB_CLIENT_SECRET
SERVER_URI
```

 could muck about with it longer, but thought it would be quicker to just ask :-)

also, just fyi: index.js fails the linter.

djbarnwal:
@bcolloran I have fixed the linting errors. I haved also DMed you the environment variables so that you can test out things easily. You need to add those environment variables (using your terminal perhaps) before you start the server using `node index.js` from the root. From there you can see the notebook served at `http://127.0.0.1:5000/`

bcolloran:
thanks @djbarnwal :+1:

src/actions/actions.js
wlach: Can you make this conditional on IODIDE_BUILD_MODE equaling heroku instead? This would allow us to use/test this functionality in dev/staging/etc environments (e.g. your own iodide-sv heroku environment) without code modifications. Can see some examples of this in the store code:

https://github.com/iodide-project/iodide/blob/14184cbbcf645ec459432e5634d683fbc2e74a58/src/store.js#L15

src/components/menu/view-controls.jsx
```diff
@@ -15,8 +18,25 @@ import AppInfoPane from '../panes/app-info-pane'
...
+    const loginItem = this.props.isAuthenticated ?
+      (
+        <NotebookTaskButton task={tasks.logoutGithub}>
+          <img src={this.props.avatar} alt="" className="user-avatar" />
+        </NotebookTaskButton>
+      )
+      :
+      (
+        <NotebookTaskButton task={tasks.loginGithub}>
+          <AccountCircle />
```
wlach: I would just use the text/link "Login" or "Sign in" here instead of an account circle, which is pretty ambiguous.

wlach:
After discussing with @bcolloran and @hamilton, I'm going to adjust this PR myself and merge it, since we really want to start testing our new server environment (http://extremely-alpha.iodide.io/)

wlach:
Closing in favour of #672.

PR #657, opened 2018-06-13
Allow importing and exporting a notebook to/from the url

wlach:
This allows crudely sharing iodide notebooks by allowing their content to
be copied into the clipboard as a pastable url.


wlach:
Here's what it looks like:

image

PR #660, opened 2018-06-15
isRowDF now rejects arrays of single values

hamilton:
Go to this ugly jsmd string for cases this solves:

```
?jsmd=%25%25%20meta%0A%7B%0A%20%20%22lastExport%22%3A%20%222018-06-15T22%3A38%3A43.590Z%22%0A%7D%0A%0A%25%25%20js%0A%5Bundefined%5D%0A%0A%25%25%20js%0A%5Bnull%5D%0A%0A%25%25%20js%0A%5B'yes!'%5D
```

### DIFF
--- a/src/tools/nb.js
+++ b/src/tools/nb.js
@@ -41,6 +41,7 @@ nb.sameKeys = function(x,y){
 
 nb.isRowDf = function(obj,rowsToCheck = 100){
   if (!_.isArray(obj) || obj.length==0) {return false}
+  if (obj.length === 1 && !_.isPlainObject(obj[0])) {return false}
   rowsToCheck = Math.min(rowsToCheck,_.size(obj))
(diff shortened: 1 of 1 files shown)


src/tools/nb.js
```diff
@@ -41,6 +41,7 @@ nb.sameKeys = function(x,y){
 
 nb.isRowDf = function(obj,rowsToCheck = 100){
   if (!_.isArray(obj) || obj.length==0) {return false}
+  if (obj.length === 1 && !_.isPlainObject(obj[0])) {return false}
```
wlach: Would probably mash this into the if statement above, since they return false. i.e.

```js
if (!_.isArray(obj) || obj.length==0 || (obj.length === 1 && !_.isPlainObject(obj[0])) {
  return false;
}
```

Also, some tests would be nice.

PR #661, opened 2018-06-15
Remove ctrl-e / ctrl-shift-e shortcuts for export

wlach:
They conflict with the browser binding for ctrl-e, which moves to the
end of line in a textbox (commonly used among emacs power users)


wlach:
Updated per discussion to add shift modifier to export notebook keybinding, and remove keybinding for "export notebook as report" entirely

PR #662, opened 2018-06-17
Update ipynb-import.js

teonbrooks:
fix importing error with pyodide plugin


src/tools/ipynb-import.js
```diff
@@ -54,7 +54,7 @@ Support for importing Jupyter notebooks into Iodide is experimental.  In particu
   "displayName": "python",
   "codeMirrorMode": "python",
   "keybinding": "p",
-  "url": "https://iodide-project.github.io/pyodide-demo/pyodide.js",
+  "url": "https://iodide.github.io/pyodide-demo/pyodide.js",
```
mdboom: I think this needs to be `iodide.io`, not `iodide.github.io`...

PR #688, opened 2018-06-22
Fix export to clipboard capitalization in ui

wlach:


hamilton:
Thanks for the PR @wlach!

PR #689, opened 2018-06-22
Add a react component for the user button

wlach:
I think this is a little easier to understand (and thus modify).

### DIFF
--- a/src/components/menu/user-button.jsx
+++ b/src/components/menu/user-button.jsx
@@ -1,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types';
+
+import AccountCircle from 'material-ui-icons/AccountCircle'
+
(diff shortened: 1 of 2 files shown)


src/components/menu/user-button.jsx
```diff
@@ -0,0 +1,31 @@
...
+import NotebookTaskButton from './notebook-task-button'
+import tasks from '../../actions/task-definitions'
+
+export default class UserButton extends React.Component {
+  static propTypes = {
+    isAuthenticated: PropTypes.bool.isRequired,
+  }
+
+  render() {
+    console.log('hi');
```
hamilton: hi! (this should be removed probably)

wlach: lol yes

PR #690, opened 2018-06-22
Enforce semi-colons in js and add them where missing

wlach:
There is no "right" answer to the question of whether semi-colons
are good or bad in modern-day js, but historically Mozilla's
web tools (and in particular our data tools) have used and enforced
a style where semicolons are always used to terminate the end of a line
or expression.

Redash, test-tube, metricsgraphics, missioncontrol for example all
use this rule. Furthermore, it's a default part of the airbnb
style guide which is recommended by Mozilla's frontend infrastructure
working group (https://github.com/mozilla-frontend-infra).

To avoid confusing people who have to move between these projects on a
frequent basis, I propose that we just adopt the rule and move on.
This commit adds the rule (really deleting an exception) and updates
the source code accordingly (done automatically via eslint's --fix
option)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,7 +15,6 @@ module.exports = {
   ],
   "extends": "airbnb",
   "rules": {
-    "semi": ["off"],
     // Restricting for..of seems pretty controversial, let's disable that.
     // See https://github.com/airbnb/javascript/issues/1271
     "no-restricted-syntax": ["off"],
--- a/src/actions/__tests__/actions.test.js
+++ b/src/actions/__tests__/actions.test.js
@@ -10,10 +10,10 @@ describe('temporarilySaveRunningCellID', () => {
 
 describe('updateAppMessages', () => {
   it('creates an object with a type: UPDATE_APP_MESSAGES and message payload', () => {
-    const message1 = { message: 'hello' }
+    const message1 = { message: 'hello' };
     expect(actions.updateAppMessages(message1).message.message).toEqual('hello');
     expect(actions.updateAppMessages(message1).message.details).toEqual('hello');
-    const message2 = { message: 'ok', details: '---', when: new Date().toString() }
+    const message2 = { message: 'ok', details: '---', when: new Date().toString() };
     expect(actions.updateAppMessages(message2).message.message).toEqual('ok');
     expect(actions.updateAppMessages(message2).message.details).toEqual('---');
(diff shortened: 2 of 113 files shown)


wlach:
@bcolloran @hamilton @djbarnwal @mdboom I apologize for bringing this up (I hate style wars like these and find them generally unproductive), but am genuinely finding it disruptive to have to change my coding style as I move through projects. I believe that making iodide's coding style consistent with other tools used by Mozilla's engineering operations team is the right way forward. If you have any objections let me know, otherwise I'd like to just land this soon and move on to more interesting topics. :)

hamilton:
@wlach though I don't like semicolons generally, I have no problem adopting a coding standard like this if it makes life easier for everyone else! I _would_ suggest, however, that perhaps we can do this after `iframe-2` is merged. This PR will change literally every file we have, and it'll likely create large swaths of merge conflicts as a result, and at any rate we're really hoping to not touch some of the stuff in `src` too much right now.

I'll let @bcolloran chime in, but tldr - happy to adopt whatever everyone else wants to do, but maybe let's wait until `iframe-2` is ready to merge and then do it.

bcolloran:
i agree that i've found it disruptive to have to tweak my normal coding style (always use single quotes! omg why?!!), but if we're going to make any changes along these lines, i'd rather have us go ahead and adopt 'prettier' and let it do the formating automatically so that you don't have to remember to omit the semi-colons, and i don't have to remember to put in semi-colons or always use single quotes, and no one has to remember to do anything other than maybe press the keybinding for the prettier plugin for their editor... or just do nothing and let the pre-commit hooks or CI take care of it :-)

does that sound like a reasonable option? care to look into the details?
https://prettier.io/docs/en/precommit.html
https://prettier.io/docs/en/eslint.html

i think we might need to change our eslint config a little bit and then have everyone install the plugin in their editors.

in any case, +1 re @hamilton 's point about the big iframe refactor -- this does feel like a potentially very disruptive set of changes to have to merge, so probably worth waiting for that

wlach:
I think using prettier is a good idea, but it's an additional step. Prettier deals with code formatting, whereas this particular issue is a linting one.

In fact, prettier actually requires the use of semi-colons to do its formatting, so adding an eslint rule to that effect would be an incremental step to adopting it.

I'm fine with deferring landing this until after the iframe refactoring lands.

PR #691, opened 2018-06-26
Fix user button

wlach:


wlach:
#689 was actually broken :( this fixes the problems.

Issue #673, opened 2018-06-19
Server-side notebook nits


wlach:
Some nits with the current server-side notebook implementation (just the notebook view with login to github + export to gist, not the more advanced version we're working on):

* You have to relogin every time you visit http://extremely-alpha.iodide.io (i.e. the github credentials don't persist)
* The current login button is not exactly intuitive -- it would be better if we just made it a standard button with "Log in" text or similar
* When logged in, clicking on your face logs you out, which is not exactly what I'd expect. At the least we could do something like what treeherder does (create a menu with a log out button):
![image](https://user-images.githubusercontent.com/20569/41629331-c50304fc-73dd-11e8-8825-2553a1c85ac5.png)

wlach:
The relogin issue is because we aren't storing the session id. Apparently with passport you typically want to store this on the server side of things. Should probably just wait on @djbarnwal's #677 landing before tackling that.

wlach:
Most of this is now fixed with the python server and #692. The only remaining issue is the behaviour of the avatar, which I'll look into once the client side churn has slowed down (e.g. the iframe-2 branch is merged)

wlach:
I'm going to close this and file a more specific issue to deal with the avatar.

PR #692, opened 2018-06-26
Use a textual button for login, and provide a tooltip

wlach:


wlach:
This is the low hanging fruit of #673, persistent logins need server side pesistence and (it seems like) menus would better wait until the frame refactoring lands.

bcolloran:
lgtm, thanks @wlach

PR #694, opened 2018-06-29
WIP: view panes in eval-frame, change scrolling, t

hamilton:
This is not ready to merge yet, but I wanted to get all my upcoming changes more visible.

- the declared variables / history panes now open in the eval-frame, though the css is still a bit difficult to tame
- scrolling is now set to "instant" while you're keying up or down, so all that movement is much less jarring
- tweaked out outputs look, especially ones that return `undefined`


bcolloran:
great stuff @hamilton, merging now. added my notes in the iframe sprint doc rather than here.

PR #696, opened 2018-07-04
Fixed submenu switching

oxmap:
Related to material_ui problem than submenu cannot easly be closed, there are a solution that @janhoeck provided. https://github.com/mui-org/material-ui/issues/11723

In his demo in subsection menu there is no iterated clicking. 

So i removed and checked. That's worked as expected.

### DIFF
--- a/src/components/menu/notebook-menu-subsection.jsx
+++ b/src/components/menu/notebook-menu-subsection.jsx
@@ -22,9 +22,6 @@ export default class NotebookMenuSubsection extends React.Component {
 
   handleClose() {
     this.setState({ anchorElement: null })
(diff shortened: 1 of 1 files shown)


hamilton:
Thank for the PR @oxmap! This does indeed make it possible to switch between the submenus, but it now also precludes clicking away from all the menus if I click outside of any of the menus / submenus, which makes for a difficult problem to solve with `material-ui`. Is there a way to enforce closing all menus if we click away from both menu & submenu?

So there's a tradeoff between these two. `material-ui` is simply not well-designed for these kind of basic desktop UI interactions, so we're considering migrating away from it and either using another library or just implementing our own.

PR #697, opened 2018-07-05
Add Cell Menu Container Tests

danielafcarey:
We noticed [cell-menu-container.js](https://github.com/iodide-project/iodide/blob/master/src/components/cells/cell-menu-container.js) had 0% coverage, so we created the test file and added 14 tests and got this file up to 94% coverage. 

The one line that says it is still not covered (line 112) is tested on line 155 in the cell-menu-container.test.js, so unsure why it says it is not covered. 

Passing Tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1210,13 +1210,13 @@
         "bytes": "3.0.0",
         "content-type": "~1.0.4",
         "debug": "2.6.9",
-        "depd": "1.1.2",
-        "http-errors": "1.6.3",
+        "depd": "~1.1.2",
+        "http-errors": "~1.6.3",
         "iconv-lite": "0.4.23",
-        "on-finished": "2.3.0",
+        "on-finished": "~2.3.0",
         "qs": "6.5.2",
         "raw-body": "2.3.3",
-        "type-is": "1.6.16"
+        "type-is": "~1.6.16"
       },
       "dependencies": {
         "http-errors": {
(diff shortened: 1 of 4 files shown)


bcolloran:
this is great, thanks for contributing @danielafcarey! One quick comment: we try to avoid snapshot testing b/c of maintainability issues in favor of testing the intent of code by checking to see that the correct subcomponents show up where we expect them to. Take a look at some of our other component tests to see how we approach this.

also in case it's of interest, i have found the following to be good guidance for how i write tests and think about unit testing in Iodide (I know we still have a long way to go, but i have found these very useful)
https://medium.freecodecamp.org/the-right-way-to-test-react-components-548a4736ab22
https://hackernoon.com/front-end-react-snapshot-testing-with-jest-what-is-it-for-7788f7bd5a2e
http://muness.blogspot.com/2008/04/testing-declarative-code.html
https://hackernoon.com/unit-testing-redux-connected-components-692fa3c4441c

(hmm i should add these notes to the contributing guidelines... :-) )

danielafcarey:
That makes sense! And thank you for the resources on testing - those are super helpful (especially [this one](https://medium.freecodecamp.org/the-right-way-to-test-react-components-548a4736ab22?gi=8687be53d427) for testing that the subcomponents are rendered correctly)! 

I am interested in continuing to work on this and updating these tests to be consistent with the rest of the Iodide tests. However, I may not be able to work on this for a few weeks, so I understand if you want to close this PR so it's not hanging around unfinished.

Also wanted to give the feedback that the documentation and contribution guidelines are stellar - thank you!

PR #698, opened 2018-07-06
Switch server implementation over to python

wlach:
For a bunch of reasons (familiarity inside Mozilla, maturity of ecosystem),
python/django is a better environment to build our server-side implementation
than node.js.

This PR supports both deployment to heroku and a local docker-compose based
set up. Still to do (probably in followups):

* More comprehensive documentation (just have some readme instructions on
  setting up a local dev environment for now)
* Tests


wlach:
@jezdez You said you wouldn't mind looking over this. This should be considered an imperfect work-in-progress, but I'd like to know if you see any big red flags.

wlach:
One known bug with this: login does not persist if you reload the page. This is actually a bug on the client side and existed with the previous node-based implementation as well. I will fix this in a followup.

wlach:
@jaredkerim actually had a look over it. He suggested I create a custom user model and move the avatar-getting code (which uses github's API) over to some middleware. Seemed like good suggestions so I landed them in a followup commit.

Theoretically I could squash all of this, but I think the history will be easier to understand as a progress so will probably be landing this as-is.

bcolloran:
:clap:

PR #706, opened 2018-07-13
migrate to stable releases for material-ui

hamilton:
I'm not sure how much longer we'll be staying on `material-ui`, but given that the Tooltips are about to be improved, this PR will upgrade off of beta to the latest stable, so the migration to the version that fixes the performance issues will be easier.

This wasn't too difficult to do, but @bcolloran it would be worth you checking all the UI elements to make sure there are no errors or warnings. I've done my best to remove most of these for now.


bcolloran:
@hamilton, still haven't had time to review this, and the version we need dropped over the weekend -- can you update to that (should be quicker now i hope)? https://github.com/mui-org/material-ui/releases

thanks!

bcolloran:
@hamilton nm, i'll update it just to streamline things a bit

Issue #748, opened 2018-08-09
once iframe2 is merge into master, implement autodeploy of static iframe assets


bcolloran:
we need to autodeploy the static assets from `iodide.app`. @mdboom i'll give you permissions to change the DNS settings however you need.

blocked by #747

mdboom:
This is deploying just fine to iodide-project/master-eval.  Setting up the redirect from `extremely-alpha.iodide.app` to this is hitting an HTTPS-related snag.  Working on it, and keeping this issue open...

mdboom:
This appears to be working now, though haven't tested end-to-end since we don't have the server and main frame assets deployed yet.

Issue #749, opened 2018-08-09
after iframe2 is merged into master, set up django server to autodeploy from master to `extremely-alpha.iodide.io`


bcolloran:
blocked by  #747
cc @mdboom 

once iframe is merged into master, we want to get the live  server up and running with the iframe. this will require a couple webpack tweaks as mentioned in the meeting

also @wlach, lmk what we level of heroku account we should have. i'll talk to matt_g and get us the upgrade.

wlach:
re: heroku accounts I think we can live with the free dyno for now, it doesn't take *that* long for the instance to spin up.

wlach:
I think we fixed this one!

Issue #674, opened 2018-06-19
extremely-alpha.iodide.io does not support https


wlach:
Apparently you need to upgrade to some kind of paid heroku plan to make this happen. We should probably do this sooner than later, as plain old http is pretty insecure.

bcolloran:
@wlach what's the cost of a plan that would support our current team and other needs? can we just get something for the whole team, or is this like a per-seat kind of thing?

i'm all in support of upgrading ASAP if it will be advantageous for us. 

i guess the other option might be to look into mozilla hosting of our heroku app. i know we had wanted to avoid dealing with ops for the time being to maintain control/flexibility/speed; but since we'll have to figure out deploying behind LDAP at some point anyway, maybe it'd be worth seeing what the internal options are? (just throwing this out there -- i'm happy to go with the external service if that seems easier and better for where we are)

bcolloran:
ok, sounds good :-)

bcolloran:
@wlach -- i'm now able to get http://extremely-alpha.iodide.io/ to load, an the eval frame loads at https://extremely-alpha.iodide.app/iodide.eval-frame.dev.html (tho as expected, it doesn't do anything when loaded as a standalone page), but loading  http://extremely-alpha.iodide.io/ gives the console message:

`Failed to execute ‘postMessage’ on ‘DOMWindow’: The target origin provided (‘https://extremely-alpha.iodide.io’) does not match the recipient window’s origin (‘http://extremely-alpha.iodide.io’).`

So we could just change the expected origin to plain http, but it seems better to pull the trigger on the if that will get us SSL upgrade.

quick links to the mozilla heroku mana pages if we want to go that way:
https://mana.mozilla.org/wiki/display/TS/Heroku
https://mana.mozilla.org/wiki/display/WebOps/Heroku

but if that seems like it'll end up being a lot of overhead, we can just upgrade the existing account. i'm happy to do it if you'd like

bcolloran:
upgraded to hobby, https working

PR #752, opened 2018-08-10
consolidate our webpack config

bcolloran:
@wlach -- with @mdboom having completed #748, can you jump in to get this over the line?

i think you should be able to build off this PR, which a bunch of changes to consolidate our webpack config and get things working and streamlined since most of our testing will be done using the django server going forward.

things to note:
- since we're not yet really versioning our resources, i don't think we need to have e.g. `iodide.X.js` for X = {master, stable, dev, iodide-server} all floating around. so this config just uses the 'dev' label everywhere other than tags (since really all this stuff is "dev" until we decide to get serious about versioning).

- we'll probably need to move some of the stuff about which origin a given instance is being served from to the `.env` file (which will need to be configured per origin anyway since we'll need different GH secrets etc). this way, the `.env` file can pass the correct origin into the python and webpack templates.

i think other changes will be needed, but this should be a start.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8964,9 +8964,9 @@
       }
     },
     "minimist": {
-      "version": "0.0.8",
-      "resolved": "https://registry.npmjs.org/minimist/-/minimist-0.0.8.tgz",
-      "integrity": "sha1-hX/Kv8M5fSYluCKCYuhqp6ARsF0=",
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.0.tgz",
+      "integrity": "sha1-o1AIsg9BOD7sH7kU9M1d95omQoQ=",
       "dev": true
     },
     "mississippi": {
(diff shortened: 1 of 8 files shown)


bcolloran:
WIP towards closing #749

wlach:
@bcolloran hey, so I think I have this mostly working, except that I'm not sure where I should be getting the assets from extremely-alpha.iodide.io?

https://extremely-alpha.iodide.app/iodide.eval-frame.dev.html

returns a 404

It was working ok when everything was served from the same origin on heroku. :)

bcolloran:
hmm ok. i thought @mdboom had gotten this with https://github.com/iodide-project/iodide/issues/748

bcolloran:
ah i see, i think we need to merge to master so that the other changes in the webpack config are picked up by mike's build script. so i'm going to go ahead and merge this because now that we're deploying off stable, doing so won't break our public site, yay :-)

bcolloran:
commented at https://github.com/iodide-project/iodide/issues/674#issuecomment-412746383 -- i think we want https for the heroku instance. then i _think_ we'll be good to go.

bcolloran:
upgraded to heroku hobby instance, all good now

PR #755, opened 2018-08-14
Cell Cut Copy Paste Highlight for Iframe2

djbarnwal:
Ports features of PR #629 into Iframe 2
Implements cut, copy, paste, highlight, unhighlight, actions using multiple-highlight (for ex. skip in Run all at once for multiple cells)

### DIFF
--- a/src/components/cells/cell-container.jsx
+++ b/src/components/cells/cell-container.jsx
@@ -5,7 +5,13 @@ import Tooltip from '@material-ui/core/Tooltip'
 
 import UnfoldLess from '@material-ui/icons/UnfoldLess'
 
-import { selectCell, updateCellProperties } from '../../actions/actions'
+import {
+  selectCell,
+  updateCellProperties,
+  highlightCell,
+  unHighlightCells,
+  multipleCellHighlight,
+} from '../../actions/actions'
 import { getCellById } from '../../tools/notebook-utils'
 
 import CellMenuContainer from './cell-menu-container'
@@ -28,6 +34,9 @@ export class CellContainerUnconnected extends React.Component {
     // action props
     selectCell: PropTypes.func.isRequired,
     updateCellProperties: PropTypes.func.isRequired,
+    highlightCell: PropTypes.func.isRequired,
+    unHighlightCells: PropTypes.func.isRequired,
+    multipleCellHighlight: PropTypes.func.isRequired,
   }
 
   constructor(props) {
@@ -36,10 +45,20 @@ export class CellContainerUnconnected extends React.Component {
     this.editor = React.createRef()
   }
 
-  handleCellClick = () => {
-    const scrollToCell = false
-    if (!this.props.selected) {
-      this.props.selectCell(this.props.cellId, scrollToCell)
+  handleCellClick = (event) => {
+    if (!event.target.classList.contains('CodeMirror-line')) {
+      const scrollToCell = false
+      if (!this.props.selected && !event.shiftKey) {
+        this.props.selectCell(this.props.cellId, scrollToCell)
+      }
+      if (event.shiftKey) {
+        event.preventDefault();
+        this.props.multipleCellHighlight(this.props.cellId)
+      } else if (event.ctrlKey || event.metaKey) {
+        this.props.highlightCell(this.props.cellId)
+      } else {
+        this.props.unHighlightCells()
+      }
     }
   }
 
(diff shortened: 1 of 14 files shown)


src/components/cells/cell-container.jsx
```diff
@@ -36,10 +45,20 @@ export class CellContainerUnconnected extends React.Component {
     this.editor = React.createRef()
   }
 
-  handleCellClick = () => {
-    const scrollToCell = false
-    if (!this.props.selected) {
-      this.props.selectCell(this.props.cellId, scrollToCell)
+  handleCellClick = (event) => {
+    if (!event.target.classList.contains('CodeMirror-line')) {
```
bcolloran: is this conditional meant to prevent clicks in the text editor from enabling highlighting? if so, it's not really working for me (clicks in the text editor are still enabling the highlighting). if that is the intent, we could just limit highlight clicks to the  cell header (since that would be pretty much the only click area left). You could  create a new handler `handleCellHeader` and attach it to the cell-header div.

But i'm not 100% sure what your objective was here, so that may not work depending on what you were trying to accomplish

src/components/cells/cell-editor.jsx
```diff
@@ -200,6 +208,7 @@ function mapStateToProps(state, ownProps) {
     lineNumbers: true,
     keyMap: 'sublime',
     comment: codeMirrorMode === 'javascript',
+    readOnly: getReadOnly(),
```
bcolloran: any reason you define this function then call it immediately? would it work just set the value of `readOnly` directly, or am i missing something?

src/reducers/cell-reducer.js
```diff
@@ -60,8 +182,17 @@ const cellReducer = (state = newNotebook(), action) => {
         { cells: moveCell(state.cells, getSelectedCellId(state), 'down') },
       )
 
-    case 'UPDATE_CELL_PROPERTIES':
+    case 'UPDATE_CELL_PROPERTIES': {
+      const approvedMultipleChanges = new Set(['skipInRunAll'])
```
bcolloran: the const `approvedMultipleChanges` can go at the top of the file (it's kind of a global) and it should have a brief comment describing what it is. also i think it can probably just be a plain list rather than a set.

src/reducers/cell-reducer.js
```diff
@@ -48,6 +52,124 @@ const cellReducer = (state = newNotebook(), action) => {
...
+      const thisCell = cells[index]
+      if (action.revert) thisCell.highlighted = !thisCell.highlighted
+      else thisCell.highlighted = true
+      nextState = Object.assign({}, state, { cells })
+      return nextState
+    }
+
+    case 'UNHIGHLIGHT_CELLS': {
+      const cells = state.cells.slice()
+      cells.forEach((c) => { c.highlighted = false }) // eslint-disable-line
```
bcolloran: if you're disabling eslint, specify the rule being disabled

bcolloran: in this case, you can avoid disabling the eslint and skip the `cells.slice` above using a map (which accomplishes the same copying as the slice). something along the lines of:
```
       const cells = state.cells.map((c) =>Object.assign({}, c, { highlighted: false })) 
```

src/reducers/cell-reducer.js
```diff
@@ -48,6 +52,124 @@ const cellReducer = (state = newNotebook(), action) => {
...
+
+    case 'MULTIPLE_CELL_HIGHLIGHT': {
+      const cells = state.cells.slice()
+      const index1 = getSelectedCellIndex(state)
+      const index2 = cells.findIndex(c => c.id === action.id)
+      const low = Math.min(index1, index2)
+      const high = Math.max(index1, index2)
+      cells.forEach((c, index) => {
+        if (low <= index && index <= high) {
+          c.highlighted = true // eslint-disable-line
```
bcolloran: see previous comment

src/reducers/cell-reducer.js
```diff
@@ -48,6 +52,124 @@ const cellReducer = (state = newNotebook(), action) => {
...
+        } else {
+          c.highlighted = false // eslint-disable-line
```
bcolloran: see previous comment

src/editor-state-prototypes.js
```diff
@@ -111,6 +112,14 @@ const stateSchema = {
...
       additionalProperties: languageSchema,
     },
+    copiedCells: {
+      type: 'array',
+      items: cellSchema,
+    },
+    cutCells: {
+      type: 'array',
+      items: cellSchema,
+    },
```
bcolloran: is it necessary to have both copiedCells and cutCells (rather than, for example, just a single cellClipboard)? it may well be necessary, and perhaps i havenot understood the code deeply enough to see why both are needed.

bcolloran:
@djbarnwal can you push your local branch of this again?

PR #760, opened 2018-08-15
Add https redirect for production deploys

wlach:
The new iframe branch of iodide expects an https url when running in
production.

Automatically redirect by default to https version of server when running in
non-debug mode, or when set by configuration variable (SECURE_SSL_REDIRECT)


wlach:
This seems to work great on our test server, asking for review mainly to make you more familiar with how I set things up.

Issue #734, opened 2018-08-03
[Feature Request] overlay help pane


teonbrooks:
It would be great if the user could press `h` or `?` and an overlay of all the keybindings could be listed. It's a super helpful tool and it would greatly reduce lookup time or doc hunting if this was available

bcolloran:
per #233, this pane should also link to our GH, api docs, other relevant links; it should app version, etc

bcolloran:
re #763, this pane can include the "file a bug" link that has been removed from the menu

PR #763, opened 2018-08-17
Remove "file an issue" menu entry

wlach:
As far as I'm aware, no one has used this feature to actually file an issue
and it's additional clutter in an already crowded menu.


bcolloran:
I'm gonna tweak this slightly -- +1 that we don't need this in the crowded menu, but we can hang onto the "file an issue" usertask for use in the help pane that is close to the top of the priority queue #734

PR #765, opened 2018-08-20
717 console eval area

bcolloran:
@wlach or @djbarnwal or @teonbrooks or @mdboom -- this PR implements a code entry area at the bottom of the console. I don't need review from all of you of course, but if one of you could give this a quick QA and pass over the code to see that it looks ok, that would be great.

This is the baseline mvp version of the console feature, just REPL entry and key back/forward through history entries. Still need to do something to enable selecting other language evaluators (which can be filed as a second issue). flag any merge blocker in the PR, and then also of course feel free to file any issues for feature requests on top of this baseline.

thanks!

ps @hamilton: i think this will be merged before you get back, but please take a look over this when you do get back to make sure that nothing here will break around promises. I tried to copy the existing pattern, but i'm still not sure how the eval queue really works

### DIFF
--- a/src/eval-frame/actions/actions.js
+++ b/src/eval-frame/actions/actions.js
@@ -107,6 +107,67 @@ export function updateUserVariables() {
   }
 }
 
+export function updateConsoleText(consoleText) {
+  return {
+    type: 'UPDATE_CONSOLE_TEXT',
+    consoleText,
+  }
+}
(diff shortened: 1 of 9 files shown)


teonbrooks:
@bcolloran just texted and it worked how I would expect it to. lgtm

PR #770, opened 2018-08-21
Server login mvp

wlach:
Minimal code to create a homepage ('/') location where user can log in
Actual listing of notebooks, etc. to come after.

### DIFF
--- a/server/templates/home.html
+++ b/server/templates/home.html
@@ -1,0 +1,8 @@
+{% extends "base.html" %}
+{% block content %}
+<div id="page"></div>
+<script>
+  window.userData = {{ user_info|safe }}
+</script>
+<script src="/server.home.dev.js"></script>
+{% endblock %}
--- a/server/views.py
+++ b/server/views.py
@@ -20,8 +20,7 @@ def get_user_info_dict(user):
 
 
 def index(request):
-    # basically just an empty placeholder for now
-    template = loader.get_template('notebook.html')
+    template = loader.get_template('home.html')
     return HttpResponse(template.render({
         'user_info': json.dumps(get_user_info_dict(request.user))
     }, request))
--- a/src/server/components/header.jsx
+++ b/src/server/components/header.jsx
@@ -1,0 +1,88 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import { withStyles } from '@material-ui/core/styles';
+import AppBar from '@material-ui/core/AppBar';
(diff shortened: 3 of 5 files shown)


wlach:
Some notes on this implementation:

* Everything is rendered server side as described in [this article](https://hackernoon.com/reconciling-djangos-mvc-templates-with-react-components-3aa986cf510a), this is not intended to be a single page web application. I believe this will simplify routing and a few other things.
* Still using react and material ui, mainly for consistency with the notebook itself. 
* *not* planning on using redux for the server (at least at this stage), as I think it's overkill for the very simple views we want to show. Instead, we can just pass state down from django to the app. If things get too complicated we can use the [context api](https://daveceddia.com/context-api-vs-redux/)

Nice things about this approach:
* Everything is fast to load
* All the client side code can be generated with one webpack file
* Lower complexity and learning curve. Hopefully changes will also be easier to review

server/templates/home.html
```diff
@@ -0,0 +1,8 @@
+{% extends "base.html" %}
+{% block content %}
+<div id="page"></div>
```
bcolloran: following from that hackernoon article -- would it make sense to have multiple react mount points for different page sections? E.g., set up the header to always mount at `<div id="header">", and then have a thing like:
```
<div id="page-content">
        <!-- Contents get replaced by mounted React.Component -->
        <i class="fa fa-lg fa-spinner fa-spin"></i><br><br>
        <i class="pending">Loading components...</i><br><br>
</div>
```

which they use to show a spinner and then be the mounting point for the main page content?

I don't know if this would be better, or if it's preferable to mount everything from a single mount point. (mostly i'm just trying to understand how this works, so i'm trying to ask some questions to see if i'm getting it at all :-) )

bcolloran:
this all seems pretty clear and understandable, so i don't really have much in the way of real review.

All of your points make sense to me. Not maintaining a Redux store seems fine since we can manage that state on the server.

I'm R+ on this, merge when you wish

server/templates/home.html
wlach: Hmm, had to think about this a bit. I'm pretty sure that functionally the approaches would be similar in terms of behavior. But intuitively I think I prefer it where each "view" is its own top-level component and all the composition happens on the client side. It's a little bit simpler and easier to understand.

wlach:
Going to merge this and move on to implementing real functionality. :)

PR #774, opened 2018-08-23
Schema dedupe

bcolloran:
this PR does two main things:
1) it dedupes the state schemas and makes explicit which parts of the state need to be _mirrored_ from the editor to the eval-frame and which parts of state belong only to one side or the other. this will reduce copy paste errors
2) it removes the state creation functions in favor of specifying default values in the state schemas.

to keep this PR of manageable size, for the time being i've just proxied the new functions through the existing files  src/editor-state-prototypes.js and  src/eval-frame/eval-frame-state-prototypes.js ; in a later PR we can go through the files that look to those and update paths to remove that proxying


bcolloran:
this will fix #730

PR #776, opened 2018-08-23
Address some fixmes

bcolloran:
you can step through the commits in this PR to see comments about individual FIXMEs

### DIFF
--- a/src/editor-state-prototypes.js
+++ b/src/editor-state-prototypes.js
@@ -127,9 +127,9 @@ const stateSchema = {
       type: 'array',
       items: { type: 'string' },
     },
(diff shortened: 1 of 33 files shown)


src/editor-state-prototypes.js
```diff
@@ -127,9 +127,9 @@ const stateSchema = {
       type: 'array',
       items: { type: 'string' },
     },
-    lastSaved: {}, // FIXME change to string ONLY with default 'never'
-    lastExport: {}, // FIXME change to string ONLY
-    sidePaneMode: {}, // FIXME change to string ONLY
```
mdboom: Should this be `{ type: 'string' }`?  Or is the FIXME obsolete and just being removed?

Issue #781, opened 2018-08-23
META: reps (output handler) issues


mdboom:
This issue to track output handler changes, particularly the move to use `react-inspector` in more places.

I've created a test notebook that tries to cover all of the basics of rep handling.

Here is the current behavior:

![image](https://user-images.githubusercontent.com/38294/44553000-1e964700-a6fa-11e8-9721-b10837023e8b.png)

![image](https://user-images.githubusercontent.com/38294/44553028-38378e80-a6fa-11e8-8197-1acf40b26728.png)

![image](https://user-images.githubusercontent.com/38294/44553046-47b6d780-a6fa-11e8-8af0-e780060bae35.png)

![image](https://user-images.githubusercontent.com/38294/44553058-53a29980-a6fa-11e8-9f82-fa1842616db3.png)

![image](https://user-images.githubusercontent.com/38294/44553130-8b114600-a6fa-11e8-892c-8ca10cd53b5e.png)

![image](https://user-images.githubusercontent.com/38294/44553144-95334480-a6fa-11e8-82c2-509f62727f30.png)

Here's the result of moving almost everything (except custom renderers) to react-inspector:

![image](https://user-images.githubusercontent.com/38294/44553230-cf044b00-a6fa-11e8-9c41-c16d51e14500.png)

![image](https://user-images.githubusercontent.com/38294/44553243-d9bee000-a6fa-11e8-83eb-872a27d8f0fe.png)

![image](https://user-images.githubusercontent.com/38294/44553260-e93e2900-a6fa-11e8-914f-03b63c2c3af1.png)

![image](https://user-images.githubusercontent.com/38294/44553283-f65b1800-a6fa-11e8-839c-eb715e731f0c.png)

![image](https://user-images.githubusercontent.com/38294/44553296-007d1680-a6fb-11e8-913d-2e9f35c1e06c.png)

mdboom:
(Sorry about the screenshots and different scales -- Firefox screenshots doesn't seem to work with iframes).

Some observations:

- [ ] For the basic Javascript types, `react-inspector` works pretty well.  It takes up less vertical space, and is more self-consistent.  This is an easy win and will remove tons of our own code :)
- [ ] In the current version, there is code to not display an entire string when it is too long.  That seems to be broken.
- [ ] `react-inspector` doesn't seem to be doing any truncation or folding of long arrays or strings, either
- [ ] For multi-dimensional arrays and tables, our existing reps are much better than `react-inspector`
[...]

PR #794, opened 2018-08-24
Use react-inspector for reps in most cases

mdboom:
See #781 for screenshots and justification.

This changes most of the basic reps and the object reps to use react-inspector.  They are generally more vertically-space-efficient and self-consistent than what we had before.

The array, matrix, data frame, error, promise and custom reps are retained from before since react-inspector doesn't handle those.  In separate PRs, these will get updated to be more stylistically consistent with what react-inspector does, but for now I think we merge this because it doesn't really regress anything.

Of special note is that react-inspector's object inspector is able to display iodide custom reps as values.  Conversely, our collection reps use react-inspector's reps when appropriate.


mdboom:
I think the test failures are server related...

src/components/reps/value-renderer.jsx
```diff
@@ -122,39 +111,34 @@ const complexHandlers = [
...
-
+let handlers = simpleHandlers.concat(complexHandlers)
 
 const userHandlers = []
 
 export function addOutputHandler(handler) {
   // insert new handlers *after* the scalar handlers
-  userHandlers.unshift(wrapHandler(handler))
-  handlers = simpleTypeHandlers.concat(userHandlers, complexHandlers)
+  userHandlers.unshift(handler)
```
bcolloran: why'd you remove the wrapHandler here? don't we especially need to wrap user-added handlers in try/catch?

src/components/reps/value-renderer.jsx
```diff
@@ -32,10 +25,13 @@ export function renderValue(value, inContainer = false) {
       } else if (resultElem.type !== undefined) {
         return resultElem
       }
-      console.warn(`Unknown output handler result type from ${handler}`)
+      console.warn(`Unknown out handler result type from ${handler}`)
```
bcolloran: "out handler"?

bcolloran:
big improvement over json tree! couple other thoughts

- as mentioned in #781, the long string rep does work for me, so perhaps we should not cut that yet?
- when i was hacking on reps last month, i found this array snippet useful as a test case:
```
// arrays with mixed types
let array = []
for (let i = 0; i < 100; ++i) {
  array.push(i)
  array.push('_"'+i)
  array.push(undefined)
  array.push(false)
  array.push(NaN)
  array.push(null) 
  array.push(i*Math.random())
  array.push(() => i)
}
array
```

- even tho the tests pass, i see this when i run them:
```
 PASS  src/components/reps/__tests__/promise-handler.test.jsx
  ● Console

    console.error src/components/reps/value-renderer.jsx:84
      output handler error TypeError: Cannot read property 'shouldHandle' of undefined
          at Object.shouldHandle (/home/bcolloran/iodide/iodide/src/components/reps/value-renderer.jsx:139:87)
          at renderValue (/home/bcolloran/iodide/iodide/src/components/reps/value-renderer.jsx:123:1082)
          at PromiseRep.render (/home/bcolloran/iodide/iodide/src/components/reps/promise-handler.jsx:40:129)
          at ReactShallowRenderer._updateClassComponent (/home/bcolloran/iodide/iodide/node_modules/enzyme-adapter-react-16/node_modules/react-test-renderer/cjs/react-test-renderer-shallow.development.js:238:39)
          at ReactShallowRenderer.render (/home/bcolloran/iodide/iodide/node_modules/enzyme-adapter-react-16/node_modules/react-test-renderer/cjs/react-test-renderer-shallow.development.js:110:12)
          at Updater.enqueueSetState (/home/bcolloran/iodide/iodide/node_modules/enzyme-adapter-react-16/node_modules/react-test-renderer/cjs/react-test-renderer-shallow.development.js:321:20)
          at PromiseRep.Object.<anonymous>.Component.setState (/home/bcolloran/iodide/iodide/node_modules/react/cjs/react.development.js:238:16)
          at PromiseRep.state.promise.props.promise.then.val (/home/bcolloran/iodide/iodide/src/components/reps/promise-handler.jsx:16:12)
          at <anonymous>
          at process._tickCallback (internal/process/next_tick.js:188:7)
```

src/components/reps/value-renderer.jsx
mdboom: Sorry -- this was just left in after experimenting with something...

src/components/reps/value-renderer.jsx
mdboom: Just an oops.  Reverted.

mdboom:
I was able to get the tests to work by forcing `ValueRenderer` to import before the specific output handlers being tested.  Not sure if there's a better solution.

mdboom:
I've updated this to reinstate the old string rep, but adjusted the styles (and how it gets used) to blend in with react-inspector.

I also discovered some CSS styles that are obsoleted by this PR and I removed those.

hamilton:
Hmm - it seems something else is at play here now. I'm finding that long strings still break the console, so it might actually not be the rep itself but the console not having `overflow-x` in the return value DOM element set properly. I know it used to be before the iframe migration.

This causes an overflow in the console, as far as I can tell.

```javascript
var randomString = (length=100) => {
  var text = "";
  var possible = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";

  for (var i = 0; i < length; i++)
    text += possible.charAt(Math.floor(Math.random() * possible.length));

  return text;
}

let a = randomString(1000)
a
```

mdboom:
I see that, too.  I thought I had fixed that exact issue -- I'm looking into it.

hamilton:
@mdboom this is great! The string overflow issue has been fixed w/ the latest commit, and the string-wrapping + truncation is MUCH more useful for debugging large strings (which users will inevitably encounter when they `fetch` a csv without parsing it).

This appears ready to merge to me.

PR #754, opened 2018-08-14
Add keybindings for resizing frames

djbarnwal:
This implements `Ctrl+left` and `Ctrl+right` keybindings which help you resize the two frames in steps of 0%, 33%, 50%, 66% and 100% .

Closes Issue #726

PR #801, opened 2018-08-24
 Add keybindings for resizing frames

djbarnwal:
This implements `Ctrl+left` and `Ctrl+right` keybindings which help you resize the two frames in steps of 0%, 33%, 50%, 66% and 100% .

(Mirror of #754)


djbarnwal:
@hamilton This PR is ready for another review.

src/eval-frame/actions/eval-frame-tasks.js
```diff
@@ -39,6 +39,20 @@ tasks.closePanes = new UserTask({
   },
 })
 
+tasks.increaseEditorWidth = new UserTask({
+  title: 'Set Editor Size',
+  keybindings: ['ctrl+right'],
```
hamilton: this should be `ctrl+shift+right` as well - if I am focused on the iframe I lose the keybinding

src/port-to-eval-frame.js
```diff
@@ -25,6 +25,8 @@ const approvedKeys = [
   'ctrl+d',
   'ctrl+h',
   'ctrl+i',
+  'ctrl+left',
```
hamilton: same as above - should be `ctrl+shift+right` etc.

djbarnwal:
@hamilton I think the Help modal error is due to some other commit. I posted an issue regarding the same ( #840 ) few hours ago.

src/components/editor-pane-container.jsx
```diff
@@ -53,15 +54,15 @@ class EditorPaneContainer extends React.Component {
...
+          minWidth={this.props.paneMinWidth}
           onResizeStop={this.onResizeStopHandler}
           size={{ width: this.props.cellResizerWidth }}
           style={{
             display: this.props.cellResizerDisplayStyle,
             flexDirection: 'column',
           }}
         >
-          <div id="cells">
+          <div style={this.props.cellsStyle} id="cells">
```
hamilton: @bcolloran do you have any thoughts on this approach?

bcolloran: @hamilton seems reasonable enough to me, do you have any particular concern?

bcolloran:
look good to me. i'm happy to merge if @hamilton is ok with the ` style={this.props.cellsStyle} ` part

PR #806, opened 2018-08-27
Make error rep more consistent

mdboom:
This makes the font slightly smaller to be more in line with the react-inspector reps, and removes the extra vertical space before and after.

This is a follow-on to #794.  Probably should hold off reviewing until #794 is merged.


mdboom:
I'm closing this -- this change will basically happen "by accident" in a better way in #817.

PR #811, opened 2018-08-28
Make notebook code built with `npm run start` have server features

wlach:
For example, the user panel / authentication. This will allow testing
of the "full stack" (both server and client features) with the
recommended workflow described in README.md.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "start": "webpack --watch --env=dev",
     "build": "webpack",
     "build-production": "webpack --env=production",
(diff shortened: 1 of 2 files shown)


wlach:
This is basically a bug fix to make the flow described in the README (https://github.com/iodide-project/iodide/blob/master/README.md#server-mode) work as advertised.

wlach:
Yeah, I think we ought to refactor this once we're far enough along with the server that we no longer need the ghpages standalone version.

PR #812, opened 2018-08-28
Improve matrix rep

### LINKED ISSUE #798: Make matrix rep more consistent / support custom reps
Unlike our other collection reps (dataframe and array), the matrix rep doesn't use our other reps to render each individual value.  This makes them stylistically inconsistent, and also we lose the ability to handle custom reps.

If easy to make the `pretty-matrix` library support custom renderers for each element, we should do that.  Short of that, we should adjust the css so it matches what `react-inspector` does for numeric scalars.

mdboom:
Fixes #798.

This improves the matrix rep:

- It can now display any rep (not just numeric types).  The "short" version of the value is displayed in each place.
- The types are displayed in a style/color/etc. that's consistent with elsewhere.
- As part of making this work, the number rep has been reinstated, and it displays itself in short or long form.
- I'm also dealing with getting the default react-inspector CSS inserted not by using the `Inspector` react component, but by copying it ito `.history-item-output`.  I think this is better (you can see the simplification in the string rep), even if it's a bit wonky.  `react-inspector` doesn't use real CSS but just generates it all from Javascript at runtime.


mdboom:
Thanks.  I think those should be easy to fix, so I might as well try to do them in this PR.

mdboom:
Also noticed from your screenshot that the string rep seems to have a bug.  It's displaying the same content before and after the "X characters ommitted..." slug.

mdboom:
I think I've fixed the long date and string issues:

image

Short strings no longer display "... X characters omitted ...", but just have a trailing "..."

Short datetimes are now in ISO format, with a convenient spot for a line break between the date and time.

mdboom:
I see that space issue, too.  I wonder if a very thin and light line would help or make things worse.  But, yes, in general CSS tweaks can be deferred (and probably to someone who understands design and CSS better than I do...)

hamilton:
Sounds good. I think a 2-3px padding on the matrix entries would probably solve the issue, but let's address that later. This is a big improvement! Will merge.

PR #817, opened 2018-08-29
Use DevExtreme React Grid rather than React Table for dataframe rep

mdboom:
(This is a follow-on for #812, and includes/requires its rep improvements).

This uses [DevExtreme React Grid](https://devexpress.github.io/devextreme-reactive/react/grid/) instead of [React Table](https://react-table.js.org/#/story/readme).

The main improvement is the use of "virtual scrolling" rather than pagination.  Having two levels of scrolling (one for the table and one for the console) is a little annoying, but it feels better than having pagination.  I suspect that comes down a matter of personal preference, however.  Note that React Grid [also supports pagination](https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/paging/) so we could provide either as a user preference, potentially.

Anyway, hopefully this PR serves as a proof-of-concept.  It's hopefully at feature parity to what we had before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,9 @@
     ]
   },
   "dependencies": {
+    "@devexpress/dx-react-core": "^1.7.0",
+    "@devexpress/dx-react-grid": "^1.7.0",
+    "@devexpress/dx-react-grid-material-ui": "^1.7.0",
     "@material-ui/core": "^1.4.0",
     "@material-ui/icons": "^1.1.0",
     "@skidding/react-codemirror": "^1.0.0",
@@ -55,7 +58,6 @@
     "react-helmet": "^5.2.0",
     "react-inspector": "^2.3.0",
     "react-redux": "^5.0.7",
-    "react-table": "^6.8.2",
     "redux": "^3.7.2",
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.2.0",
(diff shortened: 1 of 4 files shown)


hamilton:
Great start! I had a couple of thoughts based off of dogfooding it.

screen shot 2018-08-31 at 10 20 52 am

- I think the table could be twice as tall so we can cover ~10 or so entries by default. I'm thinking of the optimal number from `head` gives an analyst enough to get a sense for the shape of the data.
- the current data tables also allow me to expand a column if I need to see more. This is particularly useful for long strings (in the example above, seeing more of the `area_name` field might be useful). I could see it also being useful when one of the entries is an array, matrix, a nested data frame, or even a custom rep (like the geo rep you have in another example).
- I'm of two minds about pagination. On the one hand, it's pretty useful to be able to quickly paginate through pages with a button, and it makes it easy to make a clean 10 or so entries by default. I've used it to great effect while exploring a table. On the other, it's sort of a solution designed in the past, and perhaps infinite scrolling as a more modern paradigm is more useful. I think I'm still tending toward pagination.

mdboom:
> I think the table could be twice as tall so we can cover ~10 or so entries by default. I'm thinking of the optimal number from head gives an analyst enough to get a sense for the shape of the data.

This is easy.  Completely tweakable -- though not after it's created as far as I can tell.

> the current data tables also allow me to expand a column if I need to see more. This is particularly useful for long strings (in the example above, seeing more of the area_name field might be useful). I could see it also being useful when one of the entries is an array, matrix, a nested data frame, or even a custom rep (like the geo rep you have in another example).

This might be tricky -- I'll investigate.

> I'm of two minds about pagination...

Yeah, it's tricky...  Both are possible with this widget.  I'll probably make another PR with pagination so we can compare react grid's two modes directly.

PR #820, opened 2018-08-29
Notebook revision fixes

wlach:
Be sure that the title property of a notebook gets updated when a new
revision is sent, and that we always associate the latest revision
with the notebook when requesting the notebook by itself.

### DIFF
--- a/server/notebooks/serializers.py
+++ b/server/notebooks/serializers.py
@@ -6,8 +6,7 @@ from server.base.models import User
 
 class NotebookLatestRevisionField(serializers.RelatedField):
     def get_attribute(self, obj):
-        return NotebookRevision.objects.filter(notebook_id=obj.id).order_by(
-            '-created').first()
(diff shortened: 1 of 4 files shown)


server/notebooks/serializers.py
```diff
@@ -6,8 +6,7 @@
 
 class NotebookLatestRevisionField(serializers.RelatedField):
     def get_attribute(self, obj):
-        return NotebookRevision.objects.filter(notebook_id=obj.id).order_by(
-            '-created').first()
+        return NotebookRevision.objects.filter(notebook_id=obj.id).last()
```
wlach: We can rely on .last() providing the latest revision for a notebook because we define a sorting order in the notebook model (thanks to @jaredkerim for that trick):

https://github.com/iodide-project/iodide/blob/31bdd4022306c996a63829d76aacf6c72a76b69c/server/notebooks/models.py#L50

wlach:
Actually, probably virtually any of you could review this relatively minor set of changes. :)

PR #823, opened 2018-08-29
Fix #818: Support multi-line strings in iodide.output.text

### LINKED ISSUE #818: newlines are not correctly handled with custom evaluator
My evaluator returns a string with multiple lines, but they are ignored in the notebook output. I logged the exact same string in the console and it worked flawlessly.

I find it strange that they are displayed in a `<span class="array-output">` block, but don't know if it is related.

mdboom:
This has been updated to also support XML tag characters so that `output.text('<b>')` displays `<b>`.


wlach:
Stealing this review, it looks fine to me.

PR #831, opened 2018-09-03
Typo fix in README

donkirkby:
The project sounds very exciting. Here's a small fix I noticed while reading about the project.


wlach:
Thanks for the fix! travis is failing, but that obviously has nothing to do with this patch.

PR #832, opened 2018-09-03
Cell Cut Copy Paste Take 3

djbarnwal:
Cell Cut Copy Paste Take 3. Refer [PR#755](https://github.com/iodide-project/iodide/pull/755) and [PR#629](https://github.com/iodide-project/iodide/pull/629)

The review on PR #755 has been taken into consideration with the exception of Esc key unhighlighting all cells. Esc key is already bound for changing cell mode. 

@bcolloran With the exception of Esc key implementation, the PR is ready for another round of review.

Also, cut-cut and cut-copy sequences delete cells which cannot be recovered later. Do we want to recover these cells for these sequences or rather implement Ctrl+Z undo functionality to handle it?


src/actions/task-definitions.js
```diff
@@ -88,6 +89,38 @@ tasks.moveCellDown = new UserTask({
...
 
+tasks.highlightUp = new UserTask({
+  title: 'Highlight Cell Above',
+  displayKeybinding: 'Shift+Up', // '\u21E7 \u2191',
+  keybindings: ['shift+up'],
+  keybindingPrecondition: isCommandMode,
+  preventDefaultKeybinding: true,
+  callback() {
+    const cellAboveId = getCellAboveSelectedId()
+    dispatcher.highlightCell(getSelectedCellId(store.getState()), false)
```
bcolloran: this `store.getState()` should be moved inside of the highlightCell action creator (we're trying to get better about import the redux store in as few places as possible, it makes testing more difficult)

src/actions/task-definitions.js
```diff
@@ -88,6 +89,38 @@ tasks.moveCellDown = new UserTask({
...
+
+tasks.highlightDown = new UserTask({
+  title: 'Highlight Cell Down',
+  displayKeybinding: 'Shift+Up', // '\u21E7 \u2193',
+  keybindings: ['shift+down'],
+  keybindingPrecondition: isCommandMode,
+  preventDefaultKeybinding: true,
+  callback() {
+    const cellBelowId = getCellBelowSelectedId()
+    dispatcher.highlightCell(getSelectedCellId(store.getState()), false)
```
bcolloran: same as above--
this `store.getState()` should be moved inside of the highlightCell action creator (we're trying to get better about import the redux store in as few places as possible, it makes testing more difficult)

bcolloran:
(ps very nice work @djbarnwal! this is a complex feature with a lot of weird corner cases, you did a great job)

PR #833, opened 2018-09-04
Make closing console pane from the GUI work again (fixes #815)

### LINKED ISSUE #815: Closing Console/DV panes using cross button throws error
Closing Console/DV/App History pane by clicking the cross button does not work. The header gets changed to `PANE MODE UNDEFINED!!` while the pane is still there. Using keybindings such as `Ctrl+H` etc. works as expected.

wlach:


src/eval-frame/actions/eval-frame-tasks.js
```diff
@@ -34,7 +34,7 @@ tasks.closePanes = new UserTask({
   callback() {
     postActionToEditor({
       type: 'CHANGE_SIDE_PANE_MODE',
-      sidePaneMode: '_CLOSED',
+      mode: '_CLOSED',
```
bcolloran: hmm, tbh, i think it would be better to change the "CHANGE_SIDE_PANE_MODE" reducer in `src/eval-frame/reducers/eval-frame-reducer.js` and `/src/reducers/notebook-reducer.js` to look for `action.sidePaneMode` rather than changing this to pass "mode" since "mode" is not very descriptive and is already used elsewhere to mean the global mode

wlach: Makes sense, I was thinking that myself while working on this, but wasn't sure. I'll fix things up tomorrow.

wlach:
@bcolloran Fixed it up the way you suggested, can you take another look?

bcolloran:
looks great. fixed a merge conflict, we can merge as soon as CI completes

PR #834, opened 2018-09-04
Updated expected orderedmultidict hashes

wlach:
orderedmultidict now has a wheel distribution (gruns/orderedmultidict#17),
which means we need to update the hashes.


mdboom:
I see -- so they added wheels without adding a new version?  I guess that makes sense but sort of feels like an unintended consequence.

Anyway, this change looks fine.  Merging.

PR #837, opened 2018-09-04
Reintroduce "new notebook" menu item inside notebook

wlach:


wlach:
i.e. this: 
image

wlach:
@bcolloran I think the redux changes in this PR might be more up your wheelhouse to review.

src/eval-frame/iodide-api/__test__/environment.test.js
```diff
@@ -1,11 +1,11 @@
 import { environment } from '../environment'
 import { store } from '../../store'
-import { newNotebook } from '../../actions/actions'
+import { resetNotebook } from '../../../actions/actions'
```
bcolloran: should import from '../../actions/actions'

src/eval-frame/iodide-api/__test__/evalQueue.test.js
```diff
@@ -7,13 +7,14 @@ import {
   getRunningCellAsyncProcessStatus,
 } from '../evalQueue'
 import { store } from '../../store'
-import { temporarilySaveRunningCellID, newNotebook } from '../../actions/actions'
+import { temporarilySaveRunningCellID } from '../../actions/actions'
+import { resetNotebook } from '../../../actions/actions'
```
bcolloran: should import from '../../actions/actions'

src/eval-frame/iodide-api/__test__/environment.test.js
bcolloran: (that is `src/eval-frame/actions/actions.js`)

src/eval-frame/iodide-api/__test__/evalQueue.test.js
bcolloran: (that is `src/eval-frame/actions/actions.js`)

bcolloran:
@wlach i don't know if i'm doind something wrong, but i get unexpected behavior when i try the following:
1) in chrome, open localhost:8000 in a new incognito window (to make seure logins are cleared)
2) navigate to an existing notebook
3) click the "new notebook" menu item

i get a "save file" window with the filename pre-filled as "login", and "about:blank" appears in the url bar of the tab behind the "save file" window. Closing the "save file" window closes the "about:blank" tab, and no new notebook is available.

peek 2018-09-05 15-34

wlach:
@bcolloran hrm, it looks like the content-type is getting set incorrectly for the `/login/` endpoint to `<WSGIRequest: GET '/login'>`. It seems like Firefox renders it as html anyway, but Chrome doesn't. This problem doesn't have to do with this PR, so I think we should just go ahead and merge it if you don't see anything else wrong. I have another PR working to use a shorthand to render our templates, which seems to set the mimetypes correctly

screen shot 2018-09-06 at 5 13 35 pm

bcolloran:
ok @wlach sounds good, i'll merge.

Issue #792, opened 2018-08-24
it should be possible to view a list of any user's NBs at `<server url>/<user>`


bcolloran:
this list should be shown at this url for both logged in users and visitors, and should allow viewing of other users' notebook collections

from initial user stories at:
https://docs.google.com/document/d/1nq2mY3L57Wzfno5xj_eVcMECppB7iPA7jdyG5jRIUr4/edit?ts=5b7c6424

assigning @wlach initially, but just so that he can review, update, and assign to others as needed

wlach:
My implementation in #809 would be useful for reference as to how to make the data flow between components.

PR #838, opened 2018-09-04
implements /<username>

hamilton:
This implements #792.

screen shot 2018-09-04 at 3 31 00 pm

- ~there is a `users/` url pattern, which will default to the currently logged-in user~
- there is a `/<username>` pattern, which gives you a rudimentary user profile + list of notebooks + latest revision date
- ~both of these patterns use the same view - `server.views.user`~
- the index.jsx page now also links each notebook user name to their profile

@wlach probably best for you to review, since I primarily opted to build off of what you've already done. My Django skills are pretty rusty.


server/urls.py
```diff
@@ -8,6 +8,10 @@
 admin.autodiscover()
 
 urlpatterns = [
+    # user urls
+    url(r'^users/$', server.views.user, name='user'),
+    url(r'^users/(?P<name>[a-zA-Z0-9]+)', server.views.user, name='user'),
```
wlach: On second thought, how about we skip the /users/ part of this? Github. Less clutter this way.

server/urls.py
```diff
@@ -23,6 +27,5 @@
 
     # admin stuff
     path('admin/', admin.site.urls),
-
```
wlach: Revert this plz

server/views.py
```diff
@@ -33,6 +34,29 @@ def index(request):
...
 
+def user(request, name=None):
+    template = loader.get_template('user.html')
+    user_info = get_user_info_dict(request.user)
+    user = User.objects.get(username=name if name else user_info['name'])
+
+    this_user = {}
+    this_user['full_name'] = '{} {}'.format(user.first_name, user.last_name)
+    this_user['avatar'] = user.avatar
+    this_user['name'] = user.username
```
wlach: I would just write this as:

```py
this_user = {
   'full_name': '{} {}'.format(user.first_name, user.last_name),
   'avatar': user.avatar,
   ...
}
```

server/urls.py
```diff
@@ -8,6 +8,10 @@
 admin.autodiscover()
 
 urlpatterns = [
+    # user urls
+    url(r'^users/$', server.views.user, name='user'),
```
wlach: I would skip this route, see my suggestion below.

src/server/user.jsx
```diff
@@ -0,0 +1,40 @@
...
+        </h1>
+        <h2 style={{ fontWeight: 300, marginTop: 0 }}>{thisUser.name}</h2>
+      </div>
+      <h2>notebooks</h2>
+      <table >
+        <tbody>
+          {notebookList.map(notebook => (
+            <tr key={notebook.id}>
+              <td style={{ width: 200 }}><a href={`/notebooks/${notebook.id}/`}>{notebook.title}</a></td>
+              <td>{notebook.last_revision}</td>
```
wlach: Could you use something like [moment.js](https://momentjs.com/) to make the time column more readable? We can do this in a follow up ofc.

src/server/user.jsx
```diff
@@ -0,0 +1,40 @@
...
+            </tr>
+    ))}
+        </tbody>
+
```
wlach: Unnecessary newline?

server/views.py
```diff
@@ -33,6 +34,29 @@ def index(request):
...
+    return HttpResponse(template.render({
+        'user_info': json.dumps(user_info),
+        'this_user': json.dumps(this_user),
+        'notebook_list': json.dumps(
+            [{
+                'id': v[0],
+                'title': v[1],
+                'last_revision': NotebookRevision.objects
+                .filter(notebook_id=v[0]).last().created.isoformat(sep=' ')
+            } for v in Notebook.objects.filter(owner_id=user.id).values_list('id', 'title')
```
wlach: just remembered to check this. You can simplify the filter expression to `.filter(owner=user)...`

server/views.py
```diff
@@ -33,6 +34,29 @@ def index(request):
...
+    this_user['name'] = user.username
+    return HttpResponse(template.render({
+        'user_info': json.dumps(user_info),
+        'this_user': json.dumps(this_user),
+        'notebook_list': json.dumps(
+            [{
+                'id': v[0],
+                'title': v[1],
+                'last_revision': NotebookRevision.objects
+                .filter(notebook_id=v[0]).last().created.isoformat(sep=' ')
```
wlach: This is going to be pretty slow if we had a bunch of notebooks, we would probably do better to add a field in the notebook model itself to track this.

src/server/user.jsx
hamilton: Totally agreed, and also agreed that we can take this on when we're focusing on styling this page.

server/views.py
hamilton: Very good point, and also agreed we can handle this in a followup. I'll file an issue so we can keep track.

hamilton:
@wlach thanks for the review! I believe I've addressed all the issues you've brought up.

server/urls.py
```diff
@@ -8,6 +8,9 @@
 admin.autodiscover()
 
 urlpatterns = [
+    # user urls
+    url(r'^users/(?P<name>[a-zA-Z0-9]+)', server.views.user, name='user'),
```
wlach: Can you remove the `users/` part here per earlier discussion?

It occurs to me that you might need to move this down a bit so we match the other patterns first.

src/server/components/notebook-list.jsx
```diff
@@ -18,7 +18,7 @@ export default class NotebookList extends React.Component {
                   <a href={`/notebooks/${notebook.id}/`}>{notebook.title}</a>
                 </td>
                 <td>
-                  {notebook.owner}
+                  <a href={`/users/${notebook.owner}`}>{notebook.owner}</a>
```
wlach: Remove `users/` here.

server/urls.py
hamilton: Apologies, misunderstood the previous comment. Will do.

hamilton:
@wlach apologies for misunderstanding - fixed. Agreed that this is a much nicer url.

server/urls.py
```diff
@@ -24,6 +21,9 @@
     url(r'^logout/$', server.views.logout, name='logout'),
     url(r'^login/?', server.views.login, name='login'),
 
+    # user urls
+    url(r'^(?P<name>[a-zA-Z0-9]+)', server.views.user, name='user'),
```
wlach: I would probably move this to the very bottom, certainly below `admin/` (though in fact we might just want to delete that since we're not using it, another story)

hamilton: Sounds good - I'll move it to the very bottom. What happens when a user creates an account named `login` or `notebooks`?

wlach: At the moment this isn't possible, because we use github for auth and github doesn't allow such names. If we ever allow the creation of usernames we should probably restrict the allowed set the same way github does:

image

hamilton: Yeah, I'm assuming we'll allow the creation of user names down the line. I'm sure there is an easy django way to handle this :)

Issue #830, opened 2018-09-01
creating python cells in the web GUI


mash-graz:
as already reported in the pyodide issue tracker (https://github.com/iodide-project/pyodide/issues/133), it doesn't look possible to add new python cells or change existing ones to this type of cell.

this also effects the [official pyodide demo notebook](https://github.com/iodide-project/pyodide/issues/133) and the explanations given there...

![](http://users.mur.at/ms/attachments/pyodide1.png)

and:
![](http://users.mur.at/ms/attachments/pyodide2.png)

mash-graz:
found another open issue report concerning this flaw: #663

mdboom:
It is possible using the keybinding ("P"), but, yes, it should be visible on the menu as well, but isn't.

mash-graz:
thanks for this really important hint!

looks like nice *cheat code* to enter the next hidden level of practical usability/productivity. ;)

djbarnwal:
Working on a fix for this issue.

bcolloran:
see PR #832

mash-graz:
thanks for this really useful improvement!

PR #842, opened 2018-09-05
Add Change Language Option to Cell Menu

### LINKED ISSUE #839: Feature request: add plugin languages to cell menu
For example, on https://iodide.io/pyodide-demo/python.html, one has to press the "p" key to change a cell to Python.

It is confusing to users as most try to click on the cell menu and they don't see the language they are looking for.

djbarnwal:
Closes #839 #830

### DIFF
--- a/src/components/cells/__tests__/cell-menu.test.js
+++ b/src/components/cells/__tests__/cell-menu.test.js
@@ -21,6 +21,7 @@ describe('CellMenuUnconnected React component', () => {
       menuLabel: 'css',
       cellId: 5,
       skipInRunAll: false,
+      languages: [],
     }
     mountedMenu = undefined
   })
@@ -60,6 +61,19 @@ describe('cellMenu mapStateToProps', () => {
         id: 5,
         skipInRunAll: false,
       }],
+      languages: {
+        js: {
+          pluginType: 'language',
+          languageId: 'js',
+          displayName: 'Javascript',
+          codeMirrorMode: 'javascript',
+          codeMirrorModeLoaded: true,
+          module: 'window',
+          evaluator: 'eval',
+          keybinding: 'j',
+          url: '',
+        },
+      },
     }
   })
 
@@ -68,6 +82,7 @@ describe('cellMenu mapStateToProps', () => {
     expect(mapStateToProps(state, ownProps))
       .toEqual({
         skipInRunAll: false,
+        languages: [],
       })
   })
 })
--- a/src/components/cells/__tests__/cell-menu.test.js
+++ b/src/components/cells/__tests__/cell-menu.test.js
@@ -21,7 +21,7 @@ describe('CellMenuUnconnected React component', () => {
       menuLabel: 'css',
       cellId: 5,
       skipInRunAll: false,
-      languages: [],
+      userLanguages: [],
     }
     mountedMenu = undefined
   })
@@ -35,7 +35,7 @@ describe('CellMenuUnconnected React component', () => {
       .toBe('cell-menu-items-container')
   })
 
-  it('always renders seven NotebookMenuItem', () => {
+  it('always renders correct number of NotebookMenuItem without userLanguages', () => {
     expect(cellMenu().find(NotebookMenuItem).length).toBe(12)
   })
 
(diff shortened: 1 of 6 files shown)


hamilton:
Added @bcolloran and @mdboom on this since I believe they're more familiar & opinionated about this piece.

mdboom:
Thanks for this.  I see the new language show up on the cell drop down, but it doesn't appear on the main menu under "cell...".  Looks good to me as a user, but I'll let @bcolloran comment on the implementation.

src/reducers/notebook-reducer.js
```diff
@@ -182,6 +183,11 @@ const notebookReducer = (state = newNotebook(), action) => {
 
     case 'ADD_LANGUAGE_TO_EDITOR': {
       const { languageDefinition } = action
+      addChangeLanguageTask(
+        languageDefinition.languageId,
+        languageDefinition.displayName,
+        languageDefinition.keybinding,
+      )
```
bcolloran: this should happen in the addLanguageToEditor action creator rather than the reducer -- we want to move toward the reducer being a pure function of (state, action) with zero side effects. (when @hamilton and I were starting out with Redux, we didn't know that this was the best practice -- that it would make testing easier etc. But now that we're using thunk, we can move in that direction)

bcolloran: oops, that action creator is actually called `addLanguage` i think.

djbarnwal:
@bcolloran Updated the PR with suggested changes. @mdboom I think moving forward we are going to remove cells... submenu section (already removed from master) so I haven't added the option there.

src/components/cells/cell-menu.jsx
```diff
@@ -83,8 +95,13 @@ export class CellMenuUnconnected extends React.Component {
 
 export function mapStateToProps(state, ownProps) {
   const { cellId } = ownProps
+  let { languages } = state
+  languages = Object.keys(languages).filter(language => language !== 'js')
   const { skipInRunAll } = getCellById(state.cells, cellId)
-  return { skipInRunAll }
+  return {
+    skipInRunAll,
+    languages,
```
bcolloran: just noticed that we should add proptypes for `languages`. also, maybe change the var name to `userLanguages` (just a teeny bit more specific)

src/components/cells/cell-menu.jsx
```diff
@@ -15,8 +15,20 @@ export class CellMenuUnconnected extends React.Component {
   }
 
   render() {
+    let languagesTaskList
+    if (this.props.languages.length) {
```
bcolloran: do we need the conditional here? i can't remember how this works, but I think that if you `.map` and empty array, you always get an empty array back, and that if you pass that into the jsx below it will work just as well as the undefined `languageTaskList`. does that sound right?

src/components/cells/__tests__/cell-menu.test.js
```diff
@@ -68,6 +82,7 @@ describe('cellMenu mapStateToProps', () => {
     expect(mapStateToProps(state, ownProps))
       .toEqual({
         skipInRunAll: false,
+        languages: [],
```
bcolloran: can you add a test case for the case of having a user defined language?

bcolloran:
@djbarnwal when i load e.g http://localhost:8000/iodide.dev.html?url=https://iodide.io/pyodide-demo/data-club-demo-2018-05-29.html (in `npm start-and-serve` mode), when i click on a cell menu __before__ loading pyodide i get the following error:

```
cell-menu.jsx:23 Uncaught TypeError: Cannot read property 'title' of undefined
    at __WEBPACK_IMPORTED_MODULE_0_react___default.a.createElement.props.userLanguages.map.language (cell-menu.jsx:23)
    at Array.map (<anonymous>)
    at CellMenuUnconnected.render (cell-menu.jsx:21)
    at finishClassComponent (react-dom.development.js:8389)
    at updateClassComponent (react-dom.development.js:8357)
    at beginWork (react-dom.development.js:8982)
    at performUnitOfWork (react-dom.development.js:11814)
    at workLoop (react-dom.development.js:11843)
    at HTMLUnknownElement.callCallback (react-dom.development.js:100)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:138)
```

when i click on a cell menu __after__ loading pyodide, everything works as expected and looks great.

code also looks nice and clean too. I think once we figure out this bug we'll be good to merge. thanks @djbarnwal !

bcolloran:
looks good, thanks for that last fix @djbarnwal !!

PR #843, opened 2018-09-05
Always wait for db in Django server container

wlach:
This removes the need for some custom logic in our travis script and fixes
a race condition most commonly observed when bringing up the local dev
server environment up for the first time.


wlach:
@mdboom can you take a look at this one?

PR #844, opened 2018-09-05
Fix #789: Create new notebook button on homepage

### LINKED ISSUE #789: logged in users should be able to create new notebooks from their user landing page
- URL should become <server url>/<user>/new
- When I save the notebook, the url changes to <server url>/<user>/<notebook id>

from initial user stories at:
https://docs.google.com/document/d/1nq2mY3L57Wzfno5xj_eVcMECppB7iPA7jdyG5jRIUr4/edit?ts=5b7c6424

assigning @wlach initially, but just so that he can review, update, and assign to others as needed

mdboom:


server/notebooks/views.py
```diff
@@ -28,9 +28,6 @@ def notebook_view(request, pk):
...
 def new_notebook_view(request):
-    template = loader.get_template('notebook.html')
-    return HttpResponse(
-        template.render({
-            'user_info': _get_user_info_json(request.user),
-            'jsmd': ''
-        }, request))
+    notebook = Notebook.objects.create(owner=request.user)
+    NotebookRevision.objects.create(notebook=notebook)
+    return redirect('notebook-view', pk=notebook.id)
```
wlach: `/new/` is currently wired to create the notebook when the user decides to manually save:

https://github.com/iodide-project/iodide/blob/b4b0a06ef634cc949575d561c158ccabb7b5c9ba/src/actions/actions.js#L314

I think it would be best to leave this change out for now. Eventually we might want to do something more similar to what github does when creating a new repository (i.e. select a name + template in a form, then proceed to actually create the notebook): https://github.com/new

PR #850, opened 2018-09-06
fixes #819

### LINKED ISSUE #819: Plugin definitions stay in a %% meta tag for no apparent reason
I don't know why, but after you executed a plugin cell, it stays in the "languages" field of the JSON `%% meta` header.

It is not used if you reopen the script without the plugin cell, as the language it defines is ignored and the cell is converted to JS by default.

bcolloran:
should fix #819, but may need additional QA to make sure nothing else breaks in the language plugin world. (everything was fin in the testing I did)

### DIFF
--- a/src/components/cells/cell-editor.jsx
+++ b/src/components/cells/cell-editor.jsx
@@ -24,7 +24,6 @@ class CellEditor extends React.Component {
     cellId: PropTypes.number.isRequired,
     cellType: PropTypes.string,
     content: PropTypes.string,
-    codeMirrorModeLoaded: PropTypes.bool,
     actions: PropTypes.shape({
       selectCell: PropTypes.func.isRequired,
       changeMode: PropTypes.func.isRequired,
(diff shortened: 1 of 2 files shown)


bcolloran:
requesting review from @louisabraham

also, @mdboom  or @djbarnwal please approve

louisabraham:
Can you please upload the js (even as .txt on this thread)? I have trouble with `docker-compose`.

louisabraham:
I can confirm that when running `npm run start-and-serve`, opening http://localhost:8000/iodide.dev.html, adding a new language plugin and exporting, the resulting file does not contain a language header on the new branch and did on the master.

Great work @bcolloran !

Is it normal that iodide.dev.html doesn't look at all the same as https://iodide.io/iodide-examples/what-a-web-notebook-looks-like.html ? (the panel on the right)

bcolloran:
@louisabraham  -- regarding your question "Is it normal that iodide.dev.html doesn't look at all the same as https://iodide.io/iodide-examples/what-a-web-notebook-looks-like.html ? (the panel on the right)" -- what you see in our master branch represents the direction we are headed with the Iodide UI, which is heading in the direction of an R-Studio or MATLAB-IDE interface rather than the Mathematica or Jupyter style notebook interface. The version served at iodide.io is from our "stable" branch, and we serve the latest version of the master branch at "extremely-alpha.iodide.io".

We made this change in direction because the notebook metaphor is great for a lot of things, but it doesn't work terribly well for creating web documents, which is one of the main things we want to make easy for (data) scientists. We think the paned interface will be better, for this workflow, but it's not quite ready, so we haven't merged it into stable yet.

louisabraham:
Oh, I see. I am not an R-studio or MATLAB user. It would be great to allow the jupyter layout as an option. I think Jupyter is really the most popular today so it might satisfy the most users.
But maybe the interface will be better. For the moment it is really confusing not to see a command and its output on the same panel.

bcolloran:
@louisabraham yea, we surely have some UX work that still needs to be done to make the experience smoother (which is why we haven't rolled this all out to our "stable" branch yet), but I have no doubt that you'd quickly get the hang of this other UX style -- it has been used my millions of people successfully for many years, and is pretty straightforward, just a bit different. There are indeed pros and cons to both of these UX approaches, but we think that for main workflow we want to enable (creating web documents) the studio UX has more advantages than disadvantages. But I agree that we're not there yet! :-)

(Any feedback you have on ways to make the studio-style UX more helpful would be most appreciated! Thank you!)

PR #855, opened 2018-09-06
Use django render shortcut to render templates

wlach:
This is shorter than the more explicit code we were using before, and also
seems to set the mimetype correctly (which it wasn't before).

### DIFF
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -1,9 +1,8 @@
 import json
 
 from django.contrib.auth.decorators import login_required
-from django.http import HttpResponse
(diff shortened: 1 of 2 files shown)


wlach:
This fixes the issue @bcolloran noticed in https://github.com/iodide-project/iodide/pull/837#issuecomment-418903667

bcolloran:
@hamilton i merged #837 , so this should hopefully be ready for review/QA now (at least once the flake8 is ok)

wlach:
@hamilton flake8 issues should be fixed now

PR #860, opened 2018-09-06
view for listing all revisions of a notebook

hamilton:
screen shot 2018-09-11 at 3 38 41 pm

This PR implements a view for notebook revisions, as above as:

- the url is `<server-url>/notebooks/<NB-id>/revisions/`. 
- you can jump into a revision by going to `<server-url>/notebooks/<NB-id>/revisions/<revision-id>/`
- there is a link from the revisions page to the revision notebook outlined above.

A few things I think should probably change, which could be addressed in this PR or another:

- we need a better way of handling the css story here. Using `iodide.<tag>.css` is a non-starter since we set all sorts of funky things like `body { overflow: hidden }` there. It's already a problem because we cannot scroll on any of the server pages.
- I think we should use `emotion` to handle styling generally for the server, with the aim to eventually replace `material-ui` with our own components powered by composed styles. Would be happy to take a crack at this myself, but I think we should do _something_ sooner or later here.
- ~I'll have to adapt any changes made in other outstanding PRs (esp. @wlach's PR moving to the `render` shortcut) in this one if those land first (which I assume they will).~

Would love feedback on this.


src/server/revisions.jsx
```diff
@@ -0,0 +1,89 @@
...
+const blockLinkStyle = {
+  textDecoration: 'none',
+  color: 'black',
+}
+
+const revTable = {
+  paddingBottom: '40px',
+}
+
+const revR = {
```
bcolloran: what is revR?

src/server/revisions.jsx
```diff
@@ -0,0 +1,89 @@
...
+  paddingBottom: '5px',
+  textAlign: 'left',
+}
+
+const revC1 = {
```
bcolloran: what is revC1? revFirstColumn?

src/server/revisions.jsx
bcolloran: revRow?

wlach:
Hmm, good question. It's hard for me to know at the moment. :) I'd personally start with it in-notebook, but of course having this view in the server is better than not having anything at all. I have no objections to merging something like this if that's what you want to do, easy to rework or remove it later.

bcolloran:
cool, in that case i'd be in favor of merging this for now, and i'll add a user story for the in-notebook revision view.

hamilton:
I think we'll want both. They serve different purposes, I think. If I am a user interested in someone else's notebook, I don't necessarily want to look through revisions etc. in a modal.

But ... this is a place where React should shine :) I can easily make the list itself a component and include it in both, then see where things fall.

hamilton:
@wlach @bcolloran This is up for another review. Since we are now using `universal-router` I re-figured a few things.

@wlach I changed the location of the `revisions_view` to the top level instead of the notebook itself. It did not really feel that useful to have it in the notebooks dir. I can change it back if you disagree, but given that this is very much a server page it felt right to keep it out.

server/notebooks/views.py
```diff
@@ -23,6 +23,15 @@ def notebook_view(request, pk):
     })
 
 
+def revision_view(request, nb, pk):
+    print('------ made it here', nb, pk)
```
hamilton: gah need to delete this.

server/urls.py
```diff
@@ -3,13 +3,16 @@
 from django.urls import path
 
 import server.views
-
+from .notebooks.views import revision_view
```
wlach: Pretty sure this import isn't necessary

server/notebooks/views.py
```diff
@@ -23,6 +23,15 @@ def notebook_view(request, pk):
     })
 
 
+def revision_view(request, nb, pk):
```
wlach: I think it would be better to make revision a possible parameter for the notebook_view above (and use the latest if it isn't specified).

e.g. https://extremely-alpha.iodide.io/notebooks/1234?revision=123

You can get the revision parameter via the HTTP GET dictionary https://docs.djangoproject.com/en/2.1/ref/request-response/#django.http.HttpRequest.GET

server/views.py
```diff
@@ -56,6 +48,33 @@ def user(request, name=None):
     })
 
 
+def revisions(request, pk):
```
wlach: This should probably go under `notebooks/` and be called `notebook_revisions`

server/shared/get_user_info_dict.py
```diff
@@ -0,0 +1,13 @@
+from social_django.models import UserSocialAuth
```
wlach: A `shared/` directory isn't really idiomatic django. Can we just leave it in server.views? It seems more like a view-level helper function to me.

server/notebooks/views.py
hamilton: Good point!

hamilton:
@wlach addressed your comments. I am having trouble, however, getting revisions to show up. I've added a check for the `revision` param in `notebook_view`, which correctly finds the revision contents, but for some reason the iodide notebook is not parsing the jsmd even though it gets inserted correctly. Any ideas why this might be? Is there an insertion order issue here?

hamilton:
@wlach fixed the querystring issue. A notebook revision is now viewable, and if saved will create a new revision on the top of the stack (that's the default behavior). 

I think we can implement some sort of forking or whatever in another PR, or we can disable saving while viewing a revision for now.

server/views.py
```diff
@@ -4,16 +4,20 @@
...
 from .base.models import User
 
 
 def get_user_info_dict(user):
     if user.is_authenticated:
+        user_social_auth = UserSocialAuth.objects.get(user=user)
+        social_auth_extra_data = user_social_auth.extra_data
         return {
-            'name': user.username,
+            'name': social_auth_extra_data['login'],
```
wlach: Any reason for this change? I believe it will break if/when we ever allow non-social-auth logins.

src/handle-initial-jsmd.js
```diff
@@ -2,9 +2,12 @@ import { stateFromJsmd } from './tools/jsmd-tools'
 import handleUrlQuery from './tools/handle-url-query'
 import { updateAppMessages, importInitialJsmd, evaluateAllCells } from './actions/actions'
 
+const SERVER_QUERY_STRINGS = ['revision']
+
 export default async function handleInitialJsmd(store) {
   let state
-  if (window.location.search) {
+  if (window.location.search &&
+    SERVER_QUERY_STRINGS.every(s => !window.location.search.includes(s))) {
```
wlach: This feels kind of brittle to me. I wonder if we'd do better to only handle the url query for the case that we're in standalone (non-server) mode.

src/server/pages/revisions-page.jsx
```diff
@@ -0,0 +1,96 @@
...
+          <h3>Revisions</h3>
+          <table style={revTable}>
+            <tbody>
+              <tr style={revisionsRowStyle}>
+                <th>When</th>
+                <th>Title</th>
+              </tr>
+              {
+                        this.props.revisions.map((r, i) => {
+                            let theTitle
```
wlach: Instead of calculating the temporary, could you just do something like:
```jsx
<td>{ (i > 0 && this.props.revisions[i].title === this.props.revisions[i - 1].title) ? 
'-' : r.title }</td>
```

Don't really have a strong preference though, so feel free to leave it the way it is if you prefer it.

server/views.py
hamilton: True, not sure what happened here!

src/handle-initial-jsmd.js
hamilton: Very good point. I think I'll just check `IODIDE_BUILD_TYPE` as we do in other source files.

hamilton:
Thanks for all the reviews @wlach! Merging.

PR #862, opened 2018-09-07
Make sure user endpoint doesn't override oauth login/logout (fixes #856)

### LINKED ISSUE #856: login button when there is no authenticated github login causes an error
If I am using a browser profile (for instance, using a Private / Incognito Window) and I'm not logged onto github, and I go to the server and attempt to click the log in button, I get a django error.

wlach:


bcolloran:
i have never really learned regexes, and so i have fearful respect of them, which makes me wonder: do we have test in place for these url matchers? should we?

PR #863, opened 2018-09-07
Remove unused export-to-gist functionality

wlach:
This allows us to stop passing down the github access token to the
client, which is/was a bit of a security no-no. If we ever choose
to reimplement export-to-gist, we should do that on the server only.


bcolloran:
+1 to all that @wlach, i completely agree. haven't QAed this live b/c it looks like a very very straight forward code change, so if you're happy that it's all working how you want, then please merge whenever you're ready. thanks!

PR #864, opened 2018-09-07
Don't use notebook css when rendering server pages (fixes #859)

### LINKED ISSUE #859: server page css should _not_ use the notebook css
The notebook css is an odd duck case, and as such needs to not be used on any page generated on the server end other than the notebook.

Specifically, the `overflow:hidden` css property on `body` prevents any scrolling.

wlach:


bcolloran:
looks pretty straightforward to me assuming i'm understanding the django bits (this looks like a simple case). I think this looks fine to merge right now, but feel free to wait for @hamilton if you wanted his review.

wlach:
One slight downside is that this change makes the server pages look even more hacky than they already were. But they are still usable. On balance, let's go for it-- I don't think it should actually be that much work to clean things up later.

Issue #829, opened 2018-09-01
Deploying server-based assets to ghpages has bugs and limitations


wlach:
So we configured extremely-alpha.iodide.app (the iframe resources) to be deployed from github pages based on master. This has worked mostly ok so far, but I just ran into a strange issue where fetching the resources returns a 404:

![image](https://user-images.githubusercontent.com/20569/44947738-cb9a4f00-addf-11e8-8627-5728f1ca89a8.png)

After doing some investigation, I determined that the problem is that we just tagged v0.6, and our script thus determined that we should only produce the tagged versions of the files:

* https://github.com/iodide-project/iodide/blob/2d8b54eaebcce46d2c5b516f2eb5ace823b948d9/webpack.config.js#L55
* https://github.com/iodide-project/master-eval

The easiest workaround would be to simply unconditonally produce the dev versions of the files in ghpages, but that's a bit ugly. It also doesn't completely solve the problem where you would want to test both client and server changes simultaneously in a PR. Is there any reason we can't just make extremely-alpha.iodide.app point to the heroku server and skip this github pages step altogether? Heroku seems happy to serve the same application from multiple domains from what I can tell. It should also be easy to configure Django so that requests to the untrusted eval frame (iodide.app) don't have access to login or anything but the static content.

@bcolloran @mdboom what do you think?

mdboom:
@wlach: I think serving everything from Heroku makes sense, and is a little closer to dogfooding what we'll be doing for private installs in the future, too.  But I have to admit I don't fully understand all the details around Heroku -- I defer to you there.

The ghpages could probably be made to work, too, though -- ugly, yes, but that's sort of a part of the discussion we've had elsewhere about cleaning up the build stuff that seems to have accrued a lot of ugliness by now.

I've fixed it for now by adding a trivial commit to master, but that will break again the next time we make a tag so is not a real solution.

PR #871, opened 2018-09-10
fixes #743 -- make client and server HTML templates match (include iframe in server template)

### LINKED ISSUE #743: update server html templates to match iframe template (needed to get eval frame to load quickly)
bring the python htmle template up-to-date with the iframe-2 template at src/html-template.js (which now includes the iframe directly in the html to improve page load times).

bcolloran:
@mdboom @wlach @hamilton 
this is not at all working, just shows a start what i was thinking re #743 in https://github.com/iodide-project/iodide/issues/743#issuecomment-420099959, but i'm not sure how to get the env vars into server/notebooks/templates/notebook.html. Maybe you guys could help me with that bit? Feel free to either pick up this branch, or bin it and start fresh...

BTW: i'm totally open to cleaning out and simplifying the way the webpack bits work with this stuff. seems like we can do a much cleaner deployment if we just use .env rather than looking at the git branch? what do you all think?


mdboom:
So is the idea that the deployment configs in `.travis.yml` will write out `.env` files for each config?  Maybe we should have a set of `.env` files in the git repo and we copy/point at the right one depending on context.  It's nicer because it makes it easier to customize for a deployment that's not one of "ours", I suppose.  If that's the plan, maybe we should do even more variables done that way (such as debug vs. production).

bcolloran:
yea i think that could be in scope if it makes sense. again, i'm not sure this is actually the right thing to do, especially since i don't really know how the travis deploy works. I guess i had probably implicitly assumed that you could leave .env type files in place for your build steps.

in any case, the main point will be to make sure that the same set of origin strings are fed into the JS and the Python html templates. if having the travis.yml write out configs is not the right thing we can do something else.

server/notebooks/templates/notebook.html
```diff
@@ -6,7 +6,15 @@
 <script id="jsmd" type="text/jsmd">
 {{ jsmd|safe }}
 </script>
-<div id="page"></div>
+<div id="panes-container">
+  <iframe
+    id="eval-frame"
+    src="https:/extremely-alpha.iodide.app/iodide.eval-frame.dev.html"
```
wlach: You could turn this into a template parameter, as we do with `{{ jsmd|safe }}` above. 

https://docs.djangoproject.com/en/2.1/topics/templates/#the-django-template-language

mdboom:
But we will need to work out a solution for Travis' deployment of static (non-server) resources, right?  This currently breaks those.

wlach:
Oh right, I guess my vote would be to fix #829 alongside/before this. I don't think we should depend on ghpages for the server-side component.

mdboom:
Agreed -- I'm only concerned about ghpages for the non-server mode.  I think if we're using the server anyway, having it handle everything seems like the way to go.

wlach:
Ah, I see. I think for a static-only notebook deployed to ghpages, we can just inject some environment variables into the travis script (or circle if we switch to that). No need to muck around with .env files

mdboom:
I've updated this to move the github pages configuration to the .travis.yml file.

server/notebooks/templates/notebook.html
```diff
@@ -6,7 +6,15 @@
 <script id="jsmd" type="text/jsmd">
 {{ jsmd|safe }}
 </script>
-<div id="page"></div>
+<div id="panes-container">
+  <iframe
+    id="eval-frame"
+    src="{{ iframe_src|safe }}"
```
bcolloran: @wlach is this ok?

.travis.yml
```diff
@@ -35,7 +47,7 @@ matrix:
           keep-history: false
           repo: iodide-project/stable
           verbose: false
-          local-dir: prod
+          local-dir: dev
```
bcolloran: should this now change to `build`?

.travis.yml
```diff
@@ -47,7 +59,7 @@ matrix:
           repo: iodide-project/master-eval
           fqdn: extremely-alpha.iodide.app
           verbose: false
-          local-dir: prod
+          local-dir: dev
```
bcolloran: should this now change to `build`?

.travis.yml
mdboom: Yes.

.travis.yml
mdboom: Yes, as well.

bcolloran:
ok, changed dev->build in the travis files. any other notes @wlach and @mdboom ?

mdboom:
LGTM.  This is kind of thing that we probably won't know what in broken until we merge.  To that end, I probably don't have time to debug any Travis breakage today.  Maybe I'll merge (and then handle any fallout) on Monday?

bcolloran:
Monday works for me, thanks Mike. 

@wlach does the django stuff look ok?

server/notebooks/templates/notebook.html
wlach: the template here looks fine to me

server/notebooks/views.py
```diff
@@ -8,6 +9,8 @@
 from ..base.models import User
 from ..views import get_user_info_dict
 
+env = environ.Env()
```
wlach: more common practice would be to define some variables in the settings file (`server/settings.py`) which read from the environment and then imported them here.

Then do something like:

```py
from ..settings import (APP_VERSION_STRING,
                        EVAL_FRAME_ORIGIN)
```

bcolloran: ah that makes total sense, thanks!

bcolloran:
@mdboom -- ok, i addressed @wlach's comments, and this is working fine on my machine for the localhost case. i think i't ready for you to try merging on Monday. Thanks!

bcolloran:
(note that this appears to be failing CI because the newly expected env vars don't yet exist in travis's config)

server/settings.py
```diff
@@ -33,6 +33,8 @@
 SITE_URL = env("SERVER_URI", default="http://localhost:8000/")
 SITE_HOSTNAME = furl(SITE_URL).host
 ALLOWED_HOSTS = [SITE_HOSTNAME]
+APP_VERSION_STRING = env.str('APP_VERSION_STRING')
+EVAL_FRAME_ORIGIN = env.str('EVAL_FRAME_ORIGIN')
```
wlach: Let's set some defaults of these, this will allow the `collectstatic` command run in the Dockerfile to work. I would recommend `dev` for APP_VERSION_STRING and maybe `None` for EVAL_FRAME_ORIGIN?

Issue #851, opened 2018-09-06
as a user, when viewing another user's notebook i want the #886 interstitial to warn me that i cannot save, and give me an option to "MAKE A COPY"


bcolloran:
snackbar should pop that shows:
`This notebook belongs to another user                   MAKE A COPY`

clicking the `MAKE A COPY` link:
1) creates a copy of the current state of the notebook belonging to the logged in user
2) saves the notebook
3) change the URL without reloading the page (we think this should be possible, but haven't investgated fully; see https://stackoverflow.com/questions/3338642/updating-address-bar-with-new-url-without-hash-or-reloading-the-page )

should also add a menu item for "Make a copy"

these can both use the same user task, but for now we think that this user task should not have a keybinding.

bcolloran:
updated title to reflect the fact that we want this flow to go through the interstitial bnner thingy in #886

PR #877, opened 2018-09-11
Fix #791, #851: Save a copy of notebook to user's account

### LINKED ISSUE #791: when viewing _another_ user's NB, a logged in user can save a revision to their user accout via menu/button/ctrl+s
from initial user stories at:
https://docs.google.com/document/d/1nq2mY3L57Wzfno5xj_eVcMECppB7iPA7jdyG5jRIUr4/edit?ts=5b7c6424

assigning @wlach initially, but just so that he can review, update, and assign to others as needed

mdboom:
Strictly speaking, this closes the issues, but I'm not proud of using `window.confirm` :)

Looked into ways of presenting modal dialogs with material-ui and they all seem to either require persistent DOM content, are incomplete examples or only work with material-ui 0.x.  Maybe you all have a better idea?

### DIFF
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -289,18 +289,7 @@ export function saveNotebookToServer() {
       },
     }
 
-    if (notebookInServer) {
-      // Update Exisiting Notebook
-      fetch(`/api/v1/notebooks/${state.notebookId}/revisions/`, postRequestOptions)
-        .then(response => response.json())
-        .then(() => {
-          const message = 'Updated Notebook'
-          dispatch(updateAppMessages({
-            message,
-            details: `${message} <br />Notebook saved`,
-          }))
-        })
-    } else {
+    const createNewNotebook = () => {
       // Create a New Notebook in Database
       fetch('/api/v1/notebooks/', postRequestOptions)
         .then(response => response.json())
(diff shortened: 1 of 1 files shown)


bcolloran:
Rather than putting this logic inside of saveNotebookToServer, I wonder if it might make sense to invert control here a little bit? -- maybe replace saveNotebookToServer with a wrapper trySavingNotebookToServer that can detect (a) if a user is logged in, and (b) if a user is looking at their own notebook, inwhich case saveNotebookToServer can be called straight away; but if those things are not true, the user can be prompted to log in and/or make a copy).

I don't have particularly strong feelings that this is the "right" way to do this, just tossing it out as an idea -- seems like we might be able to keep saveNotebookToServer tightly focused on just literally the saving bit, and use a bit of wrapping to handle the extra business logic. @wlach do you have any thoughts?

see also  #878

mdboom:
I'm  not sure if it will always be the case that a user could only write to their own notebooks, and to the extent that we have more complex "sharing" logic, that naturally for security reasons needs to live on the server, not be duplicated on the client.  So I think it should be "ask the server if I can save, if not, do something I know always works (save to my own account)".  We probably need more fine-grained checking and error reporting of what the server says, though.

wlach:
Yeah, `saveNotebookToServer` is getting kind of ridiculously long. I think factoring out a `saveNotebook` function which does different things depending on (e.g.) whether the user owns the notebook or not is the way to go. Maybe it would be best actually if we put the actual logic to interface with the server in some other file, so we can clearly disambiguate this type of business logic from actual interactions with the server.

That said, I have no objections to landing this in its current state. It doesn't make things substantially worse and it's easy to refactor later.

mdboom:
Closing in favor of #903.

PR #880, opened 2018-09-12
Fix the user regex to be more inclusive

mdboom:
github usernames can have dashes and underscores and things.

I just copied the regex from another project that reportedly supports github usernames here.


wlach:
We really should add tests for this sort of thing, but I think this can go in for now.

PR #881, opened 2018-09-12
implement emotion.js, a css-in-js library, for server styling

### LINKED ISSUE #872: need to fix/update/consolidate styles and css reset for server side pages
@hamilton and @wlach have discussed some ideas around this

hamilton:
Fixes #872.

screen shot 2018-09-12 at 10 02 59 am

I'm not sure if this is the right approach, but I got in a flow last night and ended up with this solution for styling our server pages and figured I'd open a PR for feedback.

This PR does a number of things:

- applies a global css reset (`normalize.css`)
- refactors a few of the server components and standardizes them for general usage.
- utilizes `emotion` and `react-emotion` for CSS-in-JS styling for all components. This allows our css rules for components to be 100% localized to the component in question. There are many css-in-js libraries, and they all effectively solve the problem similarly, but I like `emotion` because of its composability, very small footprint, and performance.
- implements a footer (for now, as a placeholder).
- creates a table styling with `react-emotion` (found in `/src/server/components/table.jsx`) which can be used on any table element. I did this as an example of how to style a presentational container efficiently and reuse it. This component is used on the index and the user page.
- creates a `PageBody` component that allows for a sticky footer and standardized spacing for the content & footer.
- prepares us for moving away from `material-ui` for certain components, such as the toolbar, buttons, etc.

Background reading for this PR:

- [The End of Global CSS](https://medium.com/seek-blog/the-end-of-global-css-90d2a4a06284)
- [All You Need to Know About CSS-in-JS](https://hackernoon.com/all-you-need-to-know-about-css-in-js-984a72d48ebc)


src/server/components/footer.jsx
```diff
@@ -0,0 +1,62 @@
...
+span {
+  font-size: .9em;
+  display: inline-block;
+  position: relative;
+  top:-9px;
+  left: -3px;
+}
+`
+
+const Logo = () => (
```
hamilton: this will probably get replaced with an actual svg logo that scales down well. @bcolloran's logo doesn't shrink too well (which makes sense because it was designed for stickers :) )

src/server/index.jsx
```diff
@@ -6,6 +6,8 @@ import HomePage from './pages/home-page'
 import LoginPage from './pages/login-page'
 import UserPage from './pages/user-page'
 
+import BaseStyle from './style/base' // eslint-disable-line
```
hamilton: There is likely a better way of doing this with emotion's `injectGlobal` that doesn't require an `eslint-disable-line`

bcolloran: just curious, does `import './style/base'` work -- along the lines of PR #882, sometimes you can just run the code you need by invoking the import without actually pulling in any variable

Issue #884, opened 2018-09-12
all notebook actions that open a page should do so in a new tab


bcolloran:
because we don't want people to lose evaluation state in an open notebook

this applies to e.g.: 
- new notebook, see #381 
- go to user profile
- any links in help modals (or other modals)
- any links in general

bcolloran:
we can try this:
https://www.w3schools.com/TAGS/att_base_target.asp

(see also #885 for the iframe. we could know both of these out in one PR)

PR #891, opened 2018-09-13
'new notebook' button now targets _blank, user menu callbacks open in new tab

hamilton:
This finishes the last piece of #884 I believe.


bcolloran:
This looks fine, r+ from me -- but #884 actually refers to the client side, this just addresses the server side button i think?

hamilton:
@bcolloran forgot to push the server side user button

PR #892, opened 2018-09-13
Add basic page loading tests

wlach:
These just test that the index and user pages have the expected data needed
by the react fixtures for now. To make this possible, there was some light
refactoring of the unit tests (see other commits in this PR)


server/tests/test_server_pages.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_index_view(client, two_test_notebooks):
+    resp = client.get(reverse('index'))
+    assert resp.status_code == 200
+    assert set(_get_page_data(str(resp.content)).keys()) == set(['userInfo',
+                                                                 'notebookList'])
+
+
+def test_user_view(client, fake_user, two_test_notebooks):
```
hamilton: Seems like a reasonable format to me, I see how we can use this to test when a username doesn't exist.

wlach: Yep, here's an example of such a test in the current repo (note that the directory will change after this lands): 

https://github.com/iodide-project/iodide/blob/5ad54f81e97d4dbadaf645a47eb92615e30fb273/server/notebooks/tests/test_notebook_revision_api.py#L53

PR #903, opened 2018-09-14
Fix #851: Add interstitial about saving to own account

### LINKED ISSUE #851: as a user, when viewing another user's notebook i want the #886 interstitial to warn me that i cannot save, and give me an option to "MAKE A COPY"

mdboom:
**Requires #908 to be merged first**

Sorry this is sort of combining two things:

1. This adds a bar right below the toolbar to display notifications that require some sort of action. The state uses a list of these notifications, so there could be more than one.
2. Displaying a message that the user can't save because they aren't logged in, or that they need to save a copy to their own account if this notebook isn't theirs.  (This is really just a proof-of-concept of #1).

Still getting my sealegs with react.  I have some inline comments below.

image

### DIFF
--- a/server/notebooks/templates/notebook.html
+++ b/server/notebooks/templates/notebook.html
@@ -8,5 +8,6 @@
 </script>
 <div id="page"></div>
 {{ user_info|json_script:"userData" }}
+{{ notebook_info|json_script:"notebookInfo" }}
 <script src="/iodide.dev.js"></script>
 {% endblock %}
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -255,43 +255,70 @@ export function logout() {
   }
 }
 
+function getNotebookSaveRequestOptions(state) {
+  const data = {
+    title: state.title,
+    content: exportJsmdToString(state),
+  }
+
+  // Get CSRF Cookie for Django CSRF Middleware
+  function getCookie(name) {
+    if (!document.cookie) {
+      return null
+    }
+    const token = document.cookie.split(';')
+      .map(c => c.trim())
+      .filter(c => c.startsWith(`${name}=`))
+
+    if (token.length === 0) {
+      return null;
+    }
+    return decodeURIComponent(token[0].split('=')[1])
+  }
+
+  const csrftoken = getCookie('csrftoken')
+
+  const postRequestOptions = {
+    body: JSON.stringify(data),
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'X-CSRFToken': csrftoken,
+    },
+  }
+
+  return postRequestOptions
+}
+
+export function createNewNotebookOnServer() {
+  return (dispatch, getState) => {
+    const state = getState()
+
+    const postRequestOptions = getNotebookSaveRequestOptions(state)
+    // Create a New Notebook in Database
+    fetch('/api/v1/notebooks/', postRequestOptions)
+      .then(response => response.json())
+      .then((json) => {
+        const message = 'Notebook saved to server'
+        dispatch(updateAppMessages({
+          message,
+          details: `${message} <br />Notebook saved`,
+        }))
+        dispatch({ type: 'ADD_NOTEBOOK_ID', id: json.id })
+        window.history.pushState('', {}, `/notebooks/${json.id}`)
+      })
+    dispatch({ type: 'NOTEBOOK_SAVED' })
+  }
+}
+
 export function saveNotebookToServer() {
   return (dispatch, getState) => {
     const state = getState()
 
     const notebookInServer = Boolean(state.notebookId)
-    const data = {
-      title: state.title,
-      content: exportJsmdToString(state),
-    }
-
-    // Get CSRF Cookie for Django CSRF Middleware
-    function getCookie(name) {
-      if (!document.cookie) {
-        return null
-      }
-      const token = document.cookie.split(';')
-        .map(c => c.trim())
-        .filter(c => c.startsWith(`${name}=`))
-
-      if (token.length === 0) {
-        return null;
-      }
-      return decodeURIComponent(token[0].split('=')[1])
-    }
-
-    const csrftoken = getCookie('csrftoken')
-
-    const postRequestOptions = {
-      body: JSON.stringify(data),
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json',
-        'X-CSRFToken': csrftoken,
-      },
-    }
 
     if (notebookInServer) {
+      const postRequestOptions = getNotebookSaveRequestOptions(state)
       // Update Exisiting Notebook
       fetch(`/api/v1/notebooks/${state.notebookId}/revisions/`, postRequestOptions)
         .then(response => response.json())
@@ -302,21 +329,10 @@ export function saveNotebookToServer() {
             details: `${message} <br />Notebook saved`,
           }))
         })
+      dispatch({ type: 'NOTEBOOK_SAVED' })
     } else {
-      // Create a New Notebook in Database
-      fetch('/api/v1/notebooks/', postRequestOptions)
-        .then(response => response.json())
-        .then((json) => {
-          const message = 'Notebook saved to server'
-          dispatch(updateAppMessages({
-            message,
-            details: `${message} <br />Notebook saved`,
-          }))
-          dispatch({ type: 'ADD_NOTEBOOK_ID', id: json.id })
-          window.history.pushState('', {}, `/notebooks/${json.id}`)
-        })
+      createNewNotebookOnServer()(dispatch, getState)
     }
-    dispatch({ type: 'NOTEBOOK_SAVED' })
   }
 }
 
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -306,8 +306,8 @@ export function createNewNotebookOnServer() {
         }))
         dispatch({ type: 'ADD_NOTEBOOK_ID', id: json.id })
         window.history.pushState('', {}, `/notebooks/${json.id}`)
+        dispatch({ type: 'NOTEBOOK_SAVED' })
       })
-    dispatch({ type: 'NOTEBOOK_SAVED' })
   }
 }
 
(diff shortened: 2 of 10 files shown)


src/actions/actions.js
```diff
@@ -302,21 +346,10 @@ export function saveNotebookToServer() {
...
-            details: `${message} <br />Notebook saved`,
-          }))
-          dispatch({ type: 'ADD_NOTEBOOK_ID', id: json.id })
-          window.history.pushState('', {}, `/notebooks/${json.id}`)
-        })
+      createNewNotebookOnServer()(dispatch, getState)
     }
-    dispatch({ type: 'NOTEBOOK_SAVED' })
   }
 }
```
mdboom: This diff is awful.  I basically broke it up into two functions: create and save.

src/components/menu/header-messages.jsx
```diff
@@ -0,0 +1,57 @@
...
+      message: PropTypes.string,
+      button: PropTypes.string,
+      callback: PropTypes.function,
+      id: PropTypes.number,
+    })),
+  }
+  render() {
+    const handler = (id, callback) => () => {
+      if (callback()) {
+        dispatch(removeHeaderMessage(id))
```
mdboom: The callback can return false to prevent the notification from going away.

src/handle-initial-jsmd.js
```diff
@@ -27,4 +27,27 @@ export default async function handleInitialJsmd(store) {
       store.dispatch(evaluateAllCells())
     }
   }
+
+  state = store.getState();
```
mdboom: This doesn't really belong here, but I couldn't find another place to do sanity checks right after loading a notebook.  Any suggestions?

src/components/menu/header-messages.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+    const closeHandler = id => () => dispatch(removeHeaderMessage(id))
+
+    const items = this.props.messages.map(({
+      message,
+      button,
+      id,
+      callback,
+    }) => (
+      <div>
```
mdboom: I'm getting

```
Warning: Each child in an array or iterator should have a unique "key" prop.
```
with this.  I'll have to sort that out.

wlach: this one is easy, try:
```jsx
<div key={`item-${id}`}>`
```

src/handle-initial-jsmd.js
```diff
@@ -27,4 +27,27 @@ export default async function handleInitialJsmd(store) {
...
+  if (state.viewMode === 'EXPLORE_VIEW' && state.userData.name !== state.ownerData.username) {
+    if (state.userData.name === undefined) {
+      store.dispatch(addHeaderMessage(
+        'To save to the server, you need to ',
+        'login.',
+        () => {
+          store.dispatch(login())
+          return true
+        },
```
wlach: This doesn't seem right to me -- it should be up to the react component to display these messages. In this case I think we can just detect that the user is not logged in and display this kind of message unconditionally. Same idea for the one below -- display the banner if the currently loaded notebook is not owned by the current user.

src/handle-initial-jsmd.js
wlach: (although now that I look at it, I don't think you should be using redux for this particular use case)

src/handle-initial-jsmd.js
bcolloran: +1

mdboom:
Thanks for the tips about where to put the logic.  I think I have that sorted.

However, I'm not sure we don't want a close button.  Won't it be sort of annoying to use so much real estate just for not being logged in, for example?  It's a completely normal use case to just view someone else's notebook.  Or maybe we only show when the user starts to edit -- to indicate they won't be able to save.  In any case, that's probably finesse that can be put off for later.

server/notebooks/templates/notebook.html
```diff
@@ -8,7 +9,8 @@
 </script>
 <div id="page"></div>
 <script>
-  window.userData = {{ user_info|safe }}
+  window.userData = {{ user_info|json }};
+  window.notebookInfo = {{ notebook_info|json }};
```
mdboom: @wlach: I'm made this change because I discovered that the old approach doesn't always produce valid JSON.  Particularly for booleans -- you get `True` (from Python) rather than `true` for JS/JSON.  `django-argonauts` seems built explicitly to solve this use case, but we could also do this ourselves if adding the dependency isn't preferred.

mdboom:
I've pushed a different approach to this based on feedback.  I think this is closer to what you all had suggested.

server/notebooks/templates/notebook.html
mdboom: On further inspection, I saw that some places were calling `json.dumps` to output these variables, and others weren't.  I still think this might be the better approach, but less critically -- if you'd rather I just make everything use `json.dumps` consistently instead, that's probably fine, though according to the argonaut docs, it is somewhat safer against script injection attacks -- if in fact such things aren't getting caught elsewhere in the stack.

server/tests/test_server_pages.py
```diff
@@ -6,7 +6,7 @@
 
 
 def _get_page_data(page_content_str):
-    m = re.search('window.pageData = (\{.*\});', str(page_content_str))
+    m = re.search(r'window.pageData = (\{.*\});', str(page_content_str))
```
mdboom: Fixes `<unknown>:11: DeprecationWarning: invalid escape sequence \{`

server/notebooks/templates/notebook.html
wlach: `django-argonauts` looks like an old and obsolete package to me. It seems that the latest version of django has a template builtin for our exact use case:

https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#json-script

Maybe we can file an issue to convert our scripts to use that?

mdboom:
> It looks like we need to have an action and a reducer case (either new or modified existing action and reducer case) to make sure that notebookInfo.user_can_save is updated in the reducer store, and hence that this change flows forward into the interface (maybe i missed it, but i didn't see anything that looked like it was doing that part)

That's true -- there's nothing to do that part.  But I think we need to query the server to see if we can save the notebook, since that's where that logic needs to live.  We can't just change that on the client, right?  A `reload` would do it, but we could also add another API endpoint to get the `notebookInfo` for a given notebook.

bcolloran:
> > It looks like we need to have an action and a reducer case (either new or modified existing action and reducer case) to make sure that notebookInfo.user_can_save is updated in the reducer store, and hence that this change flows forward into the interface (maybe i missed it, but i didn't see anything that looked like it was doing that part)
> 
> That's true -- there's nothing to do that part. But I think we need to query the server to see if we can save the notebook, since that's where that logic needs to live. We can't just change that on the client, right? A `reload` would do it, but we could also add another API endpoint to get the `notebookInfo` for a given notebook.

Hmm yea i guess i assumed you'd be able to do it within one of those existing actions depending on the response you get from the server in the fetch (like, i assumed the server would respond with either "yep that make a copy request worked" or  "nope, failed to make a copy"), and that depending on that response you'd be able to dispatch an action like `UPDATE_NOTEBOOK_INFO` or something like that. But i have no idea how that part works. @wlach do we need to change how the server responds to requests to enable this, or is there some other way to handle this correctly?

re button vs link:
image
looks like the button actually ends up messing with the padding. path of least resistance is to stick with the `<a>` i think, and i now see that our css reset is too aggressive -- i'll fix that styling issue in a separate PR.

mdboom:
Actually, it looks not too hard to add the notebook info to the POST "/api/v1/notebooks/" endpoint.  That seems like the right way to do it.  I'll add that here.

wlach:
I'm not 100% sure if I understand the issue under discussion, but I think the right way of handling this case (until we have forking) is to create a new notebook on the server, with exactly the same content + metadata as the existing one. The only difference will be, of course, the fact that it is owned by the current user (and the loss of revision history).

mdboom:
@wlach: That's absolutely the plan (and what this currently does).  The only issue is how to update the client to show that now we have something editable after creating the new notebook (without simply reloading the page, which would be the obvious yet clunky way to do it).

wlach:
@mdboom can you not just send a redux message to change the notebook Id and user info metadata?

mdboom:
I thought the determination of "do I have the write bit on this notebook" should be determined on the server.  But maybe I'm overthinking it -- it's probably fine to assume ownership === writeability for now until we grow something more complicated.

mdboom:
Ok.  I've added the state change to the `SAVE_NOTEBOOK` event.  I think it's safe to assume that if the notebook successfully saved, it will successfully save again.

Also added some tests.

src/components/menu/header-messages.jsx
```diff
@@ -0,0 +1,72 @@
...
+          <span>
+            To save to this server, you need to <a onClick={this.props.login}>login</a>.
+          </span>
+        )
+        break
+      case 'NEED_TO_MAKE_COPY':
+        content = (
+          <span>
+            This notebook is owned by another user.
+            <a onClick={this.props.makeCopy}>Make a copy to your account</a>.
```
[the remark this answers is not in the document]
bcolloran: @mdboom ah ok, the simple trick is to include an empty bracket pair:
`This notebook is owned by another user. {}`

(from https://stackoverflow.com/questions/40264084/best-practice-when-adding-whitespace-in-jsx tested locally)

bcolloran:
@mdboom the error seems to be something about my chromium and not about this PR. I pushed a lint fix and we can merge as soon as travis passes

PR #904, opened 2018-09-15
add "delete notebook" and "delete revision" functionality

### LINKED ISSUE #888: As a user, I want an easy way to delete a notebook and revisions I've created
Users will want a way to delete notebooks (include all their revisions), as well as specific revisions.

If I am logged in and on my user page, there should be an easy way to delete the notebook from that page. Either that, or I should be able to go to something like `/notebooks/<notebook-id>/revisions/` and delete the whole notebook + revisions there.

Any deletion requires the creation of an "are you sure?" modal, which will have to be an additional component created for general server work.

Assigning to @bcolloran for proper triaging and @wlach for additional ideas & method for going about accomplishing this, but I am happy to do the actual work for this. Feel free to un-assign yourselves accordingly!

hamilton:
~This is a WIP, so probably not quite ready for review.~ Fixes #888.

sep-14-2018 18-51-49

This PR does a handful of things:

- adds a method for `DELETE` for notebook revisions via `api_views.py`. This was trivial to add but worth a set of eyes to make sure it's right.
- creates a button set under `/src/server/components/buttons.jsx` - a `TextButton`, `OutlineButton`, and `ContainedButton`. I went ahead and pulled out the `material-ui` button we were using on the user page because I found it to be a bit heavy, kind of awkward, and not very responsive to clicks. Since we'll eventually want to command our own UX at some point, it feels like now is a good time to start our own component set. Furthermore, the material-ui buttons do something odd with `z-index` that makes them always live over the modal.
- created a `Modal` element using a much more lightweight set of primitives than `material-ui`, and whose clickaway event is a bit easier to reason with.
- with those previous two components in mind, I created a `DeleteElementButton` which gracefully handles a sensitive task that needs review (I could see this be renamed to something like that), which I am using for deleting notebooks and revisions. So you have to confirm via modal if you want to delete anything. Styling and interactions are totally up for debate but I'm sure we'll want to use something like this for lots of other things (projects, workspace, comments, etc.)
- applied the `DeleteElementButton` to the user page, where you can now delete a particular notebook, and to the revisions page, where you can either delete the whole notebook or an individual revision. This was the whole point of all the above work. I'm open suggestions on what direction for these basic tasks but here's a first pass at least.
- added an `AttentionBox` which gives a bit more of a user-friendly way to jump into creating more stuff if they have no notebooks. So if you are logged in and have no notebooks, it should prompt you to get creating. The GIF above demonstrates this. I think something like this will be generally useful for all sorts of other contexts.

A few things to add to this PR before its ready for review:

- [x] I want to get all the `PropTypes` for these new components in place, and really all the server components. It's ridiculous not to have any, so that should be a prerequisite before review. ~these can be added in a future PR~
- [x] add keypress for escaping modal that is only active when modal is.
- [x] abstract out the fetch calls that need the csrf token, and apply the resulting function in the places that need them (`actions.js` and `delete-element-button.jsx`).
- [x] fix the use-case when a user deletes all revisions. The notebook should probably be deleted _in toto_ at that point.
- [x] I want to make sure if there are any API tests related to the addition of `delete` to notebook revisions that I have them appropriately covered.


server/notebooks/api_views.py
```diff
@@ -62,6 +62,11 @@ def get_serializer_class(self):
             return NotebookRevisionSerializer
         return NotebookRevisionDetailSerializer
 
+    def perform_destroy(self, instance):
+        if instance.notebook.owner != self.request.user:
+            raise PermissionDenied
+        viewsets.ModelViewSet.perform_destroy(self, instance)
```
wlach: Is this method override actually necessary? I believe this should be taken care of by the permission classes, above:

https://github.com/iodide-project/iodide/pull/904/files#diff-6c8e78b546a064d4a9debd0afe22881dR45

server/notebooks/api_views.py
```diff
@@ -45,7 +45,7 @@ class NotebookRevisionViewSet(viewsets.ModelViewSet):
     permission_classes = (permissions.IsAuthenticatedOrReadOnly,)
 
     # revisions should be considered immutable once created
-    http_method_names = ['get', 'post', 'head']
+    http_method_names = ['get', 'post', 'head', 'delete']
```
wlach: Could you add some unit tests to make sure this works as expected for all the possibilities (logged-in-as-owner, logged-in-as-non-owner, not-logged-in)?

src/server/components/delete-element-button.jsx
```diff
@@ -0,0 +1,80 @@
...
+  showModal() {
+    this.setState({ visible: true })
+  }
+
+  deleteObject() {
+    const csrfToken = getCookie('csrftoken')
+    const headers = {
+      'Content-Type': 'application/json',
+      'X-CSRFToken': csrfToken,
+    };
```
wlach: This is the second time I think we're constructing these headers. I feel like we should abstract away this csrf stuff somehow, either with our own helper library/code or something from npm.

server/notebooks/api_views.py
hamilton: Yes, that was one of the remaining prerequisites for this PR I mentioned before. Will make sure there are tests when I circle back on this.

src/server/components/delete-element-button.jsx
hamilton: Yep, good point. I'll abstract it out as a util fcn that we can reuse elsewhere.

server/notebooks/api_views.py
wlach: Oh wait, on second look, no it isn't. Definitely need this logic. ;)

server/notebooks/api_views.py
hamilton: @wlach as of the most recent commit I added these tests in `test_notebook_revision_api.py` - feel free to take a look and let me know if anything is amiss there.

wlach:
Seeing some strange jumping (all content shifts to the right?) when I press the delete button and the modal comes in:

jumpinjpegs

src/server/pages/user-page.jsx
```diff
@@ -23,7 +26,81 @@ h2 {
...
+              <th />
+            </tr>
+          </thead>
+          <tbody>
+            {this.state.notebookList.map(notebook => (
+              <tr key={notebook.id}>
+                <td><a href={`/notebooks/${notebook.id}/`}>{notebook.title}</a></td>
+                <td>{notebook.last_revision.slice(0, 19)}</td>
+                <td><a href={`/notebooks/${notebook.id}/revisions/`}>revisions</a></td>
+                <td>
```
wlach: Showing as many delete buttons as there are notebooks seems a little heavy. I wonder if we'd do better to have a menu here hidden under a '...' button with options to show the revisions and delete?

server/notebooks/api_views.py
wlach: +1 they look great

src/server/components/delete-element-button.jsx
```diff
@@ -0,0 +1,55 @@
...
+    this.setState({ visible: true })
+  }
+
+  deleteObject() {
+    fetchWithCSRFToken(this.props.url, { method: 'DELETE' })
+      .then(() => {
+        this.hideModal()
+        this.props.onDelete(this.props.elementID);
+      }).catch((err) => {
+        console.error(err)
```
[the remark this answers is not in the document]
hamilton: good idea

src/server/pages/user-page.jsx
hamilton: Yeah, that was my initial instinct as well. Might try another pass at an activity menu that includes a delete button after this PR lands.

src/server/components/delete-element-button.jsx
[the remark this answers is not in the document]
hamilton: actually, I think this should wait for another PR to see how we'd want to integrate a snackbar onto the server pages, so I'll punt on this for now.

package-lock.json
```diff
@@ -148,7 +148,7 @@
       "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.0.0.tgz",
```
wlach: On another pass we need to regenerate this file with just the changes related to the ones in `package.json`.

PR #914, opened 2018-09-18
Fix potential race when running tests on travis

wlach:
It seems like if docker-compose will attempt to bring up multiple
instances of a container if you run it once with `-d` and then
attempt to run a one-off command before it has brought up the full
environment. Get around this by waiting for our dependent container
to completely start before running our tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,8 +100,12 @@ matrix:
       script:
         - cp .env-dist .env
         - make build
-        # bring up docker containers and wait for db setup to complete
+        # bring up docker containers and wait for db setup to complete (otherwise
+        # our tests might try to bring up another instance of the db, which is a
+        # race condition)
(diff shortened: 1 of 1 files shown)


wlach:
I noticed that the travis tests have been intermittently failing since I moved the logic to wait for postgres into the docker run command in #843. This is partly a speculative fix based on what I'm seeing, but am pretty sure my logic is sound. I'll retrigger travis a few times to be sure.

mdboom:
For the record, I've seen this too on Travis.

bcolloran:
@wlach this was green when i opened this link, just restarted the CI at 12:26 PT. i think if it passes again we should merge?

wlach:
Yup, has worked well after multiple retriggers. Let's merge this.

PR #922, opened 2018-09-18
fixes the interstitial appearing with the /new url

### LINKED ISSUE #917: "create new notebook" button takes me to the /new URL, but I also get an interstitial saying "this notebook is owned by another user"
screen shot 2018-09-18 at 12 27 08 pm

hamilton:
Fixes #917. 

- created a new function in `handle-server-variables.js` which operates similarly to `handleInitialJsmd` - it helps initiate the store with non-store data, particularly using the `getNotebookInfo`
- updated the `new_notebook_view` to pass by default to `user_can_save: true` (since atm login is required to create a new notebook, we'll want to change this later)

### DIFF
--- a/src/reducers/notebook-reducer.js
+++ b/src/reducers/notebook-reducer.js
@@ -55,6 +55,9 @@ const notebookReducer = (state = newNotebook(), action) => {
     case 'TOGGLE_WRAP_IN_EDITORS':
       return Object.assign({}, state, { wrapEditors: !state.wrapEditors })
 
+    case 'UPDATE_NOTEBOOK_INFO':
+      return Object.assign({}, state, { notebookInfo: action.notebookInfo })
+
     case 'EVAL_FRAME_READY': {
       state.evalFrameMessageQueue.forEach((actionToPost) => {
         postActionToEvalFrame(actionToPost)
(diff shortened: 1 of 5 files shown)


src/reducers/notebook-reducer.js
```diff
@@ -55,6 +55,9 @@ const notebookReducer = (state = newNotebook(), action) => {
     case 'TOGGLE_WRAP_IN_EDITORS':
       return Object.assign({}, state, { wrapEditors: !state.wrapEditors })
 
+    case 'UPDATE_NOTEBOOK_INFO':
```
mdboom: I think `IMPORT_NOTEBOOK` already handles the notebookInfo...  Would it be less redundant to just make sure `IMPORT_NOTEBOOK` is fired when making a new notebook, rather than adding a new action?

hamilton: We will be deprecating `IMPORT_NOTEBOOK`, iirc (@bcolloran has more to say about this I think) - in any case I can remove the `getNotebookInfo` call from it so we don't have this redundancy

mdboom: Ok.  Sure thing.

bcolloran: ya importNotebook and IMPORT_NOTENOOK are kind of legacy -- they're meant to handle e.g. loading from localStorage and the URL trick, which may end up on the chopping block. And importNotebook is only called from within importInitialJsmd; this action+reducer will give us a way to update the clients knowledge about the notebook at any time, not just at load time.

PR #944, opened 2018-09-19
make link opening great again

### LINKED ISSUE #884: all notebook actions that open a page should do so in a new tab

### LINKED ISSUE #885: we should make sure that any links within a report view open in a new tab
if we don't force links to open in a new tab and a user clicks a link in a report, the iframe navigates to that page but the editor does not, and the result is weird and confusing

### LINKED ISSUE #941: links do not work from within the report!!! eek!!
@hamilton anything to add to this issue?

bcolloran:
fixes #941, fixes #884, fixes #885

### DIFF
--- a/server/templates/base.html
+++ b/server/templates/base.html
@@ -3,6 +3,7 @@
 <html lang="en">
 <head>
   {% block style %}{% endblock %}
+  {% block targetblank %}{% endblock %}
 </head>
 <body>
   {% block content %}{% endblock %}
--- a/server/templates/base.html
+++ b/server/templates/base.html
@@ -2,8 +2,7 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-  {% block style %}{% endblock %}
-  {% block targetblank %}{% endblock %}
+  {% block head %}{% endblock %}
 </head>
 <body>
   {% block content %}{% endblock %}
--- a/src/components/modals/help-modal.jsx
+++ b/src/components/modals/help-modal.jsx
@@ -70,7 +70,7 @@ export default class HelpModal extends React.Component {
           <Tabs value={value} onChange={this.handleChange}>
             <Tab label="Keyboard shortcuts" />
             <Tab label="More resources" />
-            <Tab label="About Iodide" href="#basic-tabs" />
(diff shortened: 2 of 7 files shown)


src/components/modals/help-modal.jsx
```diff
@@ -70,7 +70,7 @@ export default class HelpModal extends React.Component {
           <Tabs value={value} onChange={this.handleChange}>
             <Tab label="Keyboard shortcuts" />
             <Tab label="More resources" />
-            <Tab label="About Iodide" href="#basic-tabs" />
+            <Tab label="About Iodide" />
```
bcolloran: the `href` here made this tab button render as an <a> which was not needed and caused clicking the tab label to open a new browser tab

server/templates/base.html
```diff
@@ -3,6 +3,7 @@
 <html lang="en">
 <head>
   {% block style %}{% endblock %}
+  {% block targetblank %}{% endblock %}
```
wlach: Do we need this block? Why not put the content in directly?

bcolloran: My thinking was: we definitely want all links that are clicked _within a notebook_ to open a new tab (because eval state could be lost), but we might not want links that are included on server pages to always open a new tab. It was my understanding that this template is used on both server pages and on the notebook. Is that incorrect? Or alternatively, even if that is correct, do you think we should always force links to open in a new tab?

wlach: That totally makes sense to me, but you can just put the needed directive inside notebook.html, no need for the block.

bcolloran: hey will, i'm not sure how to put this directly in notebook.html while also wrapping it in the <head> tag. Would we need to replace the `{% block style %}` with a  `{% block head %}` across the board? or is there a way of injecting something into a tag from within a child template?

bcolloran: talked to @wlach -- the solution is to rename the "style" block to "head" and to but the <base> tag in the head block within the notebook template.

Issue #896, opened 2018-09-13
add interstitial banner state for "this notebook is in standalone mode"


bcolloran:
For a notebook that is not connected to iodide.io, the warning banner should give a message that warns the user they're not connect, and offers them the opportunity to load their work at iodide.io (which will require a server API of some kind), or dismiss the banner if they want to work in standalone mode

mdboom:
How do we know whether to recommend the public server or some other server?  We're just defaulting to the public one for now?

bcolloran:
I think we'll be able to read that from the .env file right? we'll have to change the host names in that file for each deployment anyway.

mdboom:
For standalone, the stuff in .env relates to where the standalone js is served from.  I don't there's a notion of pointing to "another thing" yet...

bcolloran:
i think we'll need to clarify what we mean by "standalone" mode a little bit next week -- i'm thinking of it as meaning the files that are served from an exported archive, and which are not served from any iodide server instance (including the one at localhost).

Under that interpretation, i was imagining the following:
- for a given iodide server deployment (whether our main deployement at iodide.io, a  private deployment at mozilla etc, or a localhost deployment for testing/development), you have to set the correct URIs in the .env
- those propogate forward into the webpack build of the JS assets
- so if you're using an iodide NB from a server deployment and you export an archive, when the archive is run the URI from the originating deployment will be available (iodide.io or iodide.mozilla.org or localhost (this latter will still be useful for testing))

i suspect we're just talking about different things? what did you have in mind?

mdboom:
I see.  I think currently, we have a notion of a standalone build, which is totally unaware of any server.  I think you are suggesting that when you export a notebook from a server, that exported notebook would be able to link back to the server it was created from...?

bcolloran:
blocked by #1008
fyi @hamilton

PR #952, opened 2018-09-20
Fix #766: Export should point to resources on exporting server

### LINKED ISSUE #766: [BUG] Exporting notebooks from extremely-alpha
accidentally listed this on pyodide: https://github.com/iodide-project/pyodide/issues/113

the gist:
When a notebook is exported from https://extremely-alpha.iodide.io, it does not run locally. upon further inspection, it looks as though the iodide.dev.css and the idodide.dev.js files are given relative paths instead of absolute paths at export.

mdboom:
When exporting a notebook, is was exporting a relative path to the .js and .css.  This fixes it to point back to the code from the server doing the exporting.

Fixes #766 and is a necessary first step for #896.

Also fixes `/new` not having an iframe which was broken by #871.

### DIFF
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -68,5 +68,9 @@ def new_notebook_view(request):
     return render(request, 'notebook.html', {
         'user_info': _get_user_info_json(request.user),
         'notebook_info': {'user_can_save': True},
-        'jsmd': ''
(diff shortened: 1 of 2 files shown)


bcolloran:
lgtm thanks mike

PR #961, opened 2018-09-21
fixes the "delete this notebook" menu item appearing for users who are not the current logged-in user

hamilton:


hamilton:
Deployed this on `extremely-alpha` to test it myself with other user data. These are small tweaks that don't really need review, and I'd like to unblock server sprinting next week when we're all together, so I'm going to merge this.

PR #962, opened 2018-09-21
Make username regex case insensitive.

mdboom:
As reported by @hamilton on gitter, `DevinBayly` can't list notebooks... :)


hamilton:
lgtm - thanks!

PR #963, opened 2018-09-21
Fix the username URL -- use \w instead of (?i)

mdboom:


hamilton:
This one actually passes the tests :) thanks mike!

PR #967, opened 2018-09-24
Allow serving static iodide assets for iframe via server (fixes #829)

### LINKED ISSUE #829: Deploying server-based assets to ghpages has bugs and limitations

wlach:
Use Django staticfiles support to serve the iframe assets (and the iframe
assets only, no api endpoints or anything requiring authentication) from
the heroku server.


wlach:
Once this is merged, we can direct extremely-alpha.iodide.app to heroku and not worry about using the github pages installation to support heroku. \o/ I tested this with using iodide-server.herokuapp.com to serve the static assets and it seems to work great. Attempts to fetch API endpoints from the herokuapp domain fail as expected.

bcolloran:
i have not done manual QA, but this all looks reasonable.

PR #972, opened 2018-09-27
Improve nested array reps

mdboom:
Rather than saying "n length array" for each nested array within an array, those messages are now hidden for nested arrays.

image

### DIFF
--- a/src/components/reps/array-handler.jsx
+++ b/src/components/reps/array-handler.jsx
@@ -13,12 +13,12 @@ const MAX_ELTS = 200
 const HALF_MAX_ELTS = Math.round(MAX_ELTS / 2)
 
 export default {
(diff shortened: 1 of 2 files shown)


src/components/reps/array-handler.jsx
```diff
@@ -53,12 +53,26 @@ export default {
         <span key="array_elt_empty" title="empty array" />
       )
     }
+
+    if (inContainer) {
+      return (
+        <div className="array-rep">
+          [
+          <ul>
```
bcolloran: what is the idea behind the adding the <ul> and the line breaks? I think overall i'd prefer eg
`[42, [32, 5]]` to having it broken out across multiple lines in block level elements.

PR #974, opened 2018-09-27
WIP: Make strings expand when clicking on them

mdboom:
This allows clicking on a truncated string to display the entire string.

There's probably a lot of unworking edge cases here, and I'm not sure this is the best way to do this, but I think it might be a good place to begin the discussion.


mdboom:
There must be a CSS difference here.  This is what it looks like for me:

image

Are you on Chrome or Firefox?  I wonder if that's the issue...

But point taken about adding something *next to* the string to click on, rather than clicking on the string itself.

mdboom:
Indeed, it looks to be a Chrome CSS difference.  Looking into a cross-browser solution...

PR #979, opened 2018-09-27
Fix #978: Don't deploy static assets to Github Pages from Travis

### LINKED ISSUE #978: Remove Travis autodeploy of static assets
Now that the server is serving all of the static assets, there's no longer a need for travis to automatically deploy to github pages.

*There is no plan to remove the existing assets, so URLs won't break*, we'll just stop pushing new content there.

mdboom:


bcolloran:
@wlach @hamilton  please review this one in the light of issue #980

hamilton:
Assigning myself to this so that once we have a PR that addresses #1008 I can review.

webpack.config.js
```diff
@@ -130,7 +127,7 @@ module.exports = (env) => {
         IODIDE_JS_PATH: JSON.stringify(APP_PATH_STRING),
         IODIDE_CSS_PATH: JSON.stringify(CSS_PATH_STRING),
         IODIDE_BUILD_MODE: JSON.stringify((env && env.startsWith('dev')) ? 'dev' : 'production'),
-        IODIDE_BUILD_TYPE: JSON.stringify((env && env.includes('client-only')) ? 'standalone' : 'server'),
+        IODIDE_BUILD_TYPE: 'server',
```
hamilton: Due to recent PRs we've removed any use of `IODIDE_BUILD_TYPE` from the client code. Is `IODIDE_BUILD_TYPE` needed anywhere else at this point, specifically Travis?

mdboom: I don't think so.  I'll rebase this, which should retrigger a travis run to test it.  But we basically wouldn't be using travis to deploy at all anymore with this change (just the heroku deployment which is totally separate), so if the tests pass, I think we're good to go.

hamilton:
Great. I think that means we can remove that global declaration in
package.json too

PR #986, opened 2018-09-28
WIP: upload files (data, images, scripts) to the server for notebook access

hamilton:
## __*NB: not ready for review.*__

This is part of a larger project of bringing data into notebooks. This PR:

- [x] creates a new File model, associated api, and access points in the server, along with some rudimentary tests.
- [x] adds a first-pass of an upload feature from a server page.
- [x] lists all of the data sources on the revisions page (for now)
- [ ] reformulate `fetchWithCSRFToken` to be more generic (remove `Content-Type` mainly) and migrate existing use (entirely for delete) to `deleteWithCSRFToken` or something like that
- [ ] implement `PUT` for additional copies of the data

This PR will NOT do these:
- provide an easy and filesystem-like way to access data uploaded (this will be approached once we have put together `iodide.server.fetch` and / or the fetch cell)
- have a way of saving data from a notebook directly onto the server (possibly done later with something like `iodide.server.saveFile`
- display all the available saved files in the notebook itself (this will be addressed in a later PR to prevent bloat on this one)


server/files/migrations/0003_more_updates.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.1 on 2018-09-27 19:23
```
wlach: We should squash all these migrations before merging. Basically delete all the migrations then generate a new one via:

    ./manage.py makemigrations -n add_file_model

lmk if you want me to help with this.

server/notebooks/views.py
```diff
@@ -53,6 +54,14 @@ def notebook_revisions(request, pk):
         'title': nb.title,
         'notebookId': nb.id
     }
+    files = [
+        {'filename': file.filename, 
+        'id': file.id, 
+        'last_updated': file.last_updated.isoformat(sep=' '),
+        'size': len(file.content)}
```
wlach: This is probably going to be horribly slow if the file is at all big... we're probably going to move to a better/different storage solution sooner than later but for now please at least add a `FIXME` comment about how this is inefficient.

server/notebooks/views.py
```diff
@@ -53,6 +54,14 @@ def notebook_revisions(request, pk):
...
+        for file in File.objects.filter(notebook_id=pk)
+    ]
+    print([len(file.content) for file in File.objects.filter(notebook_id=pk)])
```
wlach: Remove print statement

src/server/components/files-list.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+const bytesToHumanReadable = (bytes) => {
+  const KB = 1024
+  const MB = KB * 1024
+  const GB = MB * 1024
+  const TB = GB * 1024
+  const round = (n, k = 2) => Math.round(n * ((10 * k) || 1)) / (10 * (k) || 1)
+  if (bytes < 1024) return `${bytes}b`
+  if (bytes < MB) return `${round(bytes / KB, 0)}kb`
+  if (bytes < GB) return `${round(bytes / MB)}mb`
```
wlach: We might just want to use numeral for this: http://numeraljs.com/

No doubt there are other places where we would want to display large quantities in a more human readable form

src/shared/fetch-with-csrf-token.js
```diff
@@ -12,11 +12,20 @@ export function getCookie(name) {
...
 }
 
-export default function fetchWithCSRFToken(url, otherParts) {
+// for POST of files, primarily
+export default function fetchWithCSRFToken(url, otherParts, headers = {}) {
   const csrfToken = getCookie('csrftoken')
-  const headers = {
-    'Content-Type': 'application/json',
+  const defaultHeaders = {
+    // 'Content-Type': 'application/json', // THIS NEEDS TO BE REPLACED NOW!
```
wlach: yup this needs fixin' before landing :)

hamilton: I believe I actually separated this out already, so we can delete this commented-out line. See the function `fetchWithCSRFTokenAndJSONContent` below.

hamilton:
@wlach I ran out of time on this one before leaving town, but I did fix the tests. I'll circle back on this on Tuesday.

hamilton:
@wlach I migrated to `numeral.js` to format filesizes on the server. Believe this is ready for another review when you get a chance.

wlach:
@hamilton I get an error when I try to upload a file from the notebooks page complaining that `TypeError: this.props.onUploadFile is not a function`:

screen shot 2018-10-10 at 3 06 39 pm

I suspect something small just needs updating here.

PR #988, opened 2018-09-29
re-implements syntax highlighting for markdown / css cells

### LINKED ISSUE #942: MD / CSS cells no longer have code highlighting

hamilton:
Fixes #942.

### DIFF
--- a/src/components/cells/cell-editor.jsx
+++ b/src/components/cells/cell-editor.jsx
@@ -185,8 +185,7 @@ function mapStateToProps(state, ownProps) {
       && state.languages[cell.language]
       && state.languages[cell.language].codeMirrorModeLoaded)
     ? state.languages[cell.language].codeMirrorMode
-    : ''
-
+    : (cell.cellType || '')
   const editorOptions = {
     mode: codeMirrorMode,
     lineWrapping: false,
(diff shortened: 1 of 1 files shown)


src/components/cells/cell-editor.jsx
```diff
@@ -185,8 +185,7 @@ function mapStateToProps(state, ownProps) {
       && state.languages[cell.language]
       && state.languages[cell.language].codeMirrorModeLoaded)
     ? state.languages[cell.language].codeMirrorMode
-    : ''
-
+    : (cell.cellType || '')
```
hamilton: @mdboom @bcolloran is this enough?

hamilton: On second thought, `(cell.cellType || '')` will always work even if the language in `cell.cellType` hasn't been loaded.

bcolloran: @hamilton you mean this will always work if `cell.cellType === 'code'` but the CM language mode is not loaded? in that case it will set `codeMirrorMode = 'code'`.

This code looks clean to me, so if you've you verified that returning `codeMirrorMode = 'code'` works ok with e.g. a pyodide NB, then i think you should merge, otherwise we might need a new guard against thta condition

bcolloran:
lgtm!

PR #989, opened 2018-09-29
Iodide API now visible in the declared variables pane when a notebook with jsmd is loaded

### LINKED ISSUE #640: iodide api not in declared variables until you evaluate a cell
This is one thing that was brought up by @DevinBayly #638. I would think we want that to be clearly available / inspectable to users, especially new users.

hamilton:
Fixes #640.

### DIFF
--- a/src/eval-frame/reducers/eval-frame-reducer.js
+++ b/src/eval-frame/reducers/eval-frame-reducer.js
@@ -27,16 +27,6 @@ function addAppMessageToState(state, appMessage) {
   return state
 }
 
-const initialVariables = new Set(Object.keys(window)) // gives all global variables
-initialVariables.add('__core-js_shared__')
-initialVariables.add('Mousetrap')
-initialVariables.add('CodeMirror')
-
-function getUserDefinedVarsFromWindow() {
-  return Object.keys(window)
-    .filter(g => !initialVariables.has(g))
-}
-
 const notebookReducer = (state = newNotebook(), action) => {
   let nextState
   switch (action.type) {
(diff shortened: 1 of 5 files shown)


src/eval-frame/reducers/eval-frame-reducer.js
```diff
@@ -43,13 +47,12 @@ const notebookReducer = (state = newNotebook(), action) => {
     case 'UPDATE_EVAL_FRAME_FROM_INITIAL_JSMD': {
       const cells = action.stateUpdatesFromEditor.cells
         .map(c => Object.assign(newCell(), c))
-      console.log('cells cells cells', cells)
       const newState = Object.assign(
         newNotebook(),
         action.stateUpdatesFromEditor,
         { cells },
+        { userDefinedVarNames: getUserDefinedVarsFromWindow() },
```
bcolloran: architectural comment: since we have started being more careful about keeping impure side-effect-y functions like `getUserDefinedVarsFromWindow()` in action creators rather than the reducer, i think this would be a case in which we should follow that practice.

(just added issue #995 to track this general cleanup)

hamilton: `getUserDefinedVarsFromWindow` is not impure, though from what I'm understanding you want `userDefinedVarNames` to be built via an action creator and not collected from `window` in a reducer, correct?

hamilton: on second thought I see your point. Will circle back on this tomorrow

hamilton:
@bcolloran made the changes you suggested. Let me know if there's anything else to tweak here before merging.

bcolloran:
looks great, thanks hamilton!

PR #1001, opened 2018-10-03
removed unused css reference in index.html

### LINKED ISSUE #911: make sure correct MIME types being served
```Refused to apply style from 'http://localhost:8000/server.home.dev.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.```

hamilton:
closes #911. Not much review necessary. We weren't using the missing stylesheet that wasn't being generated at any rate.


bcolloran:
lgtm, tho i'm honestly not that familiar with the server side stuff

Issue #915, opened 2018-09-18
add UI to show parent notebook of notebook copied from another user


bcolloran:
users will want to be able to trace the provenance of notebooks; #836 is about the data model, this is about putting in some UI to expose the data

hamilton:
@bcolloran was this resolved with #1002?

bcolloran:
yep!

PR #1002, opened 2018-10-03
add forking functionality

### LINKED ISSUE #836: Be able to trace origins of forked notebooks
Follow-on from #791. I want to be able to fork a notebook from another user and for people viewing my notebook to be able to trace it back to the original.

I believe this has three parts:

* Adding a "forked_from" field (can be `null` for new notebooks) to the notebook model: https://github.com/iodide-project/iodide/blob/99f7883a996d16361094b08819fcd6bc9b7ade8b/server/notebooks/models.py#L5
* Adding this metadata to the various notebook API methods
* Exposing this information in the UI

hamilton:
screen shot 2018-10-02 at 10 42 44 pm

I think this could probably be refined, but figured I'd put this PR together to get some feedback.

Closes #836 and #915. Specifically, it:

- [x] adds a `forked_from` field in the `Notebook` model, which is a `ForeignKey` on the `NotebookRevision`. This allows the history traceability back to the notebook revision that spawned the new notebook. This does _not_ address adding a `parent_id` to each revision at the moment. I'm not sure what having a `parent_id` actually buys us in terms of workflow.
- [x] in the Notebook API, if you add `forked_from: <id>` to the post request, it will add this field, otherwise defaults to `null`. Adding a `forked_from` id that does not exist should throw an error.
- [x] now you can fork from the notebook interstitial directly, and this will be the only supported way to fork for this PR. Future PRs will add further forking support in menus etc.
- [x] on the revisions page you see if / who / which notebook the forked notebook came from.
- [x] write a few server tests to ensure forking is handled correctly.


server/notebooks/migrations/0002_notebook_forked_from.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 2.1 on 2018-10-03 01:54
```
hamilton: @wlach I should probably remove this, yeah? There is another migration added to the repository so I added.

server/tests/test_notebook_api.py
```diff
@@ -105,3 +108,47 @@ def test_delete_notebook_owner(fake_user, test_notebook, client):
...
+        'forked_from': 1
+    }
+
+
+def test_fork_notebook_not_logged_in(client, notebook_fork_post_blob):
+    resp = client.post(reverse('notebooks-list'), notebook_fork_post_blob)
+    assert resp.status_code == 403
+
+
+def test_incorrect_fork_notebook(client, fake_user, incorrect_notebook_fork_post_blob):
```
hamilton: This set of assertions feels incomplete.

server/notebooks/migrations/0002_notebook_forked_from.py
wlach: Fine to leave the comment in.

server/tests/test_notebook_api.py
```diff
@@ -105,3 +108,47 @@ def test_delete_notebook_owner(fake_user, test_notebook, client):
...
+
+def test_fork_notebook_logged_in(client,
+                                 fake_user,
+                                 fake_user2,
+                                 test_notebook,
+                                 notebook_fork_post_blob):
+    client.force_authenticate(user=fake_user)
+    resp = client.post(reverse('notebooks-list'), notebook_fork_post_blob)
+    assert resp.status_code == 201
+    assert resp.json()['forked_from'] == test_notebook.revisions.get().id
```
wlach: I would like to see another test that deletes the notebook that another was forked from, verifying that the forked notebook still exists.

src/actions/actions.js
```diff
@@ -269,6 +269,13 @@ function getNotebookSaveRequestOptions(state) {
     title: state.title,
     content: exportJsmdToString(state),
   }
+  const revisionID = state.notebookInfo.revision_id
+  const userOwns = state.notebookInfo.user_can_save
+  const forkedFrom = revisionID || undefined
```
wlach: the `|| undefined` part seems redundant?

src/actions/actions.js
```diff
@@ -269,6 +269,13 @@ function getNotebookSaveRequestOptions(state) {
...
+  // only set forked_from if user doesn't own this notebook.
```
wlach: Hmm, I feel like we might want to keep track of the history even if the user is forking their own notebook?

hamilton: good point

hamilton: Going to generalize this a little bit so we can make it easy for a user to fork their notebook from a menu item, etc.

server/tests/test_notebook_api.py
hamilton: added this test in a commit below.

hamilton:
@wlach I believe I addressed the your comments w/ this most recent set of commits - feel free to take a look at your convenience and let me know if this looks good.

wlach:
Looks good now! I'll merge.

PR #1006, opened 2018-10-03
Fix #1003: Add toggle button for default rep

### LINKED ISSUE #1003: Create a way to display rep in "default" (object inspector) rep
A button on the upper right of the rep to display it using `react-inspector`'s object inspector rather than the heuristically determined rep.

mdboom:
image

image

Known issues:
- [x] The button is a little higher than the default height of a single line of text. Something about our button CSS is making that hard to override
- [x] I have no idea what sort of icon to use for this.

### DIFF
--- a/src/components/reps/value-renderer.jsx
+++ b/src/components/reps/value-renderer.jsx
@@ -3,7 +3,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
-import StarRate from '@material-ui/icons/StarRate'
 import Tooltip from '@material-ui/core/Tooltip'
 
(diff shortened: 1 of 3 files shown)


hamilton:
Awesome. 

- I think we could easily make an svg icon that is just `{ }` to represent this.
- I don't know what the "return to default rep" button icon should look like, but it could just be an `x` or something like that as a starting point.

mdboom:
I think this is ready for final review.  I think this takes a lot of pressure off of various rep details for MVP.

bcolloran:
lgtm, let's take it so we can test it

PR #1014, opened 2018-10-04
Fix #1011: Seamlessly load language plugins

### LINKED ISSUE #1011: Seamless Python user
As a user, I'd like to be able to set a cell to "Python type" and then when evaluating, implicitly download the resource for Python and then run the cell.

mdboom:
The idea here is:

There's a set of "known" language plugins that exist in the store (`languageDefinitions`).  At sometime in the future, we might fetch these from a "directory" of plugins online somewhere, but for now it's semi-hardcoded.  These get tasks and cell menu items automatically.

When you evaluate a code cell, it checks if the given language is already loaded (in `loadedLanguages`).  If not, it does the whole async dance just like the language plugin cell, and then evaluates the cell.

The language plugin cell should continue to work for non-"known" language plugins.


src/components/cells/cell-menu.jsx
```diff
@@ -91,7 +91,8 @@ export class CellMenuUnconnected extends React.Component {
 
 export function mapStateToProps(state, ownProps) {
   const { cellId } = ownProps
-  const userLanguages = Object.keys(state.languages).filter(language => language !== 'js')
+  let userLanguages = Object.assign({}, state.languageDefinitions, state.loadedLanguages)
```
mdboom: We combine the "known" languages with the "loaded" languages, since the user may have loaded an unknown language.

src/index.jsx
```diff
@@ -36,6 +37,7 @@ initializeDefaultKeybindings()
 
 window.addEventListener('message', listenForEvalFramePortReady, false)
 
+handleLanguageDefinitions(store)
```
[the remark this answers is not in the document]
hamilton: This seems like the right place to do it, imo, but @bcolloran might have other ideas here

src/handle-language-definitions.js
```diff
@@ -0,0 +1,9 @@
+import { addChangeLanguageTask } from './actions/task-definitions'
+
+export default function handleLanguageDefinitions(store) {
+  const state = store.getState();
+  const { languageDefinitions } = state;
+  for (const language of Object.values(languageDefinitions)) {
+    addChangeLanguageTask(language.languageId, language.displayName, language.keybinding);
```
hamilton: Launched a notebook, tried typing `p` but didn't get python. Could we also make sure to add the language keybinding?

src/components/cells/cell-menu.jsx
bcolloran: i spent a while stuck on what userLanguages was referring to here. in this updated context, maybe it makes more sense to rename to something like  "exposedLanguages" or "availableLanguages" or something that clarifies that these are the languages that the UI is aware of, even if they are not yet actually available in memory.

src/components/cells/cell-menu.jsx
```diff
@@ -91,7 +91,8 @@ export class CellMenuUnconnected extends React.Component {
...
+  userLanguages = Object.keys(userLanguages).filter(language => language !== 'js')
```
bcolloran: perhaps if we're going this way, it makes just as much sense to remove this line and remove the special case NotebookMenuItem for js? won't that be generated automatically by the same mechanism that generates menu items for all the other languages?

src/eval-frame/actions/actions.js
```diff
@@ -128,6 +129,124 @@ export function consoleHistoryStepBack(consoleCursorDelta) {
...
 }
 
+export function addLanguage(languageDefinition) {
+  return {
+    type: 'ADD_LANGUAGE_TO_EVAL_FRAME',
+    languageDefinition,
+  }
+}
+
+function loadLanguagePlugin(pluginData, historyId, cell, dispatch) {
```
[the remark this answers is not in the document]
bcolloran: actions.js is getting too long. maybe it would be more readable overall (and make the diff easier to follow as well) if you broke the language functions out into language-actions.js?

src/eval-frame/actions/actions.js
```diff
@@ -195,28 +318,36 @@ function evaluateCodeCell(cell) {
...
+          evalStatus = 'ERROR'
+        }
+        const updateCellAfterEvaluation = () => {
+          const cellProperties = { rendered: true }
+          if (evalStatus === 'ERROR') {
+            cellProperties.evalStatus = evalStatus
+          }
+          dispatch(updateCellProperties(cell.id, cellProperties))
+          // dispatch(incrementExecutionNumber())
+          dispatch(updateValueInHistory(historyId, output, 'CELL_EVAL_VALUE'))
```
bcolloran: as above, i think we shouldn't have to change the historyType here

src/eval-frame/actions/actions.js
```diff
@@ -73,11 +73,12 @@ export function appendToEvalHistory(cellId, content, value, historyOptions = {})
   }
 }
 
-export function updateValueInHistory(historyId, value) {
+export function updateValueInHistory(historyId, value, historyType) {
   return {
     type: 'UPDATE_VALUE_IN_HISTORY',
     historyId,
     value,
+    historyType,
```
bcolloran: i don't think we really need to add functionality to change historyTypes, see other comments

src/eval-frame/actions/actions.js
```diff
@@ -195,28 +318,36 @@ function evaluateCodeCell(cell) {
...
-      dispatch(appendToEvalHistory(cell.id, cell.content, output))
-      dispatch(updateUserVariables())
-    }
 
-    const evaluation = Promise.resolve()
-      .then(updateCellAfterEvaluation)
-      .then(waitForExplicitContinuationStatusResolution)
-      .then(() => dispatch(temporarilySaveRunningCellID(undefined)))
-    return evaluation
+    ensureLanguageAvailable(cell.language, historyId, cell, state, dispatch)
```
bcolloran: i'm not sure i'm following everything that is happening here, but i think we can simplify a lot of the stuff around how you're handling history items and passing around historyId, and i think this is the crux of it right here

this `ensureLanguageAvailable` block should go towards the top of evaluateCodeCell, before anything else is added to the eval history; if an eval is attemped on a cell with an unknown code type, _before_ the code/output block for the cell is added to the history, we pre-empt that with the console message that says "fetching resources to run python" (or whatevs), in a 'CELL_EVAL_INFO' history item, and then once that is done we can show the code and output in a "CELL_EVAL_VALUE" history item.

src/eval-frame/reducers/eval-frame-reducer.js
```diff
@@ -79,7 +79,12 @@ const notebookReducer = (state = newNotebook(), action) => {
     case 'UPDATE_VALUE_IN_HISTORY': {
       const history = [...state.history]
       const historyEntry = history.find(h => h.historyId === action.historyId)
-      if (historyEntry) { historyEntry.value = action.value }
+      if (historyEntry) {
```
bcolloran: i don't think we really need to add functionality to change historyTypes, see other comments

src/index.jsx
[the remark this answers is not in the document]
bcolloran: ya seems fine to me

src/handle-language-definitions.js
mdboom: I think this is fixed now.

src/eval-frame/actions/actions.js
mdboom: If we do that (which was my initial pass), you end up with two console output blocks for a single cell -- one to load the language plugin and one for the output.  That didn't feel right to me.  We could add a "remove from history" action to remove the first one once it completes, but I thought it was simpler to just keep updating the same output cell, and changing the `historyType` accordingly.  But it's possible I'm missing a third option...

src/components/cells/cell-menu.jsx
mdboom: That's a good idea.

src/components/cells/cell-menu.jsx
mdboom: Yes.  The only advantage here is that `js` will always appear first (which I think has some, if small, merit).

src/eval-frame/actions/actions.js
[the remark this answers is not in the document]
mdboom: That's a good idea.  Will do.

mdboom:
In light of #1016, I've updated this to handle saving/loading of notebooks a little better.

If the user uses a language plugin cell to add a new language:
- that language is added to `languageDefinitions` as well as `loadedLanguages` in the store
- When saving, any `languageDefinitions` in the store that aren't "built-in" are saved to the metadata in the JSMD.
- When loading JSMD, those saved `languageDefinitions` are added to the store.

This allows custom language cells to be run without explicitly evaluating the language plugin cell first, just as built-in languages, and removing the need for an error message as proposed in #1016.

The reason that the JSMD saves/loads from `state.languageDefinitions` and not `state.loadedLanguages` is because if you were to load a notebook, not run any cells, and then save it again, you would lose language definitions.  That is, `languageDefinitions` is a place to easily store metadata about any languages we know about.  `loadedLanguages` is for the state of which languages are loaded, which, much like the global variables, is not something to serialize / unserialize.

Hope that all makes sense.

src/eval-frame/actions/actions.js
bcolloran: ya that is the preferred behavior -- semantically, loading the plugin is _not_ part of the cell execution, it's an extra step that need to be taken when pyodide hasn't yet been loaded; what we're really doing is equivalent to automatically inserting a load plugin cell (which we will continue to support), and then evaling the code, and we should record those two steps as separate history items to show what is going on and maintain parallelism with the case of explicitly loading a language plugin. it should look something like:

```
============
Pyodide evaluator not yet available locally, fetching resource ...
(file download progress ...)
(initialization progress ...)
pyodide ready
============
>> [x**2 for x in range(5)]
------------
[1,4,9,16,25]
============
```

 i think this is a case of explicit>implicit: we want to be explicit about when we're inserting that extra step, and we want to leave the record of that step in console history to give our users more insight into what is happening behind the scenes.

mdboom: Ok.  I've updated this to do that.

src/handle-language-definitions.js
bcolloran: ah, i'm hitting this as well when i try to go js->py or py->js, but *not* if i go js->md->py or vice versa; i think that #1020 broke this. let's take that a separate issue

bcolloran: filed at #1041

PR #1021, opened 2018-10-05
fixes #1007

### LINKED ISSUE #1007: cannot export notebooks in standalone mode
I am running a server with `npm run start-and-serve`, and I made a quick test notebook & exported it from the menu item.

I get the following error in `validated-reducer.js`:

```
Error: [
 {
  "keyword": "type",
  "dataPath": ".notebookId",
  "schemaPath": "#/properties/notebookId/type",
  "params": {
   "type": "integer"
  },
  "message": "should be integer"
 }
]
bad values:
{
 "notebookId": null
}
```

My understanding from this is we haven't worked on actually changing the schema to support `notebookId` to be `null` (which is the case in standalone mode), so this error is getting thrown. 

Here is the notebook as I wrote it in the the editor & exported:

```html

<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
<title>test - iodide</title>
<link rel="stylesheet" type="text/css" href="http://localhost:8000//iodide.dev.css">
<base target="_blank" rel="noopener noreferrer">
</head>
<body>
<script id="jsmd" type="text/jsmd">
%% meta
{
  "title": "test",
  "lastExport": "2018-10-03T17:15:55.487Z"
}
%% js
var x = 10
</script>
<iframe
  id="eval-frame"
  src="http://localhost:8000//iodide.eval-frame.dev.html"
  sandbox="allow-scripts allow-same-origin"
  allowfullscreen="true"
  allowvr="yes"
></iframe>
<script src="http://localhost:8000//iodide.dev.js"></script>
</body>
</html>

```

bcolloran:


bcolloran:
low risk commit (since it just adds a tiny validation) so i'm going to go ahead and merge when CI passes.

Issue #666, opened 2018-06-19
add URL parameter to over-ride report view


bcolloran:
notebooks that open in report view (and hence run-all immediately) can be difficult to debug if they have a cell that crashes. we should add an url param that can override this behavior

bcolloran:
see `handle-url-query.js`

iFlameing:
can you explain the issue I didn't get it regarding the param you mention.I am new to this project.

bcolloran:
@iFlameing the idea here is to be able to add a url parameter to a notebook URL like e.g.:
`iodide.io/someuser/some-notebook.html?viewMode=REPORT`, and to have that URL parameter value override the initial value of `viewMode` that gets pushed into the redux store.

PR #1027, opened 2018-10-05
Allow sharing of report view from server-side mode

wlach:
Append a "?report=1" parameter when viewing the notebook in report view,
and respect that when reloading the notebook.

### DIFF
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -116,9 +116,19 @@ export function changeMode(mode) {
 }
 
 export function setViewMode(viewMode) {
-  return {
-    type: 'SET_VIEW_MODE',
-    viewMode,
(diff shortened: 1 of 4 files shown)


bcolloran:
+1 this is a good idea, indeed we have an issue for it at #666. re the implementation: basically all of this implementation can go inside of handle-initial-jsmd and handle-url-query, in which case it will be applied to the store before the store is even mounted to redux (so we can skip an extra dispatch and an extra render step)

bcolloran:
oops, pressed wrong button above @ closed pr

wlach:
@bcolloran not sure if I completely understand, as far as I can tell I implemented it how you suggested? Can you be more specific about what you'd like to see changed?

PR #1029, opened 2018-10-06
Make github authentication optional

wlach:
Don't configure/enable github authentication unless a client id is
set in the configuration.


wlach:
@hamilton, can you review this (and merge if it looks ok)? it's actually pretty simple

bcolloran:
this looks pretty simple, reasonable, and low risk to me. i offer my r+ for what it's worth, in case merging this will unblock stuff

bcolloran:
this looks pretty simple, reasonable, and low risk to me. i offer my r+ for what it's worth, in case merging this will unblock stuff

Issue #1025, opened 2018-10-05
Support async language plugin evaluators


mdboom:
As suggested by @hamilton in a PM, it would be nice to be able to make a cell such as:

```
import numpy as np
x = np.array([1,2,3])
```

automatically load the Numpy package and then run the cell.

I'll ignore the Python-side magic necessary to make this happen (which I think is doable in 90%, but not 100% of cases), except to say that the fetch of the package has to happen asyncronously, and only after can the code content itself be executed.  Right now, the `evaluator` callback of a language plugin is synchronous, so we can't really do this.

I'd propose adding a new kind of evaluator to a language plugin `asyncEvaluator(code, messageCallback)` which returns a `Promise`.  (The old kind of sync `evaluator` would remain, since I think it's useful for the simple case -- only languages that cared to do something more sophisticated would have to implement `asyncEvaluator`).  The arguments would be:

- `code` a string of code (as in the current `evaluator`)
- `messageCallback` a callback that the evaluator can call to display progress messages

The messages would be displayed in the output console, each replacing the previous one, and ultimately be replaced by the output of the final Promise step.  Writing an evaluator would then be something like the following pseudocode:

```
function runPythonAsync(code, messageCallback) {
   messageCallback('fetching package')
   return fetch('package.js')
      .then((package) => {
          messageCallback('initializing package')
          return package.initialize()
      })
      .then((initPackage) => {
         messageCallback('running code')
         return runPython(code)
      )
}
```

It would be good to have @bcolloran and @hamilton 's thoughts before starting to implement this.

bcolloran:
tbh, my main thought is that this will take a bit of thought to really digest and consider the implications. can we table this for a while? (until after MVP launch?)

mdboom:
Sure.  I don't think it's a priority.  Just wanted to record my thoughts as it's an obvious follow-on to #1014.

hamilton:
One issue to consider. This point came up from mikes PR that loads python
when you run a cell. It’s kind of a nice trick, but it isn’t useful if I
want to load numpy or matplotlib, since you have to run the pyodide
command, when as a user j just want to start working with matplotlib. So we
might want to consider making this a somewhat higher priority, since the
language loading doesn’t actually feel like enough for the feature to feel
baked.

hamilton:
To folllow up with this, I don't think having the automatic python loading is actually that useful (or interesting) to users unless 

```
import numpy
```

[...]

PR #1038, opened 2018-10-09
Async language evaluator

mdboom:
**This is a follow-on to #1014, which makes the implementation here much easier.**  We probably want to do final review/merge on that first.  Otherwise, just look at the final commit here.

This implements the ability for language evaluators to be asynchronous, as outlined in #1025.  An optional message callback can be sent which will append messages to the console history as any asynchronous stuff happens.  We'll probably want some spinner or something as well, but I think that's probably better explored as a UX issue on its own.

This won't actually do anything with the Python plugin until https://github.com/iodide-project/pyodide/pull/217 is merged.

### DIFF
--- a/src/eval-frame/actions/actions.js
+++ b/src/eval-frame/actions/actions.js
@@ -165,7 +165,11 @@ export function evalConsoleInput(languageId) {
       dispatch(updateUserVariables())
     }
 
-    return runCodeWithLanguage(language, code)
+    const messageCallback = (msg) => {
+      dispatch(appendToEvalHistory(null, msg, undefined, { historyType: 'CELL_EVAL_INFO' }))
+    }
+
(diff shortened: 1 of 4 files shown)


bcolloran:
@mdboom can you rebase off master now that #1014 is merged? i think that'll make the diff easier to follow. Thanks!

bcolloran:
oh, also let us know when https://github.com/iodide-project/pyodide/pull/217 is merged so that we can do some live testing. thanks also on that front! :-)

mdboom:
This is now working and testable against pyodide.

hamilton:
@mdboom feel free to rebase this and I'll take a look

mdboom:
I just rebased this.  Maybe that will resolve the issues you're seeing...?

PR #1040, opened 2018-10-10
fixes sorting issues on home / user page

hamilton:
I understand we will likely be changing this around, but I found the sorting issues on the home / user page were a bit broken, so I 1) have the home page showing notebooks sorted most recent at top, and 2) order the notebooks on a user page in the same way.

### DIFF
--- a/server/views.py
+++ b/server/views.py
@@ -5,7 +5,7 @@ from django.shortcuts import (get_object_or_404,
                               render)
 from django.db.models import Max
 
-from .notebooks.models import Notebook, NotebookRevision
+from .notebooks.models import Notebook
 from .base.models import User
 
 
@@ -29,8 +29,8 @@ def index(request):
                 'userInfo': get_user_info_dict(request.user),
                 # this is horrible and will not scale
                 'notebookList': [
-                    {'id': v[0], 'title': v[1], 'owner': v[2], 'avatar': v[3]}
-                    for v in notebooks
+                    {'id': nb_id, 'title': title, 'owner': owner, 'avatar': avatar}
+                    for (nb_id, title, owner, avatar) in notebooks
                 ]
             }
         }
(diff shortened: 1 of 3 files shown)


hamilton:
good point, will do it that way then merge

hamilton:
@wlach migrated the sorting to the view. Let me know if this seems right.

src/server/pages/user-page.jsx
```diff
@@ -32,7 +32,8 @@ h2 {
 class UserNotebookList extends React.Component {
   constructor(props) {
     super(props)
-    this.state = { notebookList: this.props.notebookList }
+    const { notebookList } = this.props
+    this.state = { notebookList }
```
wlach: Can probably revert changes to this file entirely now.

server/views.py
```diff
@@ -51,7 +59,7 @@ def user(request, name=None):
                 'title': v[1],
                 'last_revision': NotebookRevision.objects
                 .filter(notebook_id=v[0]).last().created.isoformat(sep=' ')
```
wlach: I think this is going to cause a db-lookup-per-notebook, which will be slow. I suspect you can just use the annotation above with some modification.

server/views.py
```diff
@@ -18,15 +19,18 @@ def get_user_info_dict(user):
...
     return render(
         request, 'index.html', {
             'page_data': {
                 'userInfo': get_user_info_dict(request.user),
                 # this is horrible and will not scale
                 'notebookList': [
                     {'id': v[0], 'title': v[1], 'owner': v[2], 'avatar': v[3]}
-                    for v in
-                    Notebook.objects.values_list('id', 'title', 'owner__username', 'owner__avatar')
+                    for v in notebooks
```
wlach: For clarity, I think we should be able to rewrite this as:

```py
'notebookList': [
    { 'id': id, 'title': title, 'owner': owner, 'avatar': avatar } for (id, title, owner, avatar) in notebooks
]
```

server/views.py
hamilton: Good point!

hamilton: fwiw I think we noted that this would be an issue way back when we put this view together

server/views.py
hamilton: @wlach I'm going to go ahead and change this in `index` as well which is where I copied from

PR #1046, opened 2018-10-10
Add tests for error rep

mdboom:
This should have been part of #1044, which I accidentally merged.


src/components/reps/__tests__/error-handler.jsx
```diff
@@ -0,0 +1,25 @@
+import { ValueRenderer } from '../value-renderer' // eslint-disable-line
+import { trimStack } from '../error-handler'
```
hamilton: should be `import errorHandler, { trimStack } from '../error-handler'`

hamilton: (the linter is cranky)

hamilton:
lgtm, thanks!

PR #1047, opened 2018-10-10
Fix console warnings

bcolloran:
this two little changes just clear some console warnings. i think whomever has a moment to look over this first can merge as long as it looks ok (don't need two reviews)


mdboom:
LGTM.

PR #1052, opened 2018-10-10
Fix async language plugin loading

mdboom:
Ensures that a cell has finished execution and returned an output value before executing the next cell, so that run all works on implicitly loaded languages.

As reported on gitter by @bcolloran.

### DIFF
--- a/src/eval-frame/actions/actions.js
+++ b/src/eval-frame/actions/actions.js
@@ -136,11 +136,11 @@ export function consoleHistoryStepBack(consoleCursorDelta) {
 export function evalConsoleInput(languageId) {
   return (dispatch, getState) => {
     const state = getState()
-    let output
     const code = state.consoleText
     // exit if there is no code in the console to  eval
     if (!code) { return undefined }
     const evalLanguageId = languageId === undefined ? state.languageLastUsed : languageId
+    const language = state.loadedLanguages[evalLanguageId]
 
     // FIXME: deal with side-effects for console evals
     // // clear stuff relating to the side effect target before evaling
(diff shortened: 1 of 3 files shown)


bcolloran:
@hamilton i actually think you should do the deeper review here since you're much better versed on async stuff than I am.

keeping in mind the discussion from gitter:
"""
i think that once we adopt the observable approach to promises in iodide-project/iodide#597, we'll be able to say the following:
- code evaluations run in order
- code evaluations that are enqueued a while other code is being evaled will not be initiated or acted on until the the previous evaluation in the queue has returned
- if an evaled code chunk returns a promise, further evaluations will be enqueued and will not proceed until the promise resolves (this is provided as a bit of sugar to help users unfamiliar with JS async avoid some of that complexity) (depends on #597)
- if asynchronous code is used within an evaled chunk but no promise is returned by the evaled block, subsequent evals will begin immediately. in this case, it is up to the user to manage the complexity of async code, as they would in any other JS app (though we may provide additional control flow apis to help)
- some kinds of code chunks may require additional resources (evaluators) to be loaded before they can be evaled; in that case, the resources will be loaded before the evaluation is attempted (essentially, a fetch and load step will be inserted into the eval queue)
"""

it would be good to put this through the async/run+all stress tests notebooks that you used a number of months ago when you were diving deep on this stuff.

mdboom:
I agree, @hamilton, I may not have caught all async corner cases here.

mdboom:
As an aside, one thing we could easily do given this refactor is: post the input code to the console pane immediately and show a spinner until cell execution is done, and then show the output.  Right now (and previous to this change), nothing goes into the history until execution is done.

hamilton:
@mdboom there aren't a ton of async corner cases, from what I understand - my tests around the async stuff concerns (primarily) the API, and in that regard I believe the coverage is prety good. 

That said, it's clear that a lack of tests around `actions` are _really_ hurting us.

At any rate, running through some async examples I have lying around.

hamilton:
@mdboom your idea mirrors some of that internal feedback we got, so I think that's a great idea.

src/eval-frame/actions/actions.js
```diff
@@ -152,31 +152,23 @@ export function evalConsoleInput(languageId) {
...
+      dispatch(updateConsoleText(''))
+      dispatch({ type: 'CLEAR_CONSOLE_TEXT_CACHE' })
       dispatch(appendToEvalHistory(null, code, output))
       dispatch(updateUserVariables())
     }
 
-    const evaluation = Promise.resolve()
-      .then(updateAfterEvaluation)
+    return runCodeWithLanguage(language, code)
+      .then(output => updateAfterEvaluation(output))
```
hamilton: the anonymous function is actually not necessary, since 

```javascript
...
  .then(updateAfterEvaluation)
```
does the same thing (output is what the `runCodeWithLanguage` function outputs, so next in the `then` pipe is `output`)

mdboom: Good catch.  This came from the more complex version of this in `evaluateCodeCell` and I didn't fully simplify.

mdboom:
Ok.  I've fixed the linting and tests here.  Probably ready for a final review.  Once merged, I'll rebase #1038 and we should be golden for implicit package loading.

hamilton:
lgtm - going to merge

Issue #1045, opened 2018-10-10
handle-initial-jsmd throws error on loading notebook


djbarnwal:
### What I Did
`npm run start-and-serve`

### What I Expected

Cell executions to work

### What Happened 

Somehow the UI is up but executing any cell doesn't reflect any changes. Console shows error on the line `handle-initial-jsmd.js:22:2`  with the message saying `TypeError: can't convert undefined to object`

PR #1056, opened 2018-10-11
Fix bug in handle-initial-jsmd

djbarnwal:
Fixes Issue #1045


hamilton:
Sorry @djbarnwal - completely unaware of this PR, I fixed this in another branch that got merged. Thanks for the PR, and apologies for not keeping better track so that we can actually use this one! Thankfully it's a small fix :)

Issue #1000, opened 2018-10-02
as a user, I want to be able to do some basic styling on the body element of the report


hamilton:
[example notebook that demonstates the issue](https://extremely-alpha.iodide.io/notebooks/26/). To summarize:

- I can't select only the report css without affecting all other css.
- The only method I have to select the report css is really not intuitive.

bcolloran:
"should have a plan around this, though perhaps not immediately" sounds like a pretty good definition of p3 to me :-)

MVP functionality need not include supporting advanced css styling _at all_ -- for the MVP, if users can eval code and use MD, we're at parity with Jupyter; something like that is "minimum", and there is much higher priority core workflow stuff still to tackle. advanced styling will be important, but is not doing it perfectly is not _core_ to our initial offering

hamilton:
I agree that the default report view is not "broken", but css is generally broken in the eval frame right now, and badly enough that we ought to just disallow any styling if we're not going to have a plan before MVP. I can't speak for everyone, but this has been a _major_ rough edge with using Iodide for a large class of real work (dashboarding, actual reports) for me already.

bcolloran:
@hamilton has proposed an idea about using shadowDOM for the implementation of this. @hamilton please feel free to add any additional details along those lines.

PR #1057, opened 2018-10-11
Set alias for Report CSS

djbarnwal:
Solves Issue #1000 

This might be a hacky fix.
We can edit the report CSS by referencing `#report`
Something such as 
```
#report {
  background:#555;
}
```

in the CSS cell would change the report background.


bcolloran:
thanks @djbarnwal -- @hamilton and i discussed this the other day, and he has some ideas about how to do this using shadow DOM, and some wariness about introducing a temporary workaround that we're likely to end up rolling back, so i'm going to close this one

PR #1060, opened 2018-10-12
ipynb import no longer needs language definitions

mdboom:
In fact, it's best to use whatever the latest is at the time of import, which will now happen implicitly.


bcolloran:
lgmt @mdboom. i didn't do manual QA on this, but if it's working in your QA testing, please merge when you wish

PR #1061, opened 2018-10-12
Add dockerflow middleware and circleci scripts

robotblake:


server/settings.py
```diff
@@ -51,6 +51,7 @@
 # Application definition
 
 INSTALLED_APPS = [
+    'dockerflow.django',
```
wlach: Could you make this optional, configured through an environment variable above? I don't think we want dockerflow for the heroku environment we're using.

server/settings.py
```diff
@@ -78,6 +81,14 @@
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
 ]
 
+if DOCKERFLOW_ENABLED:
+    INSTALLED_APPS.extend([
```
wlach: I would just use `append` rather than `extend` here, though it's not that big a deal.

robotblake: Oof, yeah, I'll make that change.

robotblake:
Yeah, I'll wire it up before merging.

PR #1070, opened 2018-10-17
Fix dockerflow / docker container building

robotblake:
* Re-arrange the Dockerflow middleware to fix Dockerflow endpoints
* Fix collectstatic to not overwrite built assets
* Set up dockerignore file to speed up build times

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.git/
+node_modules/
+static/
+.env
+Dockerfile
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 build/
 prod/
 dev/
+static/
 
 # Logs
 logs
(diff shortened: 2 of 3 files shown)


server/templates/index.html
```diff
@@ -2,5 +2,5 @@
 {% block content %}
 <div id="page"></div>
 {{ page_data|json_script:"pageData" }}
-<script src="/server.home.dev.js"></script>
+<script src="/server.home.dev.js""></script>
```
bcolloran: two quotes in a row after 'js' looks fishy... should that be:
```suggestion
<script src="/server.home.dev.js"></script>
```

robotblake: Whoops, accidental typo from something I was working on.

bcolloran:
sweet, thanks @robotblake :-)

Issue #1022, opened 2018-10-05
we should get the notebookId in a more robust way than parsing it out of the URL


bcolloran:
re this statement in the code:
https://github.com/iodide-project/iodide/blob/5ec44be65f20a299eb2e9843fdb7e3a52e375fc3/src/reducers/notebook-reducer.js#L94

an integer parsing issue caused #1007, which is now fixed, but it seems like a brittle way of doing this. can we pass this in with the notebookInfo on the window object, or otherwise get the id via  an api response or something?

not top priority, but something to be aware of

bcolloran:
fyi @hamilton and @wlach

wlach:
Yeah we should probably just pass it down as metadata inside the document.

hamilton:
closed by above-referenced PR.

PR #1075, opened 2018-10-18
removes state.notebookId, fixes issue with brittle url parsing to retrieve notebookId, utilizes state.notebookInfo.notebook_id instead

### LINKED ISSUE #1071: saving a notebook on extremely-alpha appears to be forking instead of saving a revision
This is a code-red bug. Something has broken the server (possibly from merging #1061, but could also be #1068) such that each notebook save forks the notebook instead of saves a revision. I have no idea how this might have happened, but we need to address it as soon as we can.

I can reproduce this on extremely-alpha, but not on my own local build.

hamilton:
closes #1071, #1022. I've done some QA and it appears to fix the issue, but I want to be sure. Please test this thoroughly:

1.) on server, from `/new`, make sure that it saves appropriately and there are no errors
2.) from a saved notebook, make changes to the title and save. It should not fork. Also, check to see that the title updates on the server page, and also that under revisions it updates and save as you would expect.
3.) make sure reload + save does not fork.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4418,14 +4418,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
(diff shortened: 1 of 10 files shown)


hamilton:
@bcolloran has checked this PR and given the 👍 so I'm merging.

PR #1078, opened 2018-10-18
fixes the ?viewMode=report query param bug, and better organizes many of the notebook initialization functions

hamilton:


hamilton:
Yep, will do now and then merge if it checks out. Thanks for the reviews @wlach @bcolloran

hamilton:
Thanks for the reviews @wlach + @bcolloran

PR #1081, opened 2018-10-18
add interstitial banner stating that a user is viewing a notebook in standalone mode

### LINKED ISSUE #896: add interstitial banner state for "this notebook is in standalone mode"

hamilton:
Closes #896

### DIFF
--- a/src/components/menu/header-messages.jsx
+++ b/src/components/menu/header-messages.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 
-import { connectionModeIsServer } from '../../tools/server-tools'
+import { connectionModeIsServer, connectionModeIsStandalone } from '../../tools/server-tools'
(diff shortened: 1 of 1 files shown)


src/components/menu/header-messages.jsx
```diff
@@ -16,6 +16,13 @@ export class HeaderMessagesUnconnected extends React.Component {
   render() {
     let content;
     switch (this.props.message) {
+      case 'STANDALONE_MODE':
+        content = (
+          <span>
+            {"You're viewing this notebook in standalone mode."}
```
hamilton: I was receiving an error when I wrote

```jsx
<span>
You're viewing this notebook in standalone mode.
</span>

```

due to a weird html entity issue.

Issue #1082, opened 2018-10-19
openidc-based internal server not setting csrf headers


wlach:
API operations in our internal environment seem not to have csrf headers set, so they're getting access denied errors. Need to figure out why this is happening.

wlach:
`document.cookie`, from which we would get the csrf token, is set to "" on the internal server. I think some middleware is not working as we'd expect it to.

PR #1083, opened 2018-10-19
Remove custom whitenoise middleware

wlach:
This was initially borrowed from treeherder because I thought it had some extra
features that would be useful, but it turns out we don't need any of it.


wlach:
@robotblake requesting review more as an fyi than anything else, this change also contains some simplification of the middleware settings which brings us closer to how [experimenter implemented things](https://github.com/mozilla/experimenter/blob/master/app/experimenter/settings.py). I'm going to go ahead and merge this to help debug #1082

PR #1084, opened 2018-10-19
Ensure csrf cookies are set for all server views

wlach:
The server-side views frequently needs to make write requests via ajax, so be
sure to set the csrf cookie so that these operations succeed.


wlach:
I suspect this is going to fix things, so going to push to confirm that hypothesis.

PR #1086, opened 2018-10-20
disables menu item for 'new notebook' if in standalone mode

### LINKED ISSUE #901: all External link tasks needs a fall-back URL or should be disabled when in standalone mode
right now, selecting the "new notebook" menu breaks if not connected to the server -- we should make sure that UI items have a fallback, or are disabled+greyed out when not connected to a server.

hamilton:
Fixes #901.


src/components/menu/editor-toolbar-menu.jsx
```diff
@@ -5,17 +5,19 @@ import PropTypes from 'prop-types';
...
   static propTypes = {
+    isServer: PropTypes.bool.isRequired,
     isAuthenticated: PropTypes.bool.isRequired,
   }
 
   render() {
     return (
       <NotebookIconMenu>
-        <NotebookMenuItem task={tasks.newNotebook} />
+        {this.props.isServer ? <NotebookMenuItem task={tasks.newNotebook} /> : undefined }
```
wlach: `this.props.isServer && <NotebookMenuItem task={tasks.newNotebook} />` is a little more concise and has the same effect

Issue #1077, opened 2018-10-18
need a default avatar if there isn't one


hamilton:
Depending on the auth layer, we may or may not have avatars / paths to the avatar available. If we do not have an image, we should have a default avatar of some sort.

hamilton:
Great! Thanks @wlach

wlach:
I'm going to solve this by using [gravatar](https://gravatar.com) when github login is not enabled.

PR #1094, opened 2018-10-23
Add gravatar support for when not using github social auth (fixes: #1077)

wlach:
This revealed some flaws in the way we were handling authentication
in the unit tests (i.e. we would incorrectly use github auth if the
user had configured that for their dev server), which are also addressed
here.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,6 +14,7 @@ django-rest-framework = "*"
 faker = "*"
 mock = "*"
 dockerflow = "*"
+pytest-env = "*"
 
 [requires]
 python_version = "3.7"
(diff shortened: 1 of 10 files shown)


hamilton:
@wlach this looks good to me, but do you have a suggested way to test this beyond the tests?

wlach:
@hamilton not really :/ I suppose you could spin up a development instance without github, and then manually create a user or two with various email addresses, and see what the result would be. I'm pretty confident that this will work for our internal deployment given that the tests pass.

hamilton:
Sounds good. Feel free to merge if you feel confident. We're still in a low-risk setting right now.

Issue #990, opened 2018-09-29
As a user, I want to upload files (data, images, scripts) to the server to be used in my notebooks


hamilton:
I want:

- [x] an easy way to upload files from a server page and / or from a notebook
- [x] I want to be able to load these files into my notebook in such a way that appears as if I'm in a directory (for instance, something like `fetch('/files/dataset.csv')`). The exact API is up for debate, but I want running this notebook on an iodide server or a web server to feel seamless.
- [x] I want a way to delete and update files

hamilton:
Closed with #1115

PR #1096, opened 2018-10-23
ask if user wants to replace file upon duplicate file upload

### LINKED ISSUE #1049: As a user, I want to be warned if a file upload wil replace a pre-existing file in this context
Follow-on from #986. Right now, the UI silently fails if a user attempts to upload a file with a name that already exists. The UI should prompt the user through a modal to replace the file or cancel.

hamilton:
This PR fixes #1049, and is part of the story in #990's UX

### DIFF
--- a/src/server/components/notebook-actions-menu.jsx
+++ b/src/server/components/notebook-actions-menu.jsx
@@ -4,29 +4,7 @@ import Menu from './menu'
 import MenuItem from './menu-item'
 import MenuDivider from './menu-divider'
 import DeleteModal from './delete-modal'
-import fetchWithCSRFToken from '../../shared/fetch-with-csrf-token'
-
-function uploadFile(notebookID, successCallback = () => {}) {
-  const filePicker = document.createElement('input')
-  filePicker.type = 'file'
-  filePicker.id = 'file-picker'
-  filePicker.name = 'files[]'
-  filePicker.click();
-  filePicker.addEventListener('change', (evt) => {
-    const file = evt.target.files[0];
-    const formData = new FormData();
-    formData.append('metadata', JSON.stringify({
-      filename: file.name,
-      notebook_id: notebookID,
-    }))
-    formData.append('file', file)
-    fetchWithCSRFToken('/api/v1/files/', {
-      body: formData,
-      method: 'POST',
-    }).then(output => output.json())
-      .then(output => successCallback(output))
-  })
-}
+import uploadFile from '../../shared/upload-file'
 
 export default class NotebookActionsMenu extends React.Component {
   constructor(props) {
--- a/src/server/components/notebook-actions-menu.jsx
+++ b/src/server/components/notebook-actions-menu.jsx
@@ -5,7 +5,7 @@ import MenuItem from './menu-item'
 import MenuDivider from './menu-divider'
 import DeleteModal from './delete-modal'
 import UploadModal from './upload-modal'
-import { selectFile, uploadFile } from '../../shared/upload-file'
+import { selectFileAndFormatMetadata, uploadFile, updateFile } from '../../shared/upload-file'
 
 export default class NotebookActionsMenu extends React.Component {
   constructor(props) {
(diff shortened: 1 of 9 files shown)


src/server/components/notebook-actions-menu.jsx
```diff
@@ -91,6 +108,18 @@ export default class NotebookActionsMenu extends React.Component {
           elementID={this.props.notebookID}
           url={`/api/v1/notebooks/${this.props.notebookID}/`}
         />
+        <UploadModal
+          visible={this.state.uploadFileConfirmationVisible}
+          onClose={this.hideUploadFileConfirmationModal}
+          onCancel={this.hideUploadFileConfirmationModal}
+          onUpload={() => {
+            console.debug(this.state.currentFile)
```
wlach: Probably don't want to keep this in.

src/server/pages/revisions-page.jsx
```diff
@@ -70,8 +70,10 @@ export default class RevisionsPage extends React.Component {
   }
 
   onUploadFile(newFileInfo) {
-    const { files } = this.state
+    let { files } = this.state
+    files = files.filter(f => f.filename !== newFileInfo.filename)
```
wlach: I think it's preferable not to modify the state directly (via `let`). I would do this instead:

```js
let files = this.state.files.filter(...)
...
this.setState({ files })
```

src/server/components/upload-modal.jsx
```diff
@@ -0,0 +1,50 @@
...
+    return (
+      <Modal visible={this.props.visible} onClose={this.props.onClose}>
+        <ModalTitle>{`replace notebook file  "${(this.props.oldFile && this.props.oldFile.filename) || ''}"?`}</ModalTitle>
+        <ModalContent>This action will replace the file associated
+          with this notebook, and cannot be undone.
+          Any other files with the same name associated
+          with other notebooks will not be affected.
+        </ModalContent>
+        <ModalCall>
+          <TextButton onClick={this.props.onCancel || this.props.onClose}>cancel
```
wlach: Should this text be capitalized?

src/server/components/upload-modal.jsx
```diff
@@ -0,0 +1,50 @@
...
+  return fetchWithCSRFTokenAndJSONContent(`/api/v1/files/${fileID}`, { method: 'DELETE' })
+}
+
+export default class UploadModal extends React.Component {
+  constructor(props) {
+    super(props)
+    this.deleteAndReuploadFile = this.deleteAndReuploadFile.bind(this)
+  }
+
+  deleteAndReuploadFile() {
```
wlach: I think this is not quite the right way of doing this. I think we should just upload the new file as a `PUT` request on the resource. That's just one step as opposed to two. Am I missing something?

hamilton: From my notes, it looked like we had agreed on this approach last week, which was why I went with this. My notes might have been off, since it seems obvious that `PUT` at this point is the right approach at this step, not `DELETE` + `POST`.

src/server/components/upload-modal.jsx
hamilton: I suppose, but they're all cap by default

src/server/components/upload-modal.jsx
wlach: Interesting, I wonder what I was thinking?

hamilton: I think it was just a misunderstanding on my part. If we know a file already exists on the server and a user assents to replacing it with the selected file, `PUT` seems like obviously the right approach. The UX layer that wraps the request was what we agreed on, which primarily takes care of the conflation issue (vs overloading `PUT` to always `POST` if the resource isn't there, or something like that).

hamilton: this all said, I'll switch to `PUT` for the modal resolution part.

hamilton:
@wlach deployed and tested, works as expected. Thanks for the review!

PR #1106, opened 2018-10-25
expose kernel state (idle, loading, busy) in toolbar to give user high-level state information about what's happening to evaluation

### LINKED ISSUE #1053: as a user, I expect some feedback when some sort of async process is running
This is based on a comment @mdboom made in a PR:

> As an aside, one thing we could easily do given this refactor is: post the input code to the console pane immediately and show a spinner until cell execution is done, and then show the output. Right now (and previous to this change), nothing goes into the history until execution is done.

I know we've talked about this before, but this feels like a pretty serious UX issue that we know has already caused genuine confusion, and the fix is quite straightforward.

hamilton:
kernel

Fixes #1053. It:

- [x] adds `kernelState` explicitly into the mirrored state, which is an enum with the following values:
  - `KERNEL_LOADING` - the kernel (eval frame) is loading.
  - `KERNEL_ERROR` - a kernel-level error has broken the eval frame in some way (unimplemented as of WIP filing of PR) or an uncaught error occurs
  - `KERNEL_LOAD_ERROR` - the kernel didn't load. Could be because the resource wasn't there, or something else that prevents eval frame initialization. This is achieved through polling - the editor waits _k_ seconds and then reports the error.
  - `KERNEL_IDLE` - the kernel is idling, awaiting further messages.
  - `KERNEL_BUSY` - the kernel is busy. This busy flag occurs when 1.) there is some long-running computation that might potentially be thread-blocking (which is freed up in the editor UI on async iframe browsers like Chrome) or any computation that utilizes further evaluation blocking with the iodide api such as `iodide.evalQueue.await`. Any async processes not explicitly using this API (and thus aren't blocking any further evaluation) will not set `kernelState` to `KERNEL_BUSY`.
- [x] implements a new `KernelState` component, which subsequently reads in the state and updates the icons and tooltip.

Next steps before this PR is ready:
- [x] write tests accordingly _note: testing on the eval frame is unclear at the moment because we don't have an integrated test scaffold for some of these message-passing state changes (#1110). There are tests now for the editor side, however._
- ~[ ] keep spinners from flickering between chunked evals~ _note: going to `wontfix` this for now, and file an issue for later_
- [x] wire up `KERNEL_ERROR` to fire when `window.onerror` is trigged in eval frame / through polling (if the whole iframe blows up somehow)
- [-] ~push input code to console no matter the run state as piece of more immediate feedback~ this feels blocked by how we're pushing to the console. Implementing #1107 might fix this in a way that future-proofs the console.


hamilton:
@wlach @bcolloran I think this is ready for a review. Let me know if this feels like a worthwhile approach to this problem.

src/state-schemas/mirrored-state-schema.js
```diff
@@ -143,5 +143,10 @@ export const mirroredStateProperties = {
     type: 'integer',
     default: undefined,
   },
+  kernelState: {
```
bcolloran: can move to editor schema

src/eval-frame/actions/actions.js
```diff
@@ -133,6 +134,10 @@ export function consoleHistoryStepBack(consoleCursorDelta) {
   }
 }
 
+export function setKernelState(kernelState) {
```
bcolloran: move this whole function into port-to-editor

src/state-schemas/mirrored-state-schema.js
hamilton: Ah yep, good catch. Halfway through this PR I believe I moved the eval frame kernel setting to just a message pass, so in hindsight this def belongs only in editor.

hamilton:
@bcolloran thanks for the review!

hamilton:
@bcolloran this is ready for another review, I believe.

hamilton:
I'll need to rebase this one due to merging #1115 before this is ready for another review.

hamilton:
@bcolloran yes, I think if you run the resource cell there, you'll see in the console it says "module not defined", which is an outside-of-eval error that isn't surfaced in the console. This error doesn't happen on run-all, but only when you run the resource cell on its own. In other words, the kernel stuff is treating this as an uncaught exception outside of the scope of `window.eval`.

Wondering what the right course of action is on this. Clearly there is a non-eval error that isn't being caught, but having only an exclamation point there doesn't really give much info. The resource cell probably should get deprecated, and it clearly doesn't have good enough error catching at any rate.

Let me know how you think I should proceed on this issue.

hamilton:
And for clarification, I've rebased this PR and fixed the merge conflicts.

hamilton:
Thanks for the thorough review - this is exactly what I was hoping for.

I'll try to go through these issues 1 by 1:

a. **when does `KERNEL_ERROR` fire?** As of right now, the only two places it _should_ fire is (1) when we've lost connection to the eval frame, and (2) when there's an uncaught exception. You're correct that in the language plugin load, that should be `KERNEL_IDLE`, not `KERNEL_ERROR` (probably a leftover from when I originally was going to try to halt further evaluation on this PR, but this felt not quite in-scope for this one since I realized it's a PR's-worth of work in and of itself). Regarding (2), this happens with a `window.onerror` definition in `port-to-editor`. Technically speaking, the resource cell is throwing an uncaught error. In this particular case, it does not affect further evaluation, but we're not properly catching the fact that the script itself fails. I think there's a larger question of how we want to actually gracefully catch uncaught exceptions, and I have ideas on how to approach that which isn't related to this (see #1146 for a particularly important example). So for now, I'll kill (2), which means the only way `KERNEL_ERROR` swill get caught is when for some reason the eval frame isn't pinging back.

b. **ok, but we still need to deal with that big false positive - threadblocking computation prevents pinging home.** Great point! This is a trickier problem. If you are running something that blocks the main thread for three seconds, you'll get an error because the main thread can't ping home. The event loop is locked. (1) One option is to up the time to something like 15-20 seconds. That works for cases where 15-20 seconds is enough. We'll likely get some false positives in that scenario as well for really hairy naive computations. (2) Another possibility would be to simply turn off the health ping when we have `KERNEL_BUSY` so the threadblocking can do what it wants. This seems like a bad idea, since there may be some totally big app-busting error thrown once the cell starts evaluating in some way, and we'll be stuck in `KERNEL_BUSY` forever. This seems like a rare false negative. I think (1) is probably the right approach for now, barring something more clever. Another way to deal with this is if we are in a `KERNEL_BUSY` state, up the timeout threshold to a minute or something like that, and then fire a (?) or something like that.

One way to actually unify all these cases is to just throw out `KERNEL_ERROR` altogether. We don't know if there's an error. What we do know, however, is sometimes `KERNEL_CONNECTION_LOST` is a possibility.

hamilton:
This PR is ready for another review, I think. Going to forgo moving away from `KERNEL_ERROR` but just make it a lot more permissive (10 seconds without a ping causes a `KERNEL_ERROR` and that's pretty much it).

src/eval-frame/actions/actions.js
```diff
@@ -275,6 +277,13 @@ function evaluateCSSCell(cell) {
   }
 }
 
+function sendKernelStateAndEvaluate(pr, dispatch, cell, evaluator) {
```
bcolloran: what is "pr"? i'm sure i'm missing something obvious, but literally can't figure out what it's short for, which i guess is a sign we should use a more descriptive name here

src/eval-frame/actions/language-actions.js
```diff
@@ -85,20 +87,21 @@ function loadLanguagePlugin(pluginData, historyId, cell, dispatch) {
 
 export function evaluateLanguagePluginCell(cell) {
   return (dispatch) => {
+    // sendKernelStateToEditor('KERNEL_BUSY')
```
bcolloran: delete?

src/eval-frame/actions/language-actions.js
```diff
@@ -85,20 +87,21 @@ function loadLanguagePlugin(pluginData, historyId, cell, dispatch) {
...
       { historyId, historyType: 'CELL_EVAL_INFO' },
     ))
-
     let pluginData
     try {
       pluginData = JSON.parse(cell.content)
     } catch (err) {
       dispatch(updateCellProperties(cell.id, { evalStatus: 'ERROR', rendered: true }))
       dispatch(updateValueInHistory(historyId, `plugin definition failed to parse:\n${err.message}`))
+      // sendKernelStateToEditor('KERNEL_IDLE')
```
bcolloran: delete?

src/eval-frame/actions/language-actions.js
```diff
@@ -18,12 +18,12 @@ function loadLanguagePlugin(pluginData, historyId, cell, dispatch) {
   let evalStatus
   let languagePluginPromise
   const rendered = true
-
   if (pluginData.url === undefined) {
     value = 'plugin definition missing "url"'
     evalStatus = 'ERROR'
     dispatch(updateCellProperties(cell.id, { evalStatus, rendered }))
     dispatch(updateValueInHistory(historyId, value))
+    sendKernelStateToEditor('KERNEL_ERROR')
```
bcolloran: should this be a "kernel_idle"?

src/eval-frame/actions/actions.js
bcolloran: oh it's "promise". i know it should be obvious, but i missed it for a long time, so above still stands... let's help the slow and sleep deprived :-)

bcolloran:
@hamilton hey sorry, i lost the thread on this one -- from my last comment, i thought it was close to merging pending a few small changes. But had we decided to wait on this pending the JSMD editing?

bcolloran:
discussed with @hamilton -- we're going to reimplement this on top of jsmd editing. closing this version

PR #1114, opened 2018-10-30
Remove git-rev-sync and dependencies

### LINKED ISSUE #1109: removing dead packages
running `depcheck --specials=webpack,eslint,babel` returns:

```
Unused dependencies
* codemirror-no-newlines
* re-resizable
* react-spring
Unused devDependencies
* babel-core
* babel-jest
* babel-loader
* css-loader
* eslint-loader
* eslint-plugin-emotion
* file-loader
* git-rev-sync
* jest
* jest-cli
* jest-localstorage-mock
* raw-loader
* react-test-renderer
* redux-mock-store
* style-loader
* webpack-cli
* webpack-dev-server
* webpack-merge
```

i haven't checked all of these, but doing a text search thru the whole codebase returns zero hits (other than in package.json and package-lock.json) for a number of these. I would love everyone's help double checking these, double checking that removing them doesn't cause `run start` or `run start-and-serve` to fail, and then filing PRs to clean this stuff out (as many as you with per PR -- it could be one big PR if you wish, or you can pick of the ones you feel confident about).--

i'm pretty sure were not using any of: codemirror-no-newlines,  re-resizable, react-spring

a number of the dev deps are false positives, but a bunch of these are actually not used.

@hamilton @wlach -- i think we're not using react-test-renderer (i think we use enzyme instead). can you help me double check?

@wlach @mdboom -- i think git-rev-sync was only used by our travis + gh pages deploy situation?

i think we're not using raw-loader or style-loader

i think we're not using  redux-mock-store or jest-localstorage-mock

if you see any other deps that we're definitely not please call them out or file a PR. when in doubt, we can leave it in.

wlach:
Partially fixes #1109.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4961,17 +4961,6 @@
         "assert-plus": "^1.0.0"
       }
     },
-    "git-rev-sync": {
-      "version": "1.12.0",
-      "resolved": "https://registry.npmjs.org/git-rev-sync/-/git-rev-sync-1.12.0.tgz",
-      "integrity": "sha1-RGhAbH5sO6TPRYeZnhrbKNnRr1U=",
-      "dev": true,
-      "requires": {
-        "escape-string-regexp": "1.0.5",
-        "graceful-fs": "4.1.11",
-        "shelljs": "0.7.7"
-      }
-    },
     "glob": {
       "version": "7.1.3",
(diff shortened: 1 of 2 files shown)


bcolloran:
thanks @wlach !

Issue #1107, opened 2018-10-25
separate code evaluation inputs from outputs in console


hamilton:
This is a bit of a follow-on for #1106, and maybe blocks it.

When I run some async code, I want to separate the posting of code input from outputs when printing to the console, so that we see that the code has hit the console even if the code creates a long-running or async process. Right now, input and output are tightly coupled, which creates a number of UX limitations. Specifically, there's no clear way of giving users the immediate feedback that the code has hit the console. It looks something like this:

```javascript
function evaluateCodeCell() {
  ... 
  return ensureLanguageAvailable(...)
    .then(runCodeWithLanguage)
    .then(addInputAndOutputToHistory)
}

```

I can in theory create the history item before running the code, then update with an action. 

```javascript
function evaluateCodeCell() {
  ... 
  return ensureLanguageAvailable(...)
    .then(createCodeInputAndOutputInHistory)
    .then(runCodeWithLanguage)
    .then(updateOutputInHistoryWithNewValue)
}

```

But that would require some complexities if the language hasn't loaded, since we need to tell the user that we have to load the language plugin before running the code. Which means the output ends up something like this:

<img width="533" alt="screen shot 2018-10-25 at 3 40 14 pm" src="https://user-images.githubusercontent.com/95735/47534467-517cc880-d86c-11e8-92e1-4ee62fd631c3.png">

We could probably make some changes to handle this particular use-case more gracefully with our existing setup, but it really side-steps a more flexible system and rules out any sort of future `console.log` interception. Between the code input print and the code output print, there might actually be all sorts of messages that pop up. What I really want is to just be able to print to console the fragments I want during the moment I want them without thinking about inputs and outputs quite so explicitly. Something like this:

```javascript
function evaluateCodeCell() {
  ... 
  return ensureLanguageAvailable(...)
    .then(addCodeInputToHistory)
    .then(runCodeWithLanguage) // there could be more history pieces that go to console here, or console.log down the line
    .then(addOutputToHistory)
}

```

I think the solution would be simply to implement the following history types based on the pieces of the ones we have now:

`EVAL_INPUT` - fires first in code chunk evaluation.
`EVAL_OUTPUT` - fires last in code chunk evaluation, uses the `ValueRenderer`.

and remove the code-and-output history types in favor of this workflow.

@mdboom @wlach @bcolloran curious to see what y'all think of this idea.

mdboom:
[...]

Issue #1054, opened 2018-10-10
actions have few / no real test coverage


hamilton:
We have no method or strategy around writing tests for our actions, which I see as a pretty major code quality problem on the client. This meta issue is meant for feedback & further issue decomposition.

The seismic refactors we've undergone is probably somewhat the cause (at least the cause of not pursuing test coverage here), but at this point, given that we are not likely to refactor the entire code base any further for the forseeable future, we can do better here.

PR #1115, opened 2018-10-30
implement a "fetch" cell for external resources (data, scripts, stylesheets)

bcolloran:
semi-working WIP fetch cell:

- can load JS files, but doesn't handle errors

still needs:
- load files (blob/text/json) to a variable
- load css
- handle errors of various sorts
- fetch local path files via editor


src/eval-frame/actions/fetch-cell-eval-actions.js
```diff
@@ -0,0 +1,164 @@
...
+      .map(syntaxErrorToString)
+
+    if (syntaxErrors.length > 0) {
+      dispatch(appendToEvalHistory(
+        cell.id,
+        cell.content,
+        syntaxErrors,
+        { historyId, historyType: 'FETCH_CELL_INFO' },
+      ))
+      return Promise.resolve()
```
bcolloran: this should be a reject()

hamilton:
I'm taking on this PR in the interest of time. At this point I believe everything should be covered:

- load javascript files as scripts (`js` fetch type)
- load and apply css files (`css` fetch type)
- load `text`/ `json` / `blob` types
- for all of the above, load either from an external endpoint (like a cdn) or from the iodide server (something like `files/<filename>`) through message passing

I'd like to do one last pass through this PR before requesting a review, since there is as lot of functionality implemented and I want to be sure I've covered all the edge cases.

### issues / room for improvement

1. Because we cannot incrementally print to the console (see #1107), I have to batch-update based on the current history item functionality. This means that even though everything is requested asynchronously, the update has to happen once everything has resolved one way or another. This seems very backward to me, and is a potential usability issue. If we fix #1107 we can fix this usability issue.
2. we should probably be keeping track of what libraries, files, etc. are loaded in the Workspace, more than just the declared variables.
3. this likely a trivial case for implementing our own CodeMirror syntax highlighting, which would drastically improve readability of these cell types.

hamilton:
@wlach @bcolloran this is ready for a review. @bcolloran because this was your PR originally it doesn't look like I can add you as a reviewer.

hamilton:
Here's some tests you might be able to run to check if it does everything. Some of these will fail unless you upload something to the server.

```
// text, json, blob: PASS CONDITION (remote)
text: test=https://cdnjs.cloudflare.com/ajax/libs/animate.css/3.7.0/animate.css
json: libraries= https://api.cdnjs.com/libraries
blob: gritty=https://media.giphy.com/media/MV1oM1koBZRvfp8Hvw/giphy.gif

// text, json, blob: FAIL CONDITION (remote)
text: animateFail=https://cdnjsasdsdfsdf.cloudflare.com/ajax/libs/animate.css/3.7.0/animate.css
blob: grittyFail=https://cdnsdfdsfjs.cloudflare.com/ajax/libs/animate.css/3.7.0/animate.cssssss
json: libFail=https://api.cdsdfnjs.com/librariessss

// text, json, blob: PASS CONDITION (if files uploaded to server)
text: testLocal= files/multiTimeline.csv // defines text: 
json: librariesLocal= files/libraries.json
blob: grittyLocal= files/gritty.mp4

// text, json, blob: FAIL CONDITION (if files uploaded to server)
text: testLocFail= files/multiTimeline.csvvvvv
json: libLocFail= files/libraries.jsonssss
blob: grittyLocFail= files/gritty.mp444444

// js: PASS CONDITIONS (server / remote)
js: https://cdnjs.cloudflare.com/ajax/libs/d3/5.7.0/d3.min.js
js: files/d3.min.js

// js: FAIL CONDITIONS (server / remote)
js: https://cccccccccdnjs.cloudflare.com/ajax/libs/d3/5.7.0/d3.min.js
js: filess/d3.min.js

// FAIL syntax errors in variable declaration
text: 9test=https://cdnjs.cloudflare.com/ajax/libs/animate.css/3.7.0/animate.css
json: 9libraries= https://api.cdnjs.com/libraries
blob: 9gritty=https://media.giphy.com/media/MV1oM1koBZRvfp8Hvw/giphy.gif

// FAIL syntax errors in fetch type should be caught.
1text: testLocal= files/multiTimeline.csv
jssson: librariesLocal= files/libraries.json
blobb: grittyLocal= files/gritty.mp4

// FAIL missing fetch type
testLocal= files/multiTimeline.csv
librariesLocal= files/libraries.json
grittyLocal= files/gritty.mp4

// css - PASS / FAIL conditions (server & remote)
css: https://gist.githubusercontent.com/hamilton/3afe20511547cb4f06ce5487a70702f1/raw/60df36c1f0c8109f5c73db6452012a08f69f7514/test.css
css: files/test-style.css

css: https://gist.111githubusercontent.com/hamilton/3afe20511547cb4f06ce5487a70702f1/raw/60df36c1f0c8109f5c73db6452012a08f69f7514/test.css
css: files11111/test-style.css
```

src/eval-frame/index.jsx
```diff
@@ -25,6 +25,7 @@ import { iodide } from './iodide-api/api'
 import './port-to-editor'
 
 window.iodide = iodide
+window.FETCH_RESOLVERS = {}
```
wlach: Do we actually need to add a property to the window? Could we not define this object in a file and then import it elsewhere?

wlach:
Rebased this branch on master for testing on heroku

src/eval-frame/index.jsx
hamilton: Yep, it would be easy to have something that actually just closes over a variable in some way, along with some functions that fetches the `resolve` and `reject` as well as deletes.

hamilton: fwiw we use this pattern elsewhere, but agreed that maybe we should move to something that doesn't attach to `window`

hamilton:
We're suspecting that a rebase may have caused some issue, so I'll investigate and fix any issues. Thanks for the review @wlach.

wlach:
@hamilton seeing the same problem when deploying this to the heroku server as I did last friday, both console and report views seem to be empty. did you push all your changes?

hamilton:
Hmm, I believe I did, but will check again in a few hours.

src/components/cells/cell-editor.jsx
```diff
@@ -15,12 +15,23 @@ import 'codemirror/addon/display/autorefresh'
...
 import 'codemirror/addon/hint/show-hint'
 import 'codemirror/addon/hint/javascript-hint'
+import 'codemirror/addon/mode/simple'
 import '../../codemirror-keymap-sublime'
 
 import { getCellById } from '../../tools/notebook-utils'
 import * as actions from '../../actions/actions'
 import { postMessageToEvalFrame } from '../../port-to-eval-frame'
 
+CodeMirror.defineSimpleMode('fetch', {
```
bcolloran: this is great! let's move it to a separate file tho

src/eval-frame/actions/__tests__/fetch-eval-actions.test.js
```diff
@@ -0,0 +1,5 @@
+describe('stub', () => {
```
bcolloran: want to add some tests around "fetch-cell-eval-actions.js"? i put this stub in place, but for the stuff id had written so far i wasn't sure how to test it (fetch and promise stuff). if you know how to write tests around that, it would be awesome (it looks like some of these functions are pure enough that it might not be too bad?)

if not, let's delete this file.

src/components/cells/cell-editor.jsx
hamilton: yeah, was just thinking this earlier - will do.

src/eval-frame/actions/__tests__/fetch-eval-actions.test.js
hamilton: Going to delete this file for now, I think. It's not immediately obvious how to test this one action as-is.

hamilton: That said, I do think it's worth getting the actions better tested (see #1054) I just don't know if this is the PR to begin tackling that.

src/eval-frame/actions/fetch-cell-parser.js
```diff
@@ -0,0 +1,109 @@
...
+
+    // first, strip out comment from the end of line (if it exists)
+    const fetchCommand = fetchContent.trim().split(' //')[0].trim()
+    switch (fetchType) {
+      case 'text':
+        return Object.assign({}, { fetchType }, parseAssignmentCommand(fetchCommand))
+      case 'json':
+        return Object.assign({}, { fetchType }, parseAssignmentCommand(fetchCommand))
+      case 'blob':
+        return Object.assign({}, { fetchType }, parseAssignmentCommand(fetchCommand))
```
wlach: I think you could shorten this to:

```js
case 'text':
case 'json':
case 'blob':
   return Object.assign({}, { fetchType }, parseAssignmentCommand(fetchCommand))
```

Likewise for the js/css cases below.

src/components/cells/cell-menu-container.js
```diff
@@ -109,6 +109,8 @@ function mapStateToProps(state, ownProps) {
     label = 'resource'
   } else if (cell.cellType === 'raw') {
     label = 'raw'
+  } else if (cell.cellType === 'fetch') {
+    label = 'fetch'
```
wlach: Could you simplify/generalize this code by using `cellTypeToJsmdMap` in jsmd-tools? (might be better done in a followup)

bcolloran: nice catch will. i think it actually makes sense to do this now -- easier/cleaner/faster than having another issue+PR hanging over us

src/eval-frame/actions/fetch-cell-parser.js
bcolloran: +1 good catch

src/tools/fetch-tools.js
```diff
@@ -0,0 +1,46 @@
...
+  INVALID_VARIABLE_NAME: 'invalid variable name',
+}
+
+export function syntaxErrorToString(fetchInfo) {
+  return `Syntax error, ${errorTypeToString[fetchInfo.parsed.error]} in:
+      "${fetchInfo.line}"
+  `
+}
+
+export function handleErrors(err) {
```
wlach: Any reason this function needs to be exported?

src/tools/fetch-tools.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function successMessage(fetchInfo) {
+  const ifVarSet = fetchInfo.parsed.varName ? `\n\t(var ${fetchInfo.parsed.varName})` : ''
+  const text = `SUCCESS: ${fetchInfo.parsed.filePath} loaded${ifVarSet}\n\n`
+  return { text, id: fetchInfo.id }
+}
+
+export function errorMessage(fetchInfo, msg) {
+  let firstLine = ''
```
wlach: Probably would do:

```js
const firstLine = (fetchInfo.parsed.filePath) ? `${fetchInfo.parsed.filePath}\n\t` : '';
```

src/components/cells/cell-menu-container.js
hamilton: Sure, haven't really touched this so I'll have to look into what you mean :)

bcolloran:
>     1. Because we cannot incrementally print to the console (see #1107), I have to batch-update based on the current history item functionality. This means that even though everything is requested asynchronously, the update has to happen once everything has resolved one way or another. This seems very backward to me, and is a potential usability issue. If we fix #1107 we can fix this usability issue.

we can update history items incrementally as events land, as we do in the plugin cell evaluation. that was my intent here, though i didn't get to that part. nonetheless, the update *does not* have to wate until everything is resolved. you can add this functionality with updateValueInHistory().

i think it's worth doing that now since you've identified this as a potential usability issue, and since this code is fresh in your mind, yes?

>     2. we should probably be keeping track of what libraries, files, etc. are loaded in the Workspace somehow.

can you elaborate? why do we need to track it?

hamilton:
@bcolloran oh I see, you mean the syntax errors, not the failed fetches.

bcolloran:
> @bcolloran you can't really have that & the async block fetching everything at the same time - there's no way to intercept an already-running fetch call. The best we can do is just surface that error. I figured it'd be better if a user knew all the errors present in the fetch cell, not just the first. Happy to change that if you feel otherwise.

@hamilton -- since the fetch cell parser does a pass over the complete cell contents, you can filter the parse results for syntax errors and surface them before kicking off any fetches

hamilton:
@bcolloran right - so is the preferred behavior to show only the first error, or all the syntax errors?

bcolloran:
i think we should surface all the syntax errors since we can catch them at parse time an halt evaluation before actually kicking off a bunch of potentially expensive fetches, and requiring the user to run a bunch of fetches twice. that was my thinking on it -- fail fast and early, ya know :-)

hamilton:
@bcolloran sounds good, reread your original comment and saw "all" so I'm on it.

hamilton:
@bcolloran addressed the issues you brought up. Let me know if there's anything else here.

bcolloran:
woot!!! everything works as expected, solid code, great PR! thanks @hamilton, and nice work :-)
 :100:

PR #1116, opened 2018-10-30
Add a "root-shell" make command

wlach:


wlach:
@hamilton can you take a look at this? just some simple stuff I wrote while prototyping the server side of autosave, to make installing new python dependencies easier.

Issue #1090, opened 2018-10-23
as a user, I would like a toolbar of some sort on the console


hamilton:
Working with the console over a long period of time becomes pretty difficult. One way we can manage volume would be to:

- have a clear console button (#1015). This should probably _not_ clear the history stored in redux - I can see that still being relevant for forensics even when the console is very long. Could be wrong about that though.
- have a button that lets me hide the code inputs from the console, since these are one of the largest contributors to the console right now.
- have a button that lets me hide the code outputs. Could be different output _types_ you can hide, or console levels (show only info / warning / error / debug / etc)

bcolloran:
after dogfooding this a bit, i think this should be in a menu that lives in the right hand side of the console text entry rather than a toolbar -- the tool bar just eats up a lot of our precious screen real estate for three buttons (and just one ATM)

also, I'm seeing rendering errors with the clear button on extremely-alpha and localhost with both FF and Chrome. 

![image](https://user-images.githubusercontent.com/2380975/48035454-e7400f80-e118-11e8-958d-94e905b3f84e.png)
 i'm going to take a pass at all this

hamilton:
Let’s talk about this during tomorrow’s standup before going forward with a
change.

bcolloran:
sgtm -- in the meantime, since this feature is not working as expected and we're not sure about the UX we want, and a couple code quality things got past review, i've reverted #1119. we can take another crack after discussing :-)

hamilton:
Sounds good to me. fwiw I believe that issue you're encountering was brought up in that PR, which had a fix & review on my mbp on FF. Clearly I didn't do a good enough code review on this pass - apologies!

Render bugs, on the other hand, are no joke, and I'll do a better job in the future checking on Chrome.

bcolloran:
menu exists, reopened #1015 to add a "clear console" button back in that menu. closing this one

PR #1119, opened 2018-11-01
Add clear history button

### LINKED ISSUE #1015: [Feature Request] add "clear console" button to console menu
Was testing some things on extremely-alpha and I noticed a feature that would be nice to have: an ability to clear the console. I don't know if there are system level commands in iodide, but a `clear` command would be great.

mdboom:
Fixes #1015 and #1090 

image

### DIFF
--- a/src/components/controls/mini-toolbar.jsx
+++ b/src/components/controls/mini-toolbar.jsx
@@ -1,0 +1,26 @@
+import React from 'react'
+import { connect } from 'react-redux'
+
+import Toolbar from '@material-ui/core/Toolbar'
+
+export class MiniToolbarUnconnected extends React.Component {
+  render() {
+    return (
+      <Toolbar
+        style={{ minHeight: '0' }}
+        variant="dense"
+        disableGutters
+        classes={{ root: 'mini-toolbar' }}
+      >
+        {this.props.children}
+      </Toolbar>
+    )
+  }
+}
+
+export function mapStateToProps() {
+  return { }
+}
+
+const MiniToolbar = connect(mapStateToProps)(MiniToolbarUnconnected)
+export default MiniToolbar
--- a/src/components/controls/mini-toolbar.jsx
+++ b/src/components/controls/mini-toolbar.jsx
@@ -7,7 +7,7 @@ export class MiniToolbarUnconnected extends React.Component {
   render() {
     return (
       <Toolbar
-        style={{ minHeight: '0' }}
+        style={{ minHeight: '24px' }}
         variant="dense"
         disableGutters
         classes={{ root: 'mini-toolbar' }}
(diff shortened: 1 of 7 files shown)


hamilton:
I think the position of that tooltip could be place so that it appears in the lower right side of the icon. At the moment, the editor pane is overlapping a tooltip that is actually appearing in the eval frame. I'm not sure how we address this with this current setup, but it does seem like one solution is to just prevent the tooltip from being in the overlapping area to the best of our abilities.

Setting the tooltip placement ([outlined here](https://material-ui.com/demos/tooltips/#positioned-tooltips)) to `bottom-end` might solve the issue.

mdboom:
Oh, CSS is such a joy!  I'll poke at it and let you know if I get stuck.

mdboom:
I think I've fixed the tooltip and shrinking height issues.

hamilton:
LGTM - thanks @mdboom!

src/eval-frame/components/panes/console-toolbar.jsx
```diff
@@ -0,0 +1,54 @@
...
+          >
+            <DeleteButton fontSize="small" />
+          </IconButton>
+        </Tooltip>
+      </MiniToolbar>
+    )
+  }
+}
+
+export function mapStateToProps() {
```
bcolloran: this mapStateToProps is not needed
https://stackoverflow.com/questions/47657365/can-i-mapdispatchtoprops-without-mapstatetoprops-in-redux

src/eval-frame/components/panes/console-toolbar.jsx
```diff
@@ -0,0 +1,54 @@
...
+  return { }
+}
+
+function mapDispatchToProps(dispatch) {
```
bcolloran: preferred method is to just select out the actions you need, as in:
https://github.com/iodide-project/iodide/blob/3d14fb7a32af15e50f2dcddf6b0042a65710c478/src/components/cells/cell-container.jsx#L154

we have this pattern still floating around, but it is legacy

src/components/controls/mini-toolbar.jsx
```diff
@@ -0,0 +1,26 @@
...
+      </Toolbar>
+    )
+  }
+}
+
+export function mapStateToProps() {
+  return { }
+}
+
+const MiniToolbar = connect(mapStateToProps)(MiniToolbarUnconnected)
```
bcolloran: this does not appear to need to be connected to the redux store at all, right? it's just a simple presentational component?

hamilton:
Thanks for the more thorough review @bcolloran. I think I dropped the ball on this PR review + testing more thoroughly on Chrome.

mdboom:
Sorry about this, all.  I still feel like a monkey on a typewriter with React.  Is there a good reference / book?

bcolloran:
No worries @mdboom!! :-) We're here to help you through it. Unfortunately we have a bunch a legacy code from when we were first learning that demonstrates bad patterns, and in many places it's not a great foundation to learn from... so really it's me and hamilton who must apologize to you!

I don't know of any particular books, but i think the official react and redux docs are excellent. I've also found quite a number of great hackernoon and medium articles that have helped a lot when googling around specific topics.

Issue #894, opened 2018-09-13
re-implement localStorage-based autosaving


hamilton:
Our previous localStorage story (discussed in #351) has been deprecated, but as @louisabraham pointed out, there are still some very useful cases where localStorage might make sense (including auto-saving on a non-server instance).

There is probably some interaction around the server instance and a localStorage autosave feature as well, though I imagine that'll probably fall away as we get per-keystroke-chunk saving on by default on the server component.

With this in mind, we should come up with a design for this that solves the issue (a user loses editor progress if they accidentally close a tab or navigate away from the notebook in some settings).

bcolloran:
see also #853, #1018

SimonBiggs:
Not sure if this is helpful or not but I found IndexedDB far better for purposes like this. Also if it helped http://dexie.org is a great wrapper around IndexedDB.

wlach:
Ok, have been hacking on this a bit over the past few days. Here are my ideas on how this should work (have implemented a good chunk):

We store the following structure locally (currently in localStorage, but could be indexdb):

```js
{
   originalCopy: <jsmd>,
   originalSaved: "<date in isoformat>",
   dirtyCopy: <jsmd>,
   dirtySaved: "<date in isoformat>"
}
```

We update the dirty copy every 15 seconds or so if there are changes.

On first load of a notebook, we serialize the original contents and the last modified date to the original values. On any modification, we update the dirty copy with the current state of the jsmd and the save date.

Behaviour will be slightly different depending on which mode the notebook is operating: server-based or standalone.

Server based:
* We will clear the dirty copy and reset the original copy on any save
* On subsequent loads, if there is a dirty copy, then set the content to that, but offer in a banner to let user revert to server-based version (this will need to somehow cooperate with the other banners we have for other things, like the message we show when the user doesn't own the notebook)

Standalone:
* On any modification, warn that the changes are only saved locally. Offer to let user revert to original.

PR #1131, opened 2018-11-06
Add a simple localStorage-based autosave implementation (fixes: #894)

wlach:


wlach:
This only needs one of @hamilton or @bcolloran to review :)

src/tools/autosave.js
```diff
@@ -0,0 +1,99 @@
+import { exportJsmdToString } from './jsmd-tools'
+import { setPreviousAutosave } from '../actions/actions'
+
+function getAutosaveKey(state) {
+  const documentId = (state.notebookInfo.connectionMode === 'STANDALONE') ?
```
hamilton: This is throwing an error on my local version of the branch when I load a notebook from the server, for some reason. I believe you can use`connectionModeIsStandalone(state)` here, at any rate.

src/reducers/notebook-reducer.js
```diff
@@ -95,6 +100,13 @@ const notebookReducer = (state = newNotebook(), action) => {
       )
     }
 
+    case 'REPLACE_NOTEBOOK_CONTENT': {
+      cells = action.cells.map((cell, i) =>
+        Object.assign(newCell(i, cell.cellType), cell))
+
+      return Object.assign({}, state, { cells, title: action.title })
```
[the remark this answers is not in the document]
wlach: Originally I did use newNotebook() here as a template for state, but that caused the UI to blank out. I figured we were only supposed to use that on first load?

src/tools/autosave.js
```diff
@@ -0,0 +1,99 @@
...
+  } else {
+    const autosaveState = getAutosaveState(autosaveKey)
+    saveAutosaveState(autosaveKey, Object.assign(autosaveState, {
+      dirtyCopy: jsmd,
+      dirtySaved: new Date().toISOString(),
+    }))
+  }
+}
+
+function subscribeToAutoSave(store) {
```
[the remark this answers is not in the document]
wlach: makes sense

wlach:
@hamilton I think this is ready for another look

hamilton:
@wlach Great! It works.

Another thing I'm noticing, now that I'm testing it - if I run a notebook in standalone mode, then edit it, then reload the page and click "restore", and then reload again, I lose my changes.

I'm wondering if we only apply the subscribe step of overwriting what's in local storage if there has been an edit in some way. Otherwise, I end up losing my work if I don't immediately apply, or apply the change and then don't do anything other than close.

Otherwise I think this is looking good. Going to do another pass on the code itself now that I'm understanding how it feels under the fingers.

wlach:
@hamilton that makes sense... I believe that it will re-autosave if you make any modification after restoring, but we should probably just keep the "dirty" copy outright. this shouldn't be too hard.

hamilton:
Yes, but just to clarify, I didn't make any modification after restoring - at least not as an intentional user action. I just refreshed.

wlach:
Fixed the issue by not clearing autosave on restore.

src/tools/__tests__/autosave.test.js
```diff
@@ -0,0 +1,55 @@
+import { updateAutosave, getAutosaveState } from '../autosave'
+import { stateFromJsmd } from '../jsmd-tools'
+import { newNotebook, newCell } from '../../editor-state-prototypes'
+
+describe('updateAutosave', () => {
+  function setUp() {
```
hamilton: `beforeEach` is a Jest function that does this by default, like `beforeEach(()=>{something})`.

hamilton: https://jestjs.io/docs/en/setup-teardown

wlach:
Fixed the nit and merged!

hamilton:
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

Issue #1127, opened 2018-11-05
circle configuration does not test local docker-compose environment


wlach:
The circle ci configuration we're using for iodide does not test the local docker-compose environment. It really needs to, otherwise we can't know if a PR will break stuff (like #1126 did today)

PR #1139, opened 2018-11-07
Build webpack in test-javascript job

wlach:
This is an incomplete fix for #1127, intended to make sure we don't inadvertently merge PRs which break the entire webpack build (as happened yesterday with #947).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,6 +46,7 @@ jobs:
       - run: npm install
       - *save_javascript_cache
       - run: npm run lint
+      - run: npm run build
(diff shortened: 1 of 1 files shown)


wlach:
@bcolloran @hamilton could one of you review and merge this? I have a more proper fix in progress which splits apart the build and deploy steps for the docker container (so we can run the build step in ci), but I should probably get @robotblake to review that before we land. meanwhile this should stop obviously broken stuff from passing ci.

hamilton:
@wlach happy to take this on

PR #1140, opened 2018-11-07
Use full docker-compose environment for integration testing (fixes: #1127)

wlach:


wlach:
@robotblake Hey, so this PR moves some things around so we're testing with something closer to our local dev environment, by factoring out the build of the docker image from its deploy, and then modifying the python/server tests to use that image. Could you have a look?

wlach:
Worth mentioning that much of this was based on the excellent work that @whd did when setting up circle ci for missioncontrol: 

https://github.com/mozilla/missioncontrol/blob/master/.circleci/config.yml

robotblake:
Looks reasonable to me and the tests seem to pass! :shipit:

PR #1142, opened 2018-11-08
on notebook load, compare autosave.originalCopy with autosave.dirtyCopy

### LINKED ISSUE #1141: autosave: only add dirty copy when code cells have edits
I'm 0/2 for useful reviews recently, so I'm going to try to undo the damage and find all the edge cases on the new autosave functionality.

- `/new` on the server probably shouldn't have autosaving (possibly)
- in a notebook, if I run the cells, then refresh, the notebook believes there's a dirty copy there. Probably shouldn't.
- if I leave an empty notebook open for a bit, then refresh, it assumes there is a change if I try `cmd+shift+r`, but not `cmd+r`.
- refreshing in general seems to walk through some implicit state machine that's not what the user expects.

hamilton:
fixes #1141 - now the autosave dirty & original copies are checked for equality before alerting the user there is a more recent version.

@wlach, while working through this case, I'm wondering what the utility of `autosave.originalCopy` is when we also happen to have the actual generated jsmd to pull from. Could we not actually just compare the `dirtyCopy` to that on pageload?

### DIFF
--- a/src/tools/autosave.js
+++ b/src/tools/autosave.js
@@ -25,7 +25,9 @@ function saveAutosaveState(autosaveKey, autosaveState) {
 function checkForAutosave(store) {
   const state = store.getState()
   const autosaveState = getAutosaveState(state)
-  if (autosaveState && autosaveState.dirtyCopy) {
+  if (autosaveState &&
+      autosaveState.dirtyCopy &&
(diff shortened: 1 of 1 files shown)


wlach:
This was an obvious error. Thanks for doing the followup!

PR #1152, opened 2018-11-13
Direct jsmd editing

bcolloran:
there is a lot still to be done here and a lot that doesn't yet work, but the basics are roughed in.

highlights:
- adds `jsmd` to state, which is the text being edited
- on updating jsmd, jsmd is parsed into chunks
- on updating jsmd, MD and HTML chunks are updated in the report view immediately
- evaling (ctrl+enter) sends the chunk containing the cursor to be evaled
- codemirror modes work (not all are wired up yet, but it's easy to do)

#### things that occur to me off the top of my head still to be fixed (not exhaustive)
- run all
    - ... and b/c run all doesnt work, neither does starting in report view
    - stuff around async/eval queue needs to be looked at
- not all eval types work yet
    - css cells eval
    - 'external dependencies' (but this should be deprecated anyway?)
    - 'plugin' cells
- evaling (ctrl+enter) of highlighted text 
- eval and advance cursor to next chunk (shift+enter)
- python CM mode
- js autocomplete
- the jsmd "meta" is kind of deprecated, so we need a new solution for the notebook title (could just be a server feature?)
- all the refactorings/simplifications
    - tearing out cells!!! :-D
    - need to think about COMMAND_MODE, do we still need it?
    - etc?
- ... etc ...

i think (with a couple exceptions) most of these things are just a matter of hooking up the wiring, and that the hard parts are mostly solved.


wlach:
@bcolloran sounds really nifty! note that this is currently failing tests.

src/actions/actions.js
```diff
@@ -218,6 +242,36 @@ export function updateCellProperties(cellId, updatedProperties) {
...
+
+    let actionObj
+
+    if (!doc.somethingSelected()) {
+      const { line } = doc.getCursor()
+      const [activeChunk] = getState()
+        .jsmdChunks
+        .filter(c => c.startLine <= line && line <= c.endLine)
+
+      actionObj = Object.assign({
```
hamilton: I think we'll be able to better unit this block & the unimplemented one below it.

src/actions/jsmd-parser.js
```diff
@@ -0,0 +1,51 @@
...
+  let currentChunkStartLine = 0
+
+  for (const [i, line] of jsmdLines.entries()) {
+    if (line.slice(0, 2) === '%%') {
+      // if there is any match, a new chunk has started
+      // push the current chunk (unless it's on line 0), then reset
+      if (i !== 0) {
+        // DON'T push a chunk if we're only on line 0
+        chunks.push({
+          cellContent: currentChunkLines.join('\n'),
```
hamilton: We should be using the term `chunk` not `cell` here.

src/components/jsmd-editor.jsx
```diff
@@ -0,0 +1,189 @@
...
+    lineNumbers: true,
+    keyMap: 'sublime',
+    comment: true,
+    // readOnly: cell.highlighted ? 'nocursor' : false,
+  }
+
+  if (state.wrapEditors === true) { editorOptions.lineWrapping = true }
+
+  return {
+    thisCellBeingEdited: true,
```
hamilton: This isn't being used

src/actions/actions.js
```diff
@@ -72,6 +92,9 @@ export function importInitialJsmd(importedState) {
       type: 'UPDATE_EVAL_FRAME_FROM_INITIAL_JSMD',
       stateUpdatesFromEditor,
     })
+    // FIXME: the following is a hack to make sure the MD is available
```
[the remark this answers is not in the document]
bcolloran: hmm i have some ideas about fixing up the init/import stuff a follow-up, can we wait on this one?

src/actions/actions.js
bcolloran: we spoke, decided we'll take this in a follow up

src/actions/jsmd-parser.js
bcolloran: fixed with commit  ff8094f

src/actions/task-definitions.js
```diff
@@ -362,12 +362,12 @@ tasks.changeToEditMode = new UserTask({
   callback() { dispatcher.changeMode('EDIT_MODE') },
 })
 
-tasks.changeToCommandMode = new UserTask({
-  title: 'Change to Command Mode',
-  keybindings: ['esc'],
-  preventDefaultKeybinding: true,
-  callback() { dispatcher.changeMode('COMMAND_MODE') },
-})
+// tasks.changeToCommandMode = new UserTask({
```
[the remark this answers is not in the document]
bcolloran: i think you've got this one, since you're working on removing command mode ATM?

src/eval-frame/actions/actions.js
```diff
@@ -179,6 +179,48 @@ export function evalConsoleInput(languageId) {
...
+}
+
+export function evaluateText(evalText, evalType, evalFlags) {
+  // allowed types:
+  // md
+  return (dispatch, getState) => {
+    // exit if there is no code to eval or no eval type
+    // if (!evalText || !evalType) { return undefined }
+    // FIXME use evalFlags for something real
+    if (evalFlags.length) console.log(evalFlags)
```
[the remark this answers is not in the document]
bcolloran: fixed via https://github.com/iodide-project/iodide/pull/1152/commits/fbb0c863e58f68ed41541d951dc417a4da2afc57

src/eval-frame/components/panes/report-pane.js
```diff
@@ -1,25 +1,50 @@
...
+      startLine: PropTypes.number.isRequired,
+      endLine: PropTypes.number.isRequired,
+    })),
+  }
 
-export default class ReportPaneUnconnected extends React.Component {
   render() {
+    const mdComponents = this.props.reportChunks.map((chunk) => {
+      const key = `chunk_${chunk.startLine}-${chunk.endLine}`
+      // FIXME: 'html' chunks are really markdown chunks --
```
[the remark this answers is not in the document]
bcolloran: i'm allowing `%% html` chunks, but strictly speaking we're not treating them as "true" html, but rather as the markdown super-set of html. not really a big deal, maybe doesn't need to be changed ever

src/actions/actions.js
```diff
@@ -33,6 +36,23 @@ export function updateAppMessages(messageObj) {
...
+export function updateJsmdContent(text) {
+  return (dispatch) => {
+    const jsmdChunks = jsmdParser(text)
+    const reportChunks = jsmdChunks
+      .filter(c => c.chunkType === 'md' || c.chunkType === 'html')
+    dispatch({
+      // this dispatch really just forwards to the eval frame
+      type: 'UPDATE_MARKDOWN_CHUNKS',
+      reportChunks,
+    })
```
wlach: Maybe there's something I'm missing, but I'm wondering if maybe we should consolidate UPDATE_MARKDOWN_CHUNKS into UPDATE_JSMD_CONTENT, and have the eval frame process the latter? I don't feel like one action really makes sense without the other. Maybe we should also parse out the report chunks in the reducer?

mdboom:
Looks great!

It's a small detail, but the little line between chunks is really nice.

From dogfooding, I noticed that it doesn't seem to work with existing notebooks for me -- pressing Ctrl+Enter does nothing, nothing in the DevTools console.  Should I file an issue to make sure this works with existing notebooks (even if it's just a mass conversion of the database...), or is that a "just me" thing?

src/components/codemirror-jsmd-mode.js
```diff
@@ -0,0 +1,53 @@
...
+        state.localState = mdMode.startState()
+      } else if (stream.sol() && stream.match(/%%\s*fetch.*/, true)) {
+        thisToken = 'line-background-cm-jsmd-delim-line'
+        state.localMode = fetchMode
+        state.localState = fetchMode.startState()
+      } else if (stream.sol() && stream.match(/%%.*/, true)) {
+        // NB: THIS ELSE IF HAS TO COME LAST !!!!!
+        // any '%%' without a known chunk type just
+        // delimits the block without changing mode
+        thisToken = 'line-background-cm-jsmd-delim-line'
```
mdboom: Should it do that, or switch to a generic mode with no syntax highlighting?

mdboom:
Another case not covered (not nitpicking, because this is great work, just want to make sure these things don't regress):

```
iodide.output.element('div')
```

```
TypeError: document.getElementById(...) is null; can't access its "append" property
sideEffectDiv()@http://localhost:8000/iodide.eval-frame.dev.js:104084:3
element()@http://localhost:8000/iodide.eval-frame.dev.js:104097:17
eval()@cell:1:1
```

src/components/codemirror-jsmd-mode.js
bcolloran: important point: a %% delim without a known chunk type will inherit the chunk type above. so this is valid:
```
%% js
var x = new Array(10).fill(1)
%%
y = x.map((v,i)=>i)
```

in this case, we add the token style for the delimiter line, but we don't need to change the mode

bcolloran:
> Another case not covered (not nitpicking, because this is great work, just want to make sure these things don't regress):
> 
> ```
> iodide.output.element('div')
> ```
> 
> ```
> TypeError: document.getElementById(...) is null; can't access its "append" property
> sideEffectDiv()@http://localhost:8000/iodide.eval-frame.dev.js:104084:3
> element()@http://localhost:8000/iodide.eval-frame.dev.js:104097:17
> eval()@cell:1:1
> ```

not interpreted as a nitpick at all, thank you for the feedback!!

src/components/codemirror-jsmd-mode.js
mdboom: I think for the empty chunk type, that makes sense, but for:

```
%% js
var x = 42
%% rust

```

Say, we don't know about rust -- we don't want it to be highlighted or treated as js do we?

src/actions/actions.js
bcolloran: this is admittedly a hack for now, but i think it's the right thing pending #1175 -- for now, this weird pattern is the "correct" way of sending the report chunks content to the eval frame without sending down additional data.

src/handle-initial-jsmd.js
```diff
@@ -1,10 +1,12 @@
 import { stateFromJsmd } from './tools/jsmd-tools'
 import { getStateFromJsmdQueryParams, getUrlParams, JSMD_QUERY_PARAMS } from './tools/query-param-tools'
 import { updateAppMessages, importInitialJsmd } from './actions/actions'
+import { jsmdParser } from './actions/jsmd-parser'
 import { getNotebookInfoFromDocument } from './tools/server-tools'
 
 export default async function handleInitialJsmd(store) {
   let state
+  let jsmd = ''
```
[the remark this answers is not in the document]
bcolloran: fixed with https://github.com/iodide-project/iodide/pull/1176

bcolloran:
ah, gotcha! great point and good catch!

src/components/codemirror-jsmd-mode.js
bcolloran: ah, gotcha @mdboom! great point and good catch!

PR #1153, opened 2018-11-13
Very basic google analytics support

wlach:
If you add a GA_TRACKING_ID variable in settings, it will add the
google analytics tracking code to each page.


bcolloran:
i can't say that i follow all this, but it looks totally reasonable and sane on the face of it. @hamilton should probably do a quick pass

PR #1122, opened 2018-11-02
Add pyodide resources to Iodide build

mdboom:
@wlach: I have no idea if this is the right approach.

`pyodide-demo` (which should probably be renamed to `pyodide-deploy` at some point) is repo where the pyodide built resources get uploaded to by CI.

This works with the local `make up` server, and I can load a language plugin as follows:

```
{
  "languageId": "py",
  "displayName": "Python",
  "codeMirrorMode": "python",
  "keybinding": "p",
  "url": "/pyodide/pyodide.js",
  "module": "pyodide",
  "evaluator": "runPython",
  "asyncEvaluator": "runPythonAsync",
  "pluginType": "language"
}
```

EDIT:

To add some context for this change.  @rth's work to get Scipy working includes a Scipy package that is larger than 100MB, exceeding the maximum file size limit for Github pages.  I'd like to merge that and start making that publicly available, even though the long term plan is to do some "hard things" to get that size down.  

By deploying to Heroku we can get around this limit.  We don't *have to* include it as part of Iodide's deployment (it could be its own static Heroku project, or use some other hosting), but it seems like it saves a little effort for now to do it that way.  This also gives you a local copy of pyodide as part of `npm run build`, which could be handy for doing local/offline demos, but again not critical.

PR #1173, opened 2018-11-14
Fix unit tests broken by #1122

mdboom:
This just avoids testing of all build products.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     },
     "setupFiles": [
       "jest-localstorage-mock"
-    ]
+    ],
+    "testPathIgnorePatterns": ["/node_modules/", "/build/"]
   },
   "dependencies": {
     "@material-ui/core": "^3.0.0",


mdboom:
pyodide has a file called `test.js` which `jest` picks up and tries to run unit tests in.  It actually describes the file contents of pythons `test` package, so we can't easily just rename it.  I don't know why `jest` looks in `build` to begin with, and don't think there's anything it was picking up in there before.

wlach:
Ok, and I assume you're excluding `node_modules` just as a precautionary measure? If so, sounds good to merge.

mdboom:
Apparently `/node_modules/` is the default value, and you need to retain it to retain that default.  https://jestjs.io/docs/en/configuration.html#testpathignorepatterns-array-string

wlach:
Ah ok, might be good to note than in a comment

PR #1178, opened 2018-11-14
Make deploying Pyodide alongside Iodide work

mdboom:
This is a follow-on to #1122 to actually make everything work.

1. Had to add a mimetype for WebAssembly to Whitenoise (since browsers require that the .wasm mimetype is correct)

2. Provides a way to pass a base URL to Pyodide's language plugin loader.  This is important so that Pyodide knows the URL at which to find all it's other resources, whether loaded from github pages, a local development checkout or on Heroku.  It's a bit of a hacky pattern (setting a variable on `window` temporarily), but fwiw, this is a similar pattern to what emscripten does all over the place to load complex things.  I'm definitely open to suggestions there.  This will require https://github.com/iodide-project/pyodide/pull/259 to be merged before it will work.

3. Setting the version of Pyodide to install in `webpack.config.js` and using that in both the `install_pyodide` script and from `language-definitions.js`.  I think having the pyodide version in the url is a good thing for caching reasons, and shouldn't be an issue now that we've removed `meta.languages` so there's no "sticky" path to the language plugin in the notebooks anymore.

I feel like this is sewing Iodide and Pyodide a little more at the hip, which wasn't really my intention.  (2) would be a good idea to make language plugins "URL portable" in any event.  The other things are really sort of deployment stuff, which could always change pretty easily if we want to deploy Pyodide separately from Iodide again in the future.

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -196,3 +196,6 @@ STATIC_URL = EVAL_FRAME_ORIGIN
 # Create hashed+gzipped versions of assets during collectstatic,
 # which will then be served by WhiteNoise with a suitable max-age.
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
+
+# Add a MIME type for .wasm files (which is not included in WhiteNoise's defaults)
+WHITENOISE_MIMETYPES = {'.wasm': 'application/wasm'}
--- a/src/eval-frame/actions/language-actions.js
+++ b/src/eval-frame/actions/language-actions.js
@@ -50,6 +50,8 @@ function loadLanguagePlugin(pluginData, historyId, cell, dispatch) {
         // see the following for asynchronous loading of scripts from strings:
         // https://developer.mozilla.org/en-US/docs/Games/Techniques/Async_scripts
 
+        window.languagePluginUrl = url
+
         // Here, we wrap whatever the return value of the eval into a promise.
         // If it is simply evaling a code block, then it returns undefined.
         // But if it returns a Promise, then we can wait for that promise to resolve
@@ -64,6 +66,7 @@ function loadLanguagePlugin(pluginData, historyId, cell, dispatch) {
           postMessageToEditor('POST_LANGUAGE_DEF_TO_EDITOR', pluginData)
           dispatch(updateCellProperties(cell.id, { evalStatus, rendered }))
           dispatch(updateValueInHistory(historyId, value))
+          delete window.languagePluginUrl
           resolve()
         })
       })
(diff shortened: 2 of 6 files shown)


server/settings.py
```diff
@@ -196,3 +196,6 @@
 # Create hashed+gzipped versions of assets during collectstatic,
 # which will then be served by WhiteNoise with a suitable max-age.
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
+
+# Add a MIME type for .wasm files (which is not included in WhiteNoise's defaults)
+WHITENOISE_MIMETYPES = {'.wasm': 'application/wasm'}
```
wlach: Is this something we should (also) address upstream?

src/eval-frame/actions/language-actions.js
```diff
@@ -50,6 +50,8 @@ function loadLanguagePlugin(pluginData, historyId, cell, dispatch) {
         // see the following for asynchronous loading of scripts from strings:
         // https://developer.mozilla.org/en-US/docs/Games/Techniques/Async_scripts
 
+        window.languagePluginUrl = url
```
wlach: Hmm, yeah, this feels a little icky to me :/ Could we not modify the responseText and pass the value in there? If not, I guess this is ok.

server/settings.py
mdboom: That's not a bad idea.

mdboom:
@wlach: This is "Working for me"™ and passing the tests, so probably ready for some additional testing from you.  Thanks!

wlach:
@mdboom this works well both locally and in production. the one thing i noticed is that we're not using any kind of compression (either brotli or gzip) on any of the assets we're serving. we really ought to fix that, but maybe we should just file a followup issue as I'm pretty sure this was broken before too. feel free to merge as convenient!

mdboom:
Great @wlach.  I created https://github.com/iodide-project/iodide/issues/1184 to track the compression issue.

rth:
@mdboom So actually what is the url to get the pyodide assets from iodide?

I thought it was the same as before but, e.g. https://iodide.io/pyodide-demo/scikit-learn.js returns HTTP 404.  Looking at the code here, tried https://iodide.io/pyodide-4.0.0/pyodide.js as well without much success..

mdboom:
The old approach (using Github pages) wouldn't scale to the file sizes required by Scipy.

The new URL is: https://extremely-alpha.iodide.app/pyodide-0.4.0/pyodide.js

mdboom:
...and of course there are github releases, too, for folks who want to do their own deployment of the resources.

rth:
Thanks @mdboom ! Opened https://github.com/iodide-project/pyodide/issues/294 to discuss it in more detail..

Issue #1174, opened 2018-11-14
make `iodide.output` APIs work


bcolloran:
blocks #1152 

steps:
- need to create target divs in the report. (these could maybe somehow be keyed by text hash to prevent redrawing all the time, tho some nuance will be needed to prevent collisions when 2+ chunks have identical text; perhaps just check the preceding hashes, and if an identical one exists, append a number? so: `${hash}`, `${hash}-1`, `${hash}-2`)
- need the code evaluation stuff to have access to the div id that maps to the code chunk being evaled

PR #1179, opened 2018-11-15
makes the css cell work again, and lays groundwork for #1174

### LINKED ISSUE #1158: css cells should automatically update
The `JSMD Editor` PR #1152 implements instant-ish MD editing. Now that we have somewhat reasonable separation with the iframe, there's no reason we can't do the same with css cells.

bcolloran:
fixes #1158.

the stuff here with the chunkId based on a hash of the content will be useful for #1174. also, by making sure that each chunk to be rendered in the report has a (mostly) stable ID based on the hash of its contents, i think we can avoid some re-renders in react

### DIFF
--- a/src/actions/jsmd-parser.js
+++ b/src/actions/jsmd-parser.js
@@ -1,3 +1,17 @@
+function hashCode(str) {
+  // this is an implementation of java's hashcode method
+  // https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript
+  let hash = 0
+  let chr
+  if (str.length === 0) return hash;
+  for (let i = 0; i < str.length; i++) {
+    chr = str.charCodeAt(i);
+    hash = ((hash << 5) - hash) + chr // eslint-disable-line
+    hash |= 0 // eslint-disable-line
+  }
+  return hash.toString()
+}
+
 export function jsmdParser(fullJsmd) {
   const jsmdLines = fullJsmd.split('\n')
   const chunks = []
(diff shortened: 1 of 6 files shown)


src/actions/jsmd-parser.js
```diff
@@ -1,20 +1,49 @@
...
+    return `${hash}_${hashNum}`
+  }
 
   for (const [i, line] of jsmdLines.entries()) {
     if (line.slice(0, 2) === '%%') {
       // if there is any match, a new chunk has started
       // push the current chunk (unless it's on line 0), then reset
       if (i !== 0) {
         // DON'T push a chunk if we're only on line 0
+        chunkContent = currentChunkLines.join('\n')
```
wlach: I think it would be better to declare the chunkContent as `const` here, instead of up around line 22.

src/eval-frame/components/panes/report-pane.js
```diff
@@ -9,29 +9,46 @@ import MarkdownItAnchor from 'markdown-it-anchor'
 const mdIt = MarkdownIt({ html: true })
 mdIt.use(MarkdownItKatex).use(MarkdownItAnchor)
 
+const mdDiv = html => (<div
+  className="user-markdown"
+  dangerouslySetInnerHTML={{ __html: html }} // eslint-disable-line react/no-danger
```
wlach: Is there any way we can avoid using dangerouslySetInnerHTML here? It doesn't *seem* like it should be necessary.

bcolloran: unfortunately, i don't think there is a better way to do this. The markdown-it parser eats a markdown string and returns an HTML string. I've looked at a couple markdown-it react wrappers, but they just use dangerouslySetInnerHTML under the hood.

src/actions/jsmd-parser.js
bcolloran: addressed with https://github.com/iodide-project/iodide/pull/1179/commits/48c94233566bead592f4ef1b0c5ac3cbd03b8b80

Issue #1166, opened 2018-11-14
remove all legacy references to "cells" from iodide codebase


hamilton:
JSMD editing will drastically simplify Iodide. One such way is we will not really need cells as a concept.

bcolloran:
this is done

PR #1181, opened 2018-11-15
clears out userTasks around cells (which fixes some kbd shortcut collisions)

bcolloran:


bcolloran:
this is work towards #1166, but there is much more to do

PR #1188, opened 2018-11-16
Don't save unchanged language definitions

### LINKED ISSUE #1187: Server notebooks get stuck on old pyodide versions
Just noticed that my prototype dashboard is stuck using an old version of pyodide:

https://extremely-alpha.iodide.io/notebooks/131/

looking at the source, I see that the language plugin is hard defined inside the jsmd:

```
%% meta
{
  "title": "is nightly up to date yet?",
  "lastSaved": "2018-11-09T19:31:06.969Z",
  "languageDefinitions": {
    "py": {
      "languageId": "py",
      "displayName": "Python",
      "codeMirrorMode": "python",
      "keybinding": "p",
      "url": "https://iodide.io/pyodide-demo/pyodide.js",
      "module": "pyodide",
      "evaluator": "runPython",
      "asyncEvaluator": "runPythonAsync",
      "pluginType": "language",
      "codeMirrorModeLoaded": true
    }
  },
  "lastExport": "2018-11-15T17:59:07.115Z"
}
```

Not sure what the right thing to do here is. I can see why we would want to do this (so that old notebooks are guaranteed to stay working even across upgrades to the pyodide "stack"), but it also sucks not to be able to use the latest stuff if you want to. :) This is also rather confusing (it took me 5-10 to figure out what was going on, and I'm an iodide developer).

I guess we could offer some kind of interface to select/upgrade the language definitions/plugins used by a notebook? Any thoughts @mdboom ?

mdboom:
Fix #1187.

In 4db6194a, the behavior was changed so it would only write out the languageDefinition to the jsmd if it was different from the default.  The idea being, if the user didn't specify the defaults, they probably don't care.  Unfortunately, there is an ephemeral item that was getting tacked onto state's copy of the languageDefinition (`codeMirrorMoreLoaded`) that's causing it to always be different.  This only compares keys that should have an actual difference.

We still need a UX to choose and change language plugins, but this should at least fix things to the currently intended behavior.

### DIFF
--- a/src/tools/jsmd-tools.js
+++ b/src/tools/jsmd-tools.js
@@ -276,11 +276,10 @@ ${cell.content}`
 
   // Only save language definitions that aren't identical to the built-in ones
   const customLanguageDefinitions = {}
-  const relevantKeys = ['url', 'module', 'evaluator', 'asyncEvaluator']
   for (const language of Object.keys(state.languageDefinitions)) {
     const src = builtinLanguageDefinitions[language]
     const dst = state.languageDefinitions[language]
-    if (relevantKeys.some(x => src[x] !== dst[x])) {
+    if (['url', 'module', 'evaluator', 'asyncEvaluator'].some(x => src[x] !== dst[x])) {
       customLanguageDefinitions[language] = state.languageDefinitions[language]
     }
   }
--- a/src/tools/jsmd-tools.js
+++ b/src/tools/jsmd-tools.js
@@ -276,9 +276,11 @@ ${cell.content}`
 
   // Only save language definitions that aren't identical to the built-in ones
   const customLanguageDefinitions = {}
+  const relevantKeys = ['url', 'module', 'evaluator', 'asyncEvaluator']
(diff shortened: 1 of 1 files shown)


wlach:
@mdboom we can probably just write a small script to clean the db manually for the 4-5 existing notebooks that have been created.

I'm guessing that when we switch to direct jsmd editing, will this type of metadata would be exposed directly? that should make it much easier to clean up after ourselves.

mdboom:
I don't remember what @bcolloran's plans are for the metadata -- I think it just might be removed altogether.

hamilton:
Yes, I think we will probably deprecate `meta` with direct jsmd editing. Our paradigm has shifted a small amount since we came up with it. For one thing, we mostly use `meta` for the notebook title, but this is also saved on the server. I'm personally in favor of moving toward a frontmatter-style yaml tag at the top of the doc, used by tons of static site generators, to set the metadata for the doc. I don't think that's in-scope for the time being, however.

There's more to talk about re: `meta`.

PR #1191, opened 2018-11-19
Add an iodide logo / backlink into the notebook

wlach:
This involved some shuffling of code around, so that the "Iodide logo"
(with its Zilla Sans font) is a reusable component.

### DIFF
--- a/src/shared/iodide-logo.jsx
+++ b/src/shared/iodide-logo.jsx
@@ -9,10 +9,16 @@ export default class IodideLogo extends React.Component {
         variant="title"
         className={css`
           font-family: 'Zilla Slab', serif;
-          font-size:30px;
+          font-size: 30px;
           font-weight: 300;`}
       >
-        <a href={this.props.backLink} target="_self" style={{ color: 'white', textDecoration: 'none' }}>
+        <a
+          href={this.props.backLink}
+          target="_self"
+          className={css`
+            color: white;
+            text-decoration: none;`}
+        >
           Iodide
         </a>
       </Typography>
--- a/src/shared/style/base.jsx
+++ b/src/shared/style/base.jsx
@@ -1,9 +1,9 @@
 import { injectGlobal } from 'emotion'
-import ZillaSlab from './ZillaSlab-Light.woff'
+import ZILLA_SLAB from './ZillaSlab-Light.woff'
 
 export default injectGlobal`
 @font-face {
   font-family: 'Zilla Slab';
-  src: url("/${ZillaSlab}");
+  src: url("/${ZILLA_SLAB}");
 }
 `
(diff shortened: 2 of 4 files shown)


src/shared/iodide-logo.jsx
```diff
@@ -0,0 +1,21 @@
...
+  render() {
+    return (
+      <Typography
+        variant="title"
+        className={css`
+          font-family: 'Zilla Slab', serif;
+          font-size:30px;
+          font-weight: 300;`}
+      >
+        <a href={this.props.backLink} target="_self" style={{ color: 'white', textDecoration: 'none' }}>
```
hamilton: Minor style point - can we use `css` here as well, as we do in the `Typography` element?

src/shared/style/base.jsx
```diff
@@ -0,0 +1,9 @@
+import { injectGlobal } from 'emotion'
+import ZillaSlab from './ZillaSlab-Light.woff'
```
hamilton: Not sure if this makes sense from a style POV, but I usually namespace as if these are constants when I'm bringing in fonts, so maybe `ZILLA_SLAB`. This import is actually just a generated path string, iirc. Open to what you have as well, just wanted to bring it up.

wlach: Sure, I can do that. Was just copying what was specified initially in the base styles js file, iirc

src/shared/iodide-logo.jsx
wlach: Sure thing

wlach:
Agreed that the formatting / sizing of the logo might not be exactly what we want long-term, let's figure out something nicer later.

src/shared/style/base.jsx
hamilton: yep, I definitely didn't do this right when I added zilla slab :)

src/shared/iodide-logo.jsx
wlach: Actually it turns out that if I change this, the logo on hover gets underlined due to this css:

https://github.com/iodide-project/iodide/blob/a9276761e9cb2f1f8364eb1da0abf8105b9e74cd/src/style/menu-and-button-and-ui-styles.css#L124

I wonder if we could just get rid of it? I think the only place it's used is in the header bar, but I don't see much difference in how things work there if I take them out.

hamilton: hmm, that's annoying, but true that it is probably only used in the header, so I'd say let's remove it.

hamilton: ... one of the many, many reasons why the css-in-js paradigm is better for larger apps.

wlach: Actually I'll make the css selector more specific so it will only be applied for links within the header.

PR #1192, opened 2018-11-19
Make sure Iodide logo gets rendered by zilla slab

wlach:
The css generated from material ui, when compiled down, had different
than expected behaviour on the server. Get around this by avoiding
using the "typography" component for the logo and just using a
regular old "h2".

### DIFF
--- a/src/shared/iodide-logo.jsx
+++ b/src/shared/iodide-logo.jsx
@@ -1,13 +1,12 @@
 import React from 'react'
 import { css } from 'emotion'
-import Typography from '@material-ui/core/Typography'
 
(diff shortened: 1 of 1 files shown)


wlach:
Unfortunately my recent patch didn't work in production due to the way webpack compiled down material ui's css and combined it with our own. I fixed this by... not using material ui for rendering this particular component.

I'm not sure if this is the best solution or not, but it does seem to get the job done. In general the way emotion interacts with third-party components which specify their own css seems non-deterministic, which isn't great. 

Though, as we've discussed before, maybe we should just move away from using things like material design, period.

PR #1193, opened 2018-11-20
WIP: remove cell reducer, implement "run notebook" / fix report view, fix autosave

hamilton:
This removes `cell-reducer.js` and associated architecture. This is quite complicated - there have to be changes just about everywhere we use cells, so a lot of files were touched. Specifically, the aim is to:

- [ ] remove all instances of `state.cells` to use `state.jsmd` / `state.jsmdChunks` where appropriate.
- [ ] fix all the notebook instantiation cases beyond the ones already implemented in `direct-jsmd-editing`
- [x] fix autosave functionality, which currently relies on cells
- [x] fix "run all cells" functionality on the path to removing cells
- [x] remove all tests relating to cells
- [x] remove the outputs from the eval frame, which are unused

There's still more work to be done on removing some references to cells, but this PR gets us to a nice reduction in complexity in the state model.


hamilton:
Thanks for the review @wlach! I think there are a few options for title / metadata going forward, and it'll be worth delving into them more in-depth to make sure we're taking care of workflow needs.

src/actions/actions.js
```diff
@@ -195,31 +184,16 @@ export function updateInputContent(text) {
...
     const {
       keybinding,
-      languageId,
       codeMirrorMode,
     } = languageDefinition
     if (keybinding.length === 1 && (typeof keybinding === 'string')) {
       addLanguageKeybinding(
         [keybinding],
-        () => dispatch(changeCellType('code', languageId)),
+        () => {},
```
bcolloran: language key bindings are not needed at all in direct JSMD mode. Can we cut this whole containing block (so that `addLangaugeKeybinding` is never even called) rather than just putting in the dummy function?

src/eval-frame/actions/actions.js
```diff
@@ -216,124 +209,96 @@ export function evaluateText(
...
-        evaluationQueue = evaluationQueue.then(() => dispatch(evaluateLanguagePluginCell(cell)))
-        evaluation = evaluationQueue
-      }
-    } else {
-      cell.rendered = false
-    }
-    return evaluation
-  }
-}
+// function evaluateCodeCell(cell) {
```
bcolloran: can we cut this commented block? (np if you think we'll need it again, but it looks like it will probably go?)

src/state-schemas/__tests__/state-schema.test.js
```diff
@@ -44,16 +44,16 @@ describe('cell schemas should all have disjoint properties', () => {
 describe('state schemas should all have disjoint properties except for "cells" and "viewMode', () => {
   it('mirroredStateProperties, editorOnlyStateProperties have correct intersection', () => {
     expect(getPropertyIntersection(mirroredStateProperties, editorOnlyStateProperties))
-      .toEqual(['cells', 'cellClipboard', 'viewMode'])
+      .toEqual(['cellClipboard', 'viewMode'])
```
bcolloran: i think we can get rid of cellClipboard in another PR, but maybe stick a FIXME in here as a reminder

src/tools/autosave.js
```diff
@@ -81,7 +83,7 @@ function subscribeToAutoSave(store) {
           return
         }
 
-        const currentJsmd = exportJsmdToString(
+        const currentJsmd = exportJsmd(
           store.getState(),
           false,
```
bcolloran: arg `false` no longer needed if?

hamilton:
@bcolloran I think this is ready for another review.

hamilton:
@bcolloran just rebased after merging your `shift+enter` PR, so this is now ready for review.

src/tools/autosave.js
```diff
@@ -39,8 +42,7 @@ function clearAutosave(state) {
 
 function updateAutosave(state, original) {
   const autosaveKey = getAutosaveKey(state)
-  const jsmd = exportJsmdToString(state, false)
-
+  const jsmd = exportJsmd(state, false)
```
bcolloran: teeny nit: `false` is not needed here

src/actions/actions.js
```diff
@@ -444,34 +379,6 @@ export function saveNotebookToServer() {
...
-      postActionToEvalFrame({
-        type: 'ALIGN_OUTPUT_TO_EDITOR',
-        cellId: getSelectedCell(getState()).id,
-        pxFromViewportTop: targetPxToTopOfFrame,
-      })
-    }
-  }
-}
-
 export function insertCell(cellType, direction) {
```
bcolloran: oh should we cut `insertCell` at this time as well?

Issue #1147, opened 2018-11-08
Provide more information to users before they sign up


wlach:
Per legal, we should provide more information on what users are signing up for before they log into iodide.

I think we probably ought to not just hook directly into github's auth flow on the public server, but rather provide some kind of "login" modal which tells the user what they're consenting to.

wlach:
Yeah, we really need to fix this. :) I was just introducing a coworker the extremely alpha site and it was totally *not* obvious what they were signing up for when they clicked on "login" and get a github form.

PR #1194, opened 2018-11-20
Stop requesting permission to create gists with Github login

wlach:
We don't offer to post a notebook to gist anymore, and won't in the
near-term future. It's better to ask for less when it comes to
associating with a third party account.


wlach:
This trivial patch is a precursor to a larger scale refactoring of the login flow that I'm doing as part of #1147.

PR #1196, opened 2018-11-21
Provide a modal describing what they are signing up for on iodide site (fixes #1147)

### LINKED ISSUE #1147: Provide more information to users before they sign up

wlach:


wlach:
Ob screenshot:

image

wlach:
Note, I arbitrarily mentioned the text about sharealike license as its what wikipedia + mdn used. We should obviously review the text here between ourselves and legal to make sure it communicates the right things before we launch.

bcolloran:
in an email, Michael Feldman recommended that we should use CC BY-SA 3.0 -- https://creativecommons.org/licenses/by-sa/3.0/deed.en
perhaps we should mention that explicitly?

wlach:
I'd personally lean against using jargon like `BY-SA` in the modal but `Attribution-ShareAlike 3.0 Unported` sounds fine to me.

wlach:
Adjusted text per suggestions:

image

wlach:
@hamilton gave the verbal ok to this, let's merge!

Issue #1146, opened 2018-11-08
catch React uncaught errors on the top level eval frame container and render a KernelErrorPane msg


hamilton:
This will likely be a follow-on to #1106, but fairly critical to closing the loop on usability issues we haven't encountered. Filing to remind myself to tackle it.

Because of changes made in React 16, we can now catch all uncaught errors happening within React. So issues such as the `window` inspection security error and other future unforseen bugs relating to React rendering can be caught at the top level, and we can surface a useful error message of some kind.

From a end user perspective, these types of bugs usually just completely wreck the iframe, with absolutely no useful or graceful catch to alert the user what's happening. Without this, the screen simply goes blank.

I had originally envisioned #1106 as a way to helpfully alert the user that indeed something bad has happened, but I am now realizing we can easily add some simple messaging on all the panes. Maybe something like this:

```
The kernel has encountered the following error:
${error here}

Please file an [issue](link to github) and we'll get right to fixing it.
You can still save your changes to the editor.
[button: Save and Reload]

```

I think this would actually be a pretty good first issue for someone who is familiar with React, or wants to get more familiar. It involves (1) adding an `ErrorPane` component to the `eval-frame/panes` directory and (2) using `componentDidCatch` to the `EvalContainer` and gracefully show the error message rendered by `ErrorPane`.

iFlameing:
can I work on this issue?

PR #1197, opened 2018-11-22
Adding ErrorPane Component

iFlameing:
[#1146](https://github.com/iodide-project/iodide/issues/1146) 
I added a ErrorPane Component which give user some hint about which pane is not working.
This is the screeshot that I used for testing.
screenshot from 2018-11-23 02-26-08

### DIFF
--- a/src/eval-frame/components/eval-container.jsx
+++ b/src/eval-frame/components/eval-container.jsx
@@ -6,6 +6,7 @@ import ReportPane from './panes/report-pane'
 import DeclaredVariablesPane from './panes/declared-variables-pane'
 import ConsolePane from './panes/console-pane'
 import AppInfoPane from './panes/app-info-pane'
+import ErrorPane from './panes/ErrorPane'
 
 import FixedPositionContainer from '../../components/pane-layout/fixed-position-container'
 
@@ -14,10 +15,24 @@ export class EvalContainerUnconnected extends React.Component {
   static propTypes = {
     reportOnly: PropTypes.bool.isRequired,
   }
+  state = {
+    hasError: false,
+    error: null,
+    errorInfo: null,
+  }
+  componentDidCatch(error, info) {
+    this.setState({
+      hasError: true,
+      error,
+      errorInfo: info,
+    })
+    console.log(error, info)
+  }
+
 
   render() {
     const { reportOnly } = this.props
-    return (
+    let render = (
       <React.Fragment>
 
         <FixedPositionContainer paneId="ReportPositioner" fullscreen={reportOnly}>
@@ -35,6 +50,17 @@ export class EvalContainerUnconnected extends React.Component {
         <FixedPositionContainer paneId="AppInfoPositioner" hidden={reportOnly}>
           <AppInfoPane />
         </FixedPositionContainer>
+
+      </React.Fragment>)
+    if (this.state.hasError) {
+      render = <ErrorPane error={this.state.error} errorInfo={this.state.errorInfo} />
+    }
+
+
+    return (
+      <React.Fragment>
+
+        {render}
 
       </React.Fragment>
     )
--- a/src/eval-frame/components/eval-container.jsx
+++ b/src/eval-frame/components/eval-container.jsx
@@ -17,8 +17,8 @@ export class EvalContainerUnconnected extends React.Component {
   }
   state = {
     hasError: false,
-    error: null,
-    errorInfo: null,
+    error: '',
+    errorInfo: '',
   }
   componentDidCatch(error, info) {
     this.setState({
--- a/src/eval-frame/components/eval-container.jsx
+++ b/src/eval-frame/components/eval-container.jsx
@@ -26,13 +26,21 @@ export class EvalContainerUnconnected extends React.Component {
       error,
       errorInfo: info,
     })
-    console.log(error, info)
   }
 
 
   render() {
     const { reportOnly } = this.props
-    let render = (
+    if (this.state.hasError) {
+      return (
+        <React.Fragment>
+          <ErrorPane error={this.state.error} errorInfo={this.state.errorInfo} />
+        </React.Fragment>
+      );
+    }
+
+
+    return (
       <React.Fragment>
 
         <FixedPositionContainer paneId="ReportPositioner" fullscreen={reportOnly}>
@@ -50,17 +58,6 @@ export class EvalContainerUnconnected extends React.Component {
         <FixedPositionContainer paneId="AppInfoPositioner" hidden={reportOnly}>
           <AppInfoPane />
         </FixedPositionContainer>
-
-      </React.Fragment>)
-    if (this.state.hasError) {
-      render = <ErrorPane error={this.state.error} errorInfo={this.state.errorInfo} />
-    }
-
-
-    return (
-      <React.Fragment>
-
-        {render}
 
       </React.Fragment>
     )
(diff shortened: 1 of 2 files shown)


src/eval-frame/components/panes/ErrorPane.jsx
```diff
@@ -0,0 +1,28 @@
...
+  const { error, errorInfo } = this.props
+  return (
+    <div>
+      <p>
+        The kernel has encountered the following error:
+        {error}
+        {errorInfo}
+        Please file an [issue](https://github.com/iodide-project/iodide/issues) and we will get right to fixing it.
+        You can still save your changes to the editor.
+        [button: Save and Reload]
```
wlach: I think we actually need a button that does this here. 😄

src/eval-frame/components/panes/ErrorPane.jsx
```diff
@@ -0,0 +1,28 @@
...
+}
+render() {
+  const { error, errorInfo } = this.props
+  return (
+    <div>
+      <p>
+        The kernel has encountered the following error:
+        {error}
+        {errorInfo}
+        Please file an [issue](https://github.com/iodide-project/iodide/issues) and we will get right to fixing it.
```
wlach: This needs to be html, not markdown.

src/eval-frame/components/eval-container.jsx
```diff
@@ -36,6 +51,17 @@ export class EvalContainerUnconnected extends React.Component {
...
+    if (this.state.hasError) {
+      render = <ErrorPane error={this.state.error} errorInfo={this.state.errorInfo} />
+    }
+
+
+    return (
+      <React.Fragment>
+
+        {render}
+
```
wlach: Some extra spacing/newlines in here that we don't need. We should probably configure the linter to flag these...

src/eval-frame/components/eval-container.jsx
```diff
@@ -14,10 +15,24 @@ export class EvalContainerUnconnected extends React.Component {
...
+      errorInfo: info,
+    })
+    console.log(error, info)
+  }
+
 
   render() {
     const { reportOnly } = this.props
-    return (
+    let render = (
```
wlach: I'm not sure why you're putting this content in a temporary and returning a fragment?

I think you could just put the `if (this.state.hasError)` at the top and just return the error if that's what's going on.

src/eval-frame/components/eval-container.jsx
```diff
@@ -14,10 +15,24 @@ export class EvalContainerUnconnected extends React.Component {
   static propTypes = {
     reportOnly: PropTypes.bool.isRequired,
   }
+  state = {
+    hasError: false,
+    error: null,
+    errorInfo: null,
```
wlach: Don't think there's any need to set error/errorInfo to null here.

src/eval-frame/components/eval-container.jsx
```diff
@@ -14,10 +15,24 @@ export class EvalContainerUnconnected extends React.Component {
...
+  }
+  componentDidCatch(error, info) {
+    this.setState({
+      hasError: true,
+      error,
+      errorInfo: info,
+    })
+    console.log(error, info)
```
wlach: We should remove this `console.log` before landing.

src/eval-frame/components/panes/ErrorPane.jsx
```diff
@@ -0,0 +1,28 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+
+class ErrorPane extends Component {
```
wlach: This indentation isn't correct -- the content of the class should be indented two spaces in. Again not sure why our linter isn't picking this up. I'll investigate

src/eval-frame/components/eval-container.jsx
iFlameing: what should I do here because I need to pass the error and errorInfo to the ErrorPane component.can you help How I should do it.

src/eval-frame/components/eval-container.jsx
iFlameing: Yes I do this is just for checking.

src/eval-frame/components/eval-container.jsx
iFlameing: Got it I do the same as you said.

src/eval-frame/components/eval-container.jsx
iFlameing: ok, How much newlines space are allowed?

src/eval-frame/components/eval-container.jsx
wlach: you need to actually take out the line before you can mark this as resolved, thanks!

src/eval-frame/components/eval-container.jsx
wlach: again, please don't resolve until you've landed a new commit which fixes the issue

src/eval-frame/components/eval-container.jsx
wlach: When in doubt, look to the other jsx files for inspiration. In this case, I would do something like this (though note that we probably don't want the render element at all, see above):

```jsx
    if (this.state.hasError) {
      render = <ErrorPane error={this.state.error} errorInfo={this.state.errorInfo} />
    }
    return (
      <React.Fragment>
        {render} 
```

iFlameing:
@wlach I made a new commit please also review it.

src/eval-frame/components/eval-container.jsx
wlach: These object keys will have the value of `undefined` if they are unset, which should be enough to present the right information inside the ErrorPane component.

src/eval-frame/components/panes/ErrorPane.jsx
wlach: Couldn't quite figure out why it was passing, but I filed #1199 which will hopefully put some of the ambiguities around accepted coding style to rest for good.

src/eval-frame/components/panes/ErrorPane.jsx
```diff
@@ -0,0 +1,37 @@
...
+          <h1>{error.toString()}</h1>
+          {errorInfo.componentStack}
+          <br />
+          <h3>
+            Please file an <a href="https://github.com/iodide-project/iodide/issues">issue</a> and we will get right to fixing it.
+            You can still save your changes to the editor.
+          </h3>
+        </p>
+        <button>Save</button>
+        <button>Reload</button>
```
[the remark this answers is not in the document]
hamilton: @iFlameing what's the status on adding callbacks / styling to these buttons?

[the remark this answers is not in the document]
iFlameing: I don't know about what does save button or reload button do in this case can you tell me ?
I styled the button.

iFlameing:
@hamilton after adding the styles now error component looks like this.
screenshot from 2018-12-08 15-10-00

iFlameing:
@hamilton please review it I made some new commit.

src/eval-frame/components/eval-container.jsx
```diff
@@ -14,9 +15,31 @@ export class EvalContainerUnconnected extends React.Component {
   static propTypes = {
     reportOnly: PropTypes.bool.isRequired,
   }
+  state = {
+    hasError: false,
+    error: '',
+    errorInfo: '',
+  }
+  componentDidCatch(error, info) {
+    this.setState({
```
bcolloran: For something like an iframe error, we'd want to bubble that up through the main Redux store, not keep it in local state. Please read up on the Redux paradigm a bit -- in general, using react setState is to be avoided.

iFlameing: ok thanks for review

bcolloran:
Thanks for taking a look at this issue @iFlameing ! I'm going to close this issue for now since it's pretty stale and you're on to other things. You can reopen it later if you decide to do more work on this topic. thanks!

PR #1220, opened 2018-11-28
WIP: throw error if evalType isn't loaded

### LINKED ISSUE #1161: need to show an error in the iodide console if there is an unknown evalType for a jsmd eval
if a user attempts to eval eg
```
%% blarg
[1,2,3]
```

we should display an error like "eval type `blarg` is not defined"

hamilton:
Fixes #1161. As of filing we are still showing one line of the stack trace. It's possible that @mdboom's PR mentioned on gitter will possibly fix this, so instead of fixing it myself I'll wait for that one.


bcolloran:
looks good, but we need to whitelist all known chunk types including md, css, etc before merging

PR #1221, opened 2018-11-28
Use source filename to determine the last stack frame

mdboom:
We don't always eval user code through the same path, and we also want to be able to display exceptions to the user directly from iodide code (and thus have no stack trace).  This should fix that.

### DIFF
--- a/src/components/reps/__tests__/error-handler.jsx
+++ b/src/components/reps/__tests__/error-handler.jsx
@@ -1,6 +1,5 @@
 import { ValueRenderer } from '../value-renderer' // eslint-disable-line
 import errorHandler, { trimStack } from '../error-handler'
-import { runCodeWithLanguage } from '../../../eval-frame/actions/language-actions'
 
 describe('errorHandler shouldHandle', () => {
   it('handles the correct type', () => {
@@ -12,6 +11,7 @@ describe('errorHandler shouldHandle', () => {
   // testing available.
 
   it('trims stack frames', async () => {
+    /* eslint-disable */
     const mockFrames = [
       {
         "isEval": true,
@@ -160,6 +160,7 @@ describe('errorHandler shouldHandle', () => {
         "source": "@http://localhost:8000/iodide.eval-frame.dev.js:1:11"
       }
     ]
+    /* eslint-enable */
 
     expect(trimStack(mockFrames).split('\n').length).toBe(2)
   })


src/components/reps/error-handler.jsx
```diff
@@ -101,10 +99,7 @@ export function trimStack(e) {
   const outputFrames = []
 
   for (const frame of frames) {
-    if ((frame.functionName !== undefined &&
-         frame.functionName.startsWith(runCodeWithLanguage.name)) ||
-        (frame.functionName === undefined &&
-         frame.fileName !== 'cell')) {
+    if (frame.fileName !== undefined || frame.fileName !== 'cell') {
```
hamilton: @mdboom We lose the line number with this, but it's possible that we didn't have it in the first place because `cell` isn't a thing (possibly?). Wondering what we might need to do to get some kind of line number if we want it when code is eval'd.

mdboom: Yeah, line numbers used to work.  I think this is still broken.

mdboom: I was testing on master before -- this seems to behave differently on the jsmd branch...

mdboom:
Ok.  I have something that works locally in Firefox and Chrome with JSMD editing.

However (1) I'm not sure how to test it, since the source paths are different during testing and an actual packaged deployment, and (2) line numbers work, but are relative to the chunk.  We'll need some way of passing in the line number offset of the code that was run to add to those numbers on output.  Or, it might be possible for the code that catches the error (before sending it to the output rep) to update the stack frames before the output handler gets it.  Note that stack frames are a non-standard feature so they behave differently between browser vendors :barf:

src/components/reps/error-handler.jsx
hamilton: Yeah, we haven't addressed this issue yet. Not sure how to atm.

bcolloran:
yea i agree @mdboom, finding a way to pass in the offset seems like it would be good, but off the top of my head i can think of any way of doing it that won't break with anything async/callback. there might be a back door if we eval code by appending script tags with URL blobs, in which case we can get a kind of file name attached to each "eval", but i haven't thought about it that much yet.


in any case, this PR seems like an incremental improvement, so about this PR in particular:

when i run
```
%%
foo = () => bar

%%
foo()
``

i'm seeing output like

```
ReferenceError: bar is not defined
foo()@cell:1:13
eval()@cell:1:1
```
i'm not sure where "cell" is being added, but we should probably track that down and remove it at this point.


also, tests are failing locally for me

```
 FAIL  src/components/reps/__tests__/error-handler.jsx
  ● errorHandler shouldHandle › trims stack frames

    Error: expect(received).toBe(expected) // Object.is equality

    Expected: 2
    Received: 10

      17 |       expect(trimStack(err).split('\n').length).toBe(2)
      18 |     }).catch((err) => {
    > 19 |       throw new Error(err)
         |             ^
      20 |     })
      21 |   })
      22 | })

      at catch.catch.err (src/components/reps/__tests__/error-handler.jsx:19:13)
```

mdboom:
Yeah.  The tests are failing because I can't figure out how to test it without writing a completely different implementation for testing alone.  The filenames that you get in the stack trace are different in testing (you get the original source files) vs. browser (you get the single webpacked `iodide.eval-frame.dev.js`).  As much as I hate to reduce coverage on purpose, I'm prone to just remove the test for now and move on, but maybe you have better ideas.

bcolloran:
as discussed in meeting, we'll write unit tests based on mocked stack frames.

mdboom:
Updated to mock the test data as discussed.  Ready for another round of review.

PR #1222, opened 2018-11-28
restores iodide.output. also fixed a bug in chunk id (hash) generation

bcolloran:
here's some sample content to test:
```
%% md
# a header

%% js
[1,2,3,4].forEach((i) => iodide.output.text(`scripted text ${i}`))

%%md
# another header

%% md
## Using numpy and matplotlib

%% py
import numpy as np
x = np.linspace(0, 2.0 * np.pi, 100)
y = np.sin(x)
y

%% py
z = np.exp(np.sin(x))
z

%% py
from matplotlib import pyplot as plt
plt.figure()
plt.plot(x, y)
plt.show()
```

IMPORTANT CAVEAT (which is basically a reminder of the caveat we discussed in toronto):

because we don't have actual cell objects with stable, persistent ids, and because the list of chunks can be mutated in arbitrarily weird ways via cut and paste, typing, etc, there is not a great way of linking text chunks to divs in the report that allows a stable mapping to be preserved when a chunk is edited (without doing a ton of extra work using eg diffs of operative transforms or something).

as a consequence, when you eval a chunk that uses `iodide.output`, it renders to the div corresponding to that chunk; but as soon as you edit that chunk, the mapping is broken (b/c without doing heavy sorcery, we can't know if you deleted the whole chunk, pasted new code over it, etc) so the output goes away _immediately_; and thus you have to eval the chunk again to see the new output.

this PR still gets us to close to parity with where we are on master in terms of being able to use iodide.output to put stuff in the report, so i don't want to over-rotate on this. in general, i'm less and less sure that `iodide.output` is a good idea in the jsmd context, and i'd rather provide it "as-is" with documentation about it's foibles and a warning that it's a convenience function and that for robust control over positioning etc it's better to target a div. (also recall that, as @hamilton noted, there is actually not a way to make this convenience function target the correct report div when async evals are involved (i think there is not a way to make it work at all, and there is certainly not away to do it without more deep sorcery))

### DIFF
--- a/src/actions/__tests__/jsmd-parser-test.js
+++ b/src/actions/__tests__/jsmd-parser-test.js
@@ -210,6 +210,22 @@ js`
   })
 })
 
+describe('jsmd hashes are distinct for empty cells', () => {
+  const jsmdSample = `%% js
+%% js
+%% js
+%% js`
+
+  it('length of set of hashes is correct', () => {
+    expect(new Set(jsmdParser(jsmdSample).map(c => c.chunkId)).size)
+      .toEqual(4)
+  })
+
+  it('hashes of identical content have correct suffixes', () => {
+    expect(jsmdParser(jsmdSample).map(c => c.chunkId.split('_')[1]))
+      .toEqual(['0', '1', '2', '3'])
+  })
+})
 
 describe('correct chunk types in tricky cases', () => {
   const jsmdSample = `%% chunk1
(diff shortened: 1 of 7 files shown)


bcolloran:
@hamilton also, i noticed that when i "run all" on the jsmd above, it gives the `Loading Python language plugin` message every time (and doesn't end up working). I think you already mentioned something like this to me yesterday, but in case this behavior is not what you were thinking of, i thought i'd mention it.

hamilton:
@bcolloran Hmmm, yeah this behavior appears to be an issue in the base branch itself & run-all, which doesn't appear to be working quite as expected. This matches w/ the behavior I noticed around re-implementing the language plugin stuff. Will have to investigate that separately.

PR #1224, opened 2018-11-28
Make jquery a devDependency

wlach:
It is implicity required by golden-layout (see golden-layout/golden-layout#191) at the
moment, but there is no need for us to require it explicitly.


wlach:
Argh, it turns out we still need to require jquery explicitly for the tests to work. Still, we can at least make it a dev-only dependency.

PR #1227, opened 2018-11-29
Add language drop down for the console

### LINKED ISSUE #1088: add language drop down to iodide console
@mdboom was asked at PyData if pyodide could actually be run in a terminal, so he [slapped together this hilariously lightweight gist](http://htmlpreview.github.io/?https://gist.githubusercontent.com/mdboom/686a05ce026f87d6005c1cc806eef0bb/raw/3d6fa10eae2313be0c7b26b381feeec15ac12ed0/pyodide_console.html). 

It brings to mind a few things:

- the console in iodide should work like a console - hitting `enter` should probably run the REPL line
- having a small dropdown to the left of the arrow which sets the language would be useful. I want to be able to type `import numpy` and start working directly in pyodide. maybe when the console line piece is active we can just use keyboard shortcuts used in the editor to set that.

screen shot 2018-10-22 at 6 51 26 pm

mdboom:
Fixes #1088.

Adds a drop down to the console input line to select a language.  It automatically updates based on the last language evaluated from the JSMD editor as well (this used to work in the cell-based days, but looks like it got disconnected in the JSMD refactor).

It also supports automatically loading the language if it hasn't already been loaded.

### DIFF
--- a/src/eval-frame/actions/actions.js
+++ b/src/eval-frame/actions/actions.js
@@ -140,6 +140,13 @@ export function consoleHistoryStepBack(consoleCursorDelta) {
   }
 }
 
+export function setConsoleLanguage(language) {
+  return {
+    type: 'SET_CONSOLE_LANGUAGE',
+    language,
+  }
+}
+
 export function evalConsoleInput(languageId) {
   return (dispatch, getState) => {
     const state = getState()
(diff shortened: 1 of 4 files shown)


hamilton:
I'm going to continue the work that @mdboom already started with this, since there is one more major architectural change via #1250 that is liable to cause some conflicts. Thanks for getting the ball rolling on this one!

hamilton:
This is superceded by #1489, which was made trivial by the code in this PR - thanks @mdboom!

PR #1251, opened 2018-12-07
Removing   Typography Deprection warning and other console warning  #1164 

### LINKED ISSUE #1164: fix material-ui warnings from the Typography component
Renovate recently upgraded `material-ui`, which now surfaces as deprecation warning. The instructions from the warning on how to migrate the params that'll disappear is pretty straightforward. This should be a quick issue for anyone.

iFlameing:
Fixes #1164


src/components/menu/notebook-header.jsx
```diff
@@ -11,6 +11,10 @@ import AppMessages from '../app-messages/app-messages'
 import IodideModalsRoot from '../modals/iodide-modals-root'
 
 const theme = createMuiTheme({
+  typography: {
+    useNextVariants: true,
+    suppressDeprecationWarnings: true,
```
bcolloran: i must preface this by saying that i dont actually know how this code works, but in general, "solving" a deprecation warning problem by adding a flag called "suppressDeprecationWarnings" seems *very* unlikely to be an actual solution.

i'll let @hamilton look more deeply into this, but i'm inclined to reject this on that basis -- we want to actually update our code to follow the new bet practice for MUI so that we don't get mysterious breakage during some update down the road -- we don't just want to hide the problem

bcolloran:
actually, i'm going to go ahead and close this. there is a lot of noise in the PR stack right now. @iFlameing you are welcome to submit a new PR later if you want to take another try at fixing the underlying problem, but i think that this approach is almost surely not what we want

Issue #1253, opened 2018-12-08
Styling the Profile page


iFlameing:
Changing some style of profile page like notebooks written in small case N must be capital case . The footer of Iodide is not styled.

bcolloran:
@hamilton fixed all this stuff thanks @iFlameing !

PR #1254, opened 2018-12-08
updating the style of profile page #1253

iFlameing:
I changed some css properety of link and also change some color of username and fullname .
Before the changes website looks like this.
screenshot from 2018-12-08 23-13-59
After changing the property it looks like this.
screenshot from 2018-12-08 23-15-19
changes are minor but it looks good from previous.

### DIFF
--- a/src/server/components/footer.jsx
+++ b/src/server/components/footer.jsx
@@ -43,6 +43,9 @@ span {
   left: -3px;
 }
 `
+const aStyle = {
+  textDecoration: 'none',
+};
 
 const Logo = () => (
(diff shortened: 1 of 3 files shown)


wlach:
Hmm, I don't feel like this PR is going in the quite the right direction. Some of the styles are applied locally to some links and not others and other changes are rather arbitrary without clear rationale (i.e. expanding the width of the notebook list). I think I'm going to close this for now -- this page definitely needs work, but we should probably figure out what we really want/need to do with it rather than making small tweaks here and there. Hope that isn't too discouraging.

PR #1257, opened 2018-12-09
Test for EditorModeToolbarUnconnected

iFlameing:


src/components/menu/__tests__/editor-mode-toolbar.test.js
```diff
@@ -0,0 +1,35 @@
...
+  it('Should render EditiorModeTitle', () => {
+    const viewMode = ['EXPLORE_VIEW', '']
+    viewMode.forEach((value) => {
+      wrapper.setProps({ viewMode: value })
+      expect(wrapper.find(EditorModeTitle)).toHaveLength(1)
+      expect(wrapper.find(ViewControls)).toHaveLength(1)
+      expect(wrapper.find(IodideLogo)).toHaveLength(1)
+    })
+  })
+  it('should have display none property', () => {
```
bcolloran: in these descriptions, say more about the condition you're testing, e.g.:
'should have display none when viewMode !== EXPLORE_VIEW'

src/components/menu/__tests__/editor-mode-toolbar.test.js
```diff
@@ -0,0 +1,35 @@
...
+import EditorModeTitle from '../editor-mode-title'
+import ViewControls from '../view-controls'
+import IodideLogo from '../../../shared/iodide-logo'
+
+describe('<EditorModeToolbarUnconnected', () => {
+  let wrapper
+  beforeEach(() => {
+    wrapper = shallow(<EditorModeToolbarUnconnected />)
+  })
+  it('Should render EditiorModeTitle', () => {
```
bcolloran: what are you actually testing with this "it" block? looks like it's more than EditiorModeTitle, right?

this description should be something like: "should render all child components correctly"  ... but why are you only testing 3 of the 4 components? EditorModeControls is missing.

iFlameing:
@bcolloran we don't need to test mapstaetoprops because I am already setting props using setProps method provide by jest and viewmodetoolbar.js doesn't depend on mapstatetoprops.

can you review this once again?

bcolloran:
@iFlameing you are testing _EditorModeToolbarUnconnected_  **NOT** "viewmodetoolbar.js" -- in fact, no such file or component exists, so i'm not sure what you have in mind.

But the EditorModeToolbar component _does_ depend on mapStateToProps -- see:
https://github.com/iodide-project/iodide/blob/master/src/components/menu/editor-mode-toolbar.jsx
 -- and thus we *do* need to test the mapStateToProps function.

Now that you mention it, using setProps is not the preferred pattern for setting props in these tests. Please consider the pattern in the files below:
https://github.com/iodide-project/iodide/tree/master/src/eval-frame/components/panes/__tests__

Here is an example of a simple instance of mapStateToProps:
https://github.com/iodide-project/iodide/blob/master/src/eval-frame/components/panes/console-pane.jsx
And here is a simple test for that:
https://github.com/iodide-project/iodide/blob/master/src/eval-frame/components/panes/__tests__/console-pane.test.js

You're very close to having a successful PR here! Thanks for sticking with it @iFlameing  :-)

iFlameing:
@bcolloran can you merge this now or I need to make some more changes.

src/components/menu/__tests__/editor-mode-toolbar.test.js
```diff
@@ -0,0 +1,65 @@
...
+import EditorModeTitle from '../editor-mode-title'
+import ViewControls from '../view-controls'
+import IodideLogo from '../../../shared/iodide-logo'
+import EditorModeControls from '../editor-mode-controls'
+
+describe('EditorModeToolbarUnconnected', () => {
+  let mountedPane
+  let props
+  const wrapper = () => {
+    if (!mountedPane) {
```
bcolloran: maybe change `mountedPane` -> `mountedComponent`, which is a bit more generic.
(in this case the component we're testing is not a pane)

iFlameing: Sorry I missed it. I made the changes.

bcolloran:
Nice job @iFlameing, thank you for the PR and for sticking with it during the review!

This is a very nice component test, and we could really use help writing more of them. This would be a great way to contribute while also learning about our code base and our coding practices. I would highly recommend contributing more component tests as a great way to get involved! Thanks!

PR #1269, opened 2018-12-10
Upgrade the pyodide version

mdboom:
This includes Scipy (for the first time in this deployment), NetworkX and fixes a nasty race condition crash that @acmiyaguchi found last week.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -26,7 +26,7 @@ let { EDITOR_ORIGIN } = process.env
 let { EVAL_FRAME_ORIGIN } = process.env
 let { IODIDE_PUBLIC_TOS } = process.env || false
 
-const PYODIDE_VERSION = process.env.PYODIDE_VERSION || '0.2.0'
+const PYODIDE_VERSION = process.env.PYODIDE_VERSION || '0.4.0'
 process.env.PYODIDE_VERSION = PYODIDE_VERSION
 
 const APP_VERSION_STRING = process.env.APP_VERSION_STRING || 'dev'


hamilton:
@mdboom I'm getting an error attempting to run this:

```python
import networkx as nx

options = {
	'node_color': 'black',
	'node_size': 100,
	'width': 3,
}

G = nx.dodecahedral_graph()
shells = [[2, 3, 4, 5, 6], [8, 1, 0, 19, 18, 17, 16, 15, 14, 7], [9, 10, 11, 12, 13]]
nx.draw_shell(G, nlist=shells, **options)

```

Resulting in 
```
Error: Traceback (most recent call last):
  File "/lib/python3.7/site-packages/pyodide.py", line 41, in eval_code
    return eval(compile(expr, '<eval>', mode='eval'), ns, ns)
  File "<eval>", line 12, in <module>
  File "/lib/python3.7/site-packages/networkx/drawing/nx_pylab.py", line 1056, in draw_shell
    draw(G, shell_layout(G, nlist=nlist), **kwargs)
  File "/lib/python3.7/site-packages/networkx/drawing/layout.py", line 223, in shell_layout
    import numpy as np
ModuleNotFoundError: No module named 'numpy'

_hiwire_throw_error()@http://localhost:8000/pyodide-0.4.0/pyodide.asm.js:8:39275
{anonymous}()@http://localhost:8000/pyodide-0.4.0/pyodide.asm.wasm:wasm-function[359]:0xcf55f
{anonymous}()@http://localhost:8000/pyodide-0.4.0/pyodide.asm.wasm:wasm-function[368]:0xd0b12
pyodide/Module.__runPython()@http://localhost:8000/pyodide-0.4.0/pyodide.asm.js:8:1947571
_runpython_init_js/Module._runPythonInternal()@http://localhost:8000/pyodide-0.4.0/pyodide.asm.js:8:42686
internal()@http://localhost:8000/pyodide-0.4.0/pyodide.asm.js:8:43256
runInternal()@http://localhost:8000/pyodide-0.4.0/pyodide.asm.js:8:43628
```

Seems like perhaps `numpy` is not listed as a dependency for `networkx`?

hamilton:
Yep, in fact I don't see it pulling down `numpy`:

screen shot 2018-12-11 at 10 33 18 am

hamilton:
FWIW this code chunk works if I include the explicit imports of numpy and matplotlib, so it looks like the dependencies don't get pulled down properly.

mdboom:
This is actually kinda complicated.  Numpy and matplotlib are not actually hard dependencies of networkx, since it's possible to do a lot of the graph theoretical things without them.  It's only when you try to plot that matplotlib tries to import.  It looks like anaconda suffers this same fate if you make a clean environment and put networkx in it.  So I'm not sure that putting `numpy` and `matplotlib` as dependencies on the package is the right solution.

Unfortunately, the *magic* that happens to make `import matplotlib` go out and fetch a package only happens by parsing the initial chunk of code send to pyodide.  This is both because (1) being too magical too far down the stack could have a lot of unintended consequences and (2) we can't do an asynchronous import, so we have to do it before running any Python at all.

We could trap the import error and say something helpful like "add `import matplotlib` to your code to use this feature"...

hamilton:
@mdboom Dang, that's a real shame. 

So it seems like the choice at this juncture is either (1) expect the user to work through that import problem on their own (this kind of goes against the paradigm that we can just copy and paste code and it works), or (2) force them to pay the penalty to get numpy + matplotlib in case they actually end up using it (also not ideal, but at least code will always somehow magically work). If we're not going to do (2), I'd be in favor of some help text in the error that suggests that they add an `import` statement, which is _almost_ magic :) 

That all said, I'd hate to block the pyodide update since that race condition would be nice to take care of. I otherwise have no problem with the PR and think we should merge it so we get the other parts too.

mdboom:
I'm in favor of merging this now and creating a new issue to come back to all the nonsense (which I agree is not ideal) later.

PR #1270, opened 2018-12-11
Make Brotli compression work

### LINKED ISSUE #1184: Make brotli compression of static resources work in server deployment
As @wlach mentioned [here](https://github.com/iodide-project/iodide/pull/1178#issuecomment-439198007), I don't think the static resources are actually being served compressed.  Our toolchain has the ability to do so, so we should probably get that working.

pradeepgangwar:
As suggested by @wlach in the [here](https://github.com/iodide-project/iodide/issues/1184#issuecomment-439444163) I executed the commands and create this PR. Let me know if this needs changes.
Fixes #1184 

CC- @wlach @bcolloran @hamilton  

Please review.


wlach:
Unfortunately my instructions were wrong, at least for deploying on heroku. We need to install brotli as a default package, because heroku's buildpack doesn't install "dev" packages by default, so the whitenoise step won't work. I'm going to close this and open up another PR. Thanks for the contribution all the same!

PR #1276, opened 2018-12-12
move evaluation queue to editor

hamilton:
This PR will bring the evaluation queue directly into the editor, waiting for a response from the eval frame after evaluation before sending the next code chunk.

The base branch breaks plugin cells, and because of how we've moved things around, this PR was necessary to fix plugin cell loads when `evaluateNotebook` was called. That said, we were planning on doing this anyway.

The single biggest UX gain here is that when we receive an `ERROR` status we can halt further evaluation in the queue. Once the queue is reset, we can resume evaluation, bringing our evaluation model much closer to Jupyter's (in spirit). Further improvement could be to highlight the chunk that caused the `ERROR` status to be generated, giving better traceability. This PR doesn't accomplish this, but it makes it possible to link status to code in the editor itself.

A few things left to do to clean this up:

- [x] verify that this approach (specifically in `evaluation-queue.js`) feels right
- [x] clean up the test cases. Though I've added tests that cover the main use cases, they feel a bit messy due to how Jest handles the microtask queue. My workarounds are successful but feel messy.
- [x] I'd like to use this opportunity to bring in the kernel state code I wrote before so we have some semblance of run state visible to the user.


src/actions/__tests__/evaluation-queue.test.js
```diff
@@ -0,0 +1,71 @@
+import { getEvaluationResolvers, resolveEvaluation, rejectEvaluation,
```
hamilton: Testing this is tricky and I think I should probably refactor the API around the queue to make the tests feel a little cleaner.

src/actions/evaluation-queue.js
```diff
@@ -0,0 +1,41 @@
+import { IdFactory } from '../actions/history-id-generator'
+
+export const evalIdGenerator = new IdFactory()
+
+let evaluationQueue = Promise.resolve()
+let evaluationResolvers = {}
+let queueSize = 0
```
hamilton: This could probably live in the state, but not sure if its necessary. I included it mostly so we can keep track via testing.

src/actions/__tests__/evaluation-queue.test.js
```diff
@@ -0,0 +1,72 @@
...
+  appendChunkToEvaluationQueue, evalIdGenerator, getQueueSize } from '../evaluation-queue'
+
+
+describe('appendChunkToEvaluationQueue', () => {
+  it('appropriately resolves the evaluationQueue', async () => {
+    appendChunkToEvaluationQueue({}, () => {})
+    // the await below forces the above microtask queue to finish before moving on.
+    await Promise.resolve().then(() => {})
+    const currentId = evalIdGenerator.state
```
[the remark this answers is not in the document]
hamilton: yeah, this is done only in this test, and frankly I can probably just get rid of it since I know the ID. I don't do this elsewhere

src/actions/evaluation-queue.js
```diff
@@ -0,0 +1,43 @@
+import { IdFactory } from '../actions/history-id-generator'
+
+export const evalIdGenerator = new IdFactory()
```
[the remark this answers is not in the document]
hamilton: Yeah I like that better - will change

src/actions/__tests__/evaluation-queue.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+    await Promise.resolve().then(() => {})
+    expect(getQueueSize()).toBe(0)
+
+    await Promise.resolve().then(() => {})
+  })
+  it('stops the evaluationQueue when rejectEvaluation is called', async () => {
+    appendChunkToEvaluationQueue({}, () => {}) // 4
+    appendChunkToEvaluationQueue({}, () => {}) // 5
+    appendChunkToEvaluationQueue({}, () => {}) // 6, REJECT
```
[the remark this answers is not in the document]
hamilton: 6 rejects because I make it reject below. This all said, even I have trouble reading these tests because they seem maddening! Hence my hesitation.

hamilton:
Refactored a little but, but not much needed to change. Moving toward a class-based model, I think it's somewhat more parse-able from the end-user point of view (they can probably guess what's going on if they encounter `evalQueue.evaluate`, but I don't think it buys that much more in terms of code cleanliness. Both the first approach and the new one already completely hides promises and the queue itself from the user, exposing only a bare API. That said, I figured out the uncomfortable test issues so I can test the whole thing synchronously just fine, and the microtask queue shuffles out nicely.

Going to revisit this tomorrow morning to make sure it's where it needs to be. There appears to be an outstanding issue with the kernel & python code running sending a state update I don't want it to send.

src/actions/__tests__/evaluation-queue.test.js
hamilton: The latest commits refactor the whole thing, and the tests now feel nicer & the API a bit cleaner.

hamilton:
I just checked that outstanding issue. The problem is, the `CircularProgress` component from `material-ui` breaks under heavy computational loads, discussed [here](https://material-ui.com/demos/progress/#limitations). Thus this is not strictly speaking an iodide render bug, but just a drawback of using this component. There are others out there that probably achieve the same effect, but for now I think this one will do just fine.

src/actions/__tests__/verify-action-store-consistency.test.js
```diff
@@ -122,3 +127,18 @@ describe('make sure action creators leave store in a consitent state', () => {
...
+
+describe('setKernelState', () => {
+  it('createValidatedReducer should throw an error if we pass an invalid arg into setKernelState', () => {
+    expect(() => store.dispatch(actions.setKernelState('fake state')))
+      .toThrowError(SchemaValidationError)
+    expect(() => store.dispatch(actions.setKernelState(12342323)))
+      .toThrowError(SchemaValidationError)
+  })
+  it('passes on correct set of enums for setKernelState', () => {
+    const enums = ['KERNEL_LOADING', 'KERNEL_LOAD_ERROR', 'KERNEL_ERROR', 'KERNEL_IDLE', 'KERNEL_BUSY']
```
bcolloran: i think you should be able to import the enum from the state-schema directly, so that this test will always be in sync with the possible option in case we need to add/remove anything later on

PR #1280, opened 2018-12-13
Adding babel-lodash-plugin and webpack-lodash-plugin 

### LINKED ISSUE #1265: make sure we're using the smallest portion of lodash possible
@hamilton noticed in  https://github.com/iodide-project/iodide/pull/1250#discussion_r240007725 that we're not importing lodash optimally.

according to this: https://medium.com/@martin_hotell/tree-shake-lodash-with-webpack-jest-and-typescript-2734fa13b5cd
-- webpack should be mostly tree shakeable

> UPDATE August 2017:
> 
> Webpack ( Tobias Koppers ) is rockin’ hard and with the new pure-module feature, lodash will be tree-shake able by default, so no sub paths import will be needed.
> 
> → `import {get,pick} from 'lodash-es'` will just work! more info :

i don't know if we'd need lodash-es or regular lodash, but investigating that is part of this issue

figuring this out and correcting the imports should be an easy fix

iFlameing:
Fixes #1265 . Now weback only import those function which is used in project and other discarded making bundle size small .

@bcolloran please review this .


package.json
```diff
@@ -48,6 +51,8 @@
     "font-awesome": "^4.7.0",
     "golden-layout": "^1.5.9",
     "lodash": "^4.17.10",
+    "lodash-es": "^4.17.11",
```
bcolloran: if i am understanding correctly, we don't need to add lodash-es, right?

package.json
```diff
@@ -106,7 +112,7 @@
     "redux-mock-store": "1.5.3",
     "style-loader": "0.23.1",
     "uglifyjs-webpack-plugin": "1.3.0",
-    "webpack": "4.27.1",
+    "webpack": "^4.27.1",
```
bcolloran: this change is not needed

package.json
```diff
@@ -38,6 +40,7 @@
     "@material-ui/icons": "^3.0.0",
     "@skidding/react-codemirror": "^1.0.0",
     "ajv": "^6.5.0",
+    "babel-plugin-lodash": "^3.3.4",
```
bcolloran: i think this should be a devDependency

package.json
```diff
@@ -19,7 +20,8 @@
   "jest": {
     "setupTestFrameworkScriptFile": "<rootDir>/test/setupTests.js",
     "transformIgnorePatterns": [
-      "/node_modules/(?!test-component).+\\.js$"
+      "/node_modules/(?!test-component).+\\.js$",
+      "/node_modules/(?!MODULE_NAME_HERE).+\\.js$"
```
bcolloran: what does this line do?

package.json
iFlameing: Yes

package.json
iFlameing: It help jest for transforming any es2015 module used in developing to es5.

iFlameing:
@bcolloran  @hamilton  I just made  changes to previous commit as said by @bcolloran  can you review this ? please

package.json
wlach: This doesn't look right to me, `MODULE_NAME_HERE` seems like a placeholder for something else. i.e. you're supposed to put your module name in to solve an issue:

https://github.com/facebook/jest/issues/6229#issuecomment-403539460

Let's revert this change, as I don't think it's serving any functional purpose (we could also consider taking this line out entirely, perhaps, but as part of another pr)

package.json
```diff
@@ -48,6 +49,7 @@
     "font-awesome": "^4.7.0",
     "golden-layout": "^1.5.9",
     "lodash": "^4.17.10",
+    "lodash-webpack-plugin": "^0.11.5",
```
wlach: I am pretty sure this should be a devDependency also.

webpack.config.js
```diff
@@ -24,7 +25,7 @@ let CSS_PATH_STRING
 
 let { EDITOR_ORIGIN } = process.env
 let { EVAL_FRAME_ORIGIN } = process.env
-let { IODIDE_PUBLIC_TOS } = process.env || false
+const { IODIDE_PUBLIC_TOS } = process.env || false
```
wlach: Please revert this, it's not related to the PR

webpack.config.js
```diff
@@ -134,14 +143,14 @@ module.exports = (env) => {
         IODIDE_BUILD_MODE: JSON.stringify((env && env.startsWith('dev')) ? 'dev' : 'production'),
         IODIDE_REDUX_LOG_MODE: JSON.stringify(reduxLogMode),
         PYODIDE_VERSION: JSON.stringify(PYODIDE_VERSION),
-        IODIDE_PUBLIC_TOS: !!IODIDE_PUBLIC_TOS
+        IODIDE_PUBLIC_TOS: !!IODIDE_PUBLIC_TOS,
```
wlach: Likewise

webpack.config.js
iFlameing: This is a linting issue  what I suppose to do?

wlach: The linter in circle ci doesn't seem to be complaining about this line of code, so reverting it should not result in any errors on the PR.

In fact this probably *should* be const (same with the missing comma below), but we try to keep PRs focused on particular issues, and not an amalgamation of other changes. If you like, you can open up a follow-up to figure out why the linter isn't complaing about these lines in ci.

iFlameing:
@wlach can you now merge this ?

PR #1282, opened 2018-12-13
remove iodide.evalQueue from the API.

hamilton:
# NOTE: do not merge until iodide-project/pyodide#280 is merged

This removes `iodide.evalQueue` entirely. We built this API to make better sense of synchronizing asynchronous code bits within the cell-based notebook environment. We realized over time two things: 

1. cell-based programming feels unmanageable, so we have moved away from that paradigm entirely, so this API doesn't make sense (and indeed doesn't even work anymore), and 
2. because we are primarily only needing to use `iodide.evalQueue` for calls to `fetch`, we created a much cleaner, easier-to-use dependency / file fetching chunk type.


mdboom:
This works great with the change to Pyodide in #280.  I just made a new 0.5.0 release and tag of pyodide so we can get that.

This PR just needs to be updated to change the version of Pyodide in `webpack.config.js` and then we should be golden (which I just did).  It will take a few hours for the new version of Pyodide to end up on github releases.

PR #1283, opened 2018-12-13
remove obsolete stylesheets & styles regarding cells

hamilton:


bcolloran:
this looks good, i think we can merge as soon as the conflict is fixed :+1:

PR #1293, opened 2018-12-20
fixes console

### LINKED ISSUE #1111: should the console follow console or notebook interaction paradigms?
This issue follows on from #1088, which originally captured this idea.

At the moment, the console follows a notebook interaction convention - you evaluate the console the way you do a code chunk or cell. The main argument is, if we allow plain `enter` to evaluate the console but require a different keybinding to evaluate a code chunk from the editor, users will have to manage two different keybindings.

For my part, I think it's good we're taking those keybinding issues seriously because we should definitely strive to get it right - but I'm not sure I'm convinced that following the repl approach would be more confusing than what we currently have. 

Thinking about it a bit more, a console as users know it is a pretty different set of interactions than a notebook, and for good reason - consoles allow for really quick exploration without having to worry about big chunks of code, which I think is a bit harder to do in a notebook environment (even with jsmd). And users may already expect the console to interact like a console, the way they expect the notebook to interact like a notebook. So it remains to be seen that what essentially amounts to a throwaway cell that clears upon evaluation is what users want in a console vs. the thing that they're used to.

The motivation behind the issue: I'm obviously one data point, but it took me a handful of seconds before realizing that the console wasn't broken because plain old `enter` didn't work as I expected. I don't usually use the console, but wanted to explore why an old python script I wrote wasn't working in pyodide. My impulse was to open @mdboom's pyodide repl page to try out some ideas.

bcolloran:
this fixes the console problem mentioned in #1250 by adding local state to the console. it also:
- adds some tests
- pulls the console "actions" from tasks to local handlers
- switches console eval to just use `Enter` (fixes #1111)
- fixes a mismatch in rendering on chrome vs ff
- fixes a bug in where you are in the console history after an eval
- introduces a template for testably connecting eval frame components to message passers (in follow up steps, this can be moved out of the component and made more robust)


bcolloran:
cool added #1323 to track that

PR #1300, opened 2018-12-23
Adding test for ViewModeToggleButtonUnconnected

iFlameing:
@bcolloran can you review this?


bcolloran:
@iFlameing please review the changes i made in commit https://github.com/iodide-project/iodide/pull/1300/commits/754f8addeef82156cd3fc70ede8632225cdb3915 (you can ignore the change in the file src/components/menu/presentation-mode-toolbar.jsx ) to see what kind of refactoring is helpful in these cases.

the idea is to move as much logic as possible out of the component, which is hard to test, and as much of the logic as possible into the mapStateToProps, which is just a pure function and therefore is easy to test. After making these changes, we see that there is only ONE place in the component where the components behavior changes based on it's input (the action fired based on whether or not we're in report view), so this is the only thing we actually need to test because the rest of the component is purely "declarative" (read those articles i linked to to get a better understanding of that)

@hamilton @wlach @mdboom just fyi for you guys as well -- i think you have seen this pattern as well by now, but this is a good example/reminder of my current understanding of how we want our components to look and how we want the relevant tests to look. some highlights:
- as much logic as possible in mapStateToProps; ideally, we want purely declarative components with ZERO logic
- pass in all action creators via mapDispatchToProps
- add proptypes for everything, including the passed-in action creators
- test action creators by passing in jest mock functions as props
- only need to test non-declarative logic bits of components, not the declarative stuff that actually serves as the layout spec

mdboom:
@bcolloran : This bulleted list is golden.  Maybe there's a place we could stick it in the developer docs?

bcolloran:
@mdboom will do

bcolloran:
@mdboom et al. -- added https://github.com/iodide-project/iodide/blob/master/CONTRIBUTING.md#tips-for-writing-and-testing-reactredux-components

mdboom:
Cool, thanks!

PR #1302, opened 2018-12-27
Use iodide-style component for user menu

wlach:
This looks a little nicer than the material design interface and is
consistent between notebook and the rest of the UI.


wlach:
Before:

image


After:

image

src/shared/user-menu.jsx
```diff
@@ -106,46 +98,34 @@ export default class UserMenu extends React.Component {
...
-                    disableRipple
-                    aria-label="more"
-                    aria-owns={anchorElement ? 'user-controls' : null}
-                    aria-haspopup="true"
-                    onClick={this.handleClick}
-                    style={{ color: 'white' }}
+                <div style={{ marginRight: '20px' }}>
+                  <Popover
+                    title={
+                      <div style={{ alignItems: 'center', display: 'inline-flex', color: 'white' }}>
```
hamilton: Could we either define these styles in css (via `import {css} from 'emotion'` or from `import styled from 'react-emotion'`?

Two routes:

- `css`: This means you defined something like ``const popoverContainerStyle = css`align-items: center; ...` `` and then for the div here, it'd be `<div className={popoverContainerStyle}>` or
- `styled`: define ``const PopoverContainer = styled('div')`align-items: center; ...` `` then use it like this on line 109: `<PopoverContainer>...</PopoverContainer>`. I generally prefer this style because it's a bit semantically easier to follow what the heck that div is doing.

I understand there is a broader question about how we should handle style generally in iodide which I'm sure we'll get to in the new year, but for now I personally like the tagged template literal approach because it's more readable and easier to follow if you know css already, and it actually ends up creating some separation of style from structure / behavior (with a nice way to tie them together) which I appreciate.

PR #1305, opened 2018-12-28
Pressing "log in" from action bar does not show modal(#Fixes1230)

### LINKED ISSUE #1230: Pressing "log in" from action bar does not show modal
We just added a login modal in #1196. This works great (mostly) though I just noticed it doesn't popup if you click on the log in link in the action bar:

image

Unfortunately this isn't trivial to fix (it will require moving some logic from `user-menu.jsx` to some new shared component/file, I think) but it is something we need to resolve before we publicly launch.

iFlameing:
@wlach can you review this?

### DIFF
--- a/src/components/menu/header-messages.jsx
+++ b/src/components/menu/header-messages.jsx
@@ -1,6 +1,8 @@
+/* eslint-disable react/jsx-no-bind */
 import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
+import LoginModal from '../../shared/login-modal'
 
 import { connectionModeIsServer, connectionModeIsStandalone } from '../../tools/server-tools'
 import {
@@ -20,9 +22,33 @@ export class HeaderMessagesUnconnected extends React.Component {
     makeCopy: PropTypes.func.isRequired,
     revisionId: PropTypes.number,
   }
+  constructor(props) {
+    super(props)
+    this.state = {
+      showLoginModal: false,
+    }
+    this.showLoginModal = this.showLoginModal.bind(this)
+    this.hideLoginModal = this.hideLoginModal.bind(this)
+  }
+  showLoginModal() {
+    this.setState({
+      showLoginModal: true,
+    })
+  }
+  hideLoginModal() {
+    this.setState({ showLoginModal: false })
+  }
 
   render() {
     let content;
+    let showLoginModal;
+    if (this.state.showLoginModal) {
+      showLoginModal = (<LoginModal
+        visible={this.state.showLoginModal}
+        onClose={this.hideLoginModal}
+        login={this.props.login}
+      />)
+    }
     switch (this.props.message) {
       case 'HAS_PREVIOUS_AUTOSAVE':
         content = (
@@ -53,7 +79,11 @@ export class HeaderMessagesUnconnected extends React.Component {
       case 'NEED_TO_LOGIN':
         content = (
           <span>
-            To save to this server, you need to <a onClick={this.props.login}>login</a>.
+            To save to this server, you need to
+            <a
+              onClick={this.showLoginModal.bind(this)}
+            >login
+            </a>.
           </span>
         )
         break
@@ -77,6 +107,7 @@ export class HeaderMessagesUnconnected extends React.Component {
         className="notebook-header-messages-container"
       >
         {content}
+        {showLoginModal}
       </div>
     )
   }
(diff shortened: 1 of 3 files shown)


iFlameing:
@wlach  sorry for inconvince. If I have to show the login modal to the user?

iFlameing:
@walch can you merge this now . I tested locally and its working fine.

src/components/menu/header-messages.jsx
```diff
@@ -77,6 +107,7 @@ export class HeaderMessagesUnconnected extends React.Component {
         className="notebook-header-messages-container"
       >
         {content}
+        {showLoginModal}
```
wlach: I think you should be able to just inject the login modal jsx properties directly in here, rather than using a variable intermediary. We did that for content because it could be one of many things, but I think we can just unconditionally dump in the `LoginModal` component. The visibility property will take care of whether to show it or not.

src/components/menu/header-messages.jsx
```diff
@@ -53,7 +79,11 @@ export class HeaderMessagesUnconnected extends React.Component {
       case 'NEED_TO_LOGIN':
         content = (
           <span>
-            To save to this server, you need to <a onClick={this.props.login}>login</a>.
+            To save to this server, you need to
+            <a
+              onClick={this.showLoginModal.bind(this)}
```
wlach: I don't think we want to bind this here-- you want to do that in the constructor, then just call it here via `onClick={this.showLoginModal()}`. Lots of examples of this elsewhere in iodide.

iFlameing:
@wlach I made the changes as you suggested and I tested it locally and it is working fine.

src/shared/user-menu.jsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
```
wlach: Please revert the changes to this file, as discussed.

src/components/menu/header-messages.jsx
```diff
@@ -53,7 +73,11 @@ export class HeaderMessagesUnconnected extends React.Component {
       case 'NEED_TO_LOGIN':
         content = (
           <span>
-            To save to this server, you need to <a onClick={this.props.login}>login</a>.
+            To save to this server, you need to
```
wlach: I think you need to put this back on one line, webpack is jamming everything together with the line breaks:
image

(another alternative is to put an `&nbsp;` after `need to`, but I think it's better to just put things on one line)

src/components/menu/header-messages.jsx
```diff
@@ -21,6 +22,25 @@ export class HeaderMessagesUnconnected extends React.Component {
...
+    super(props)
+    this.state = {
+      showLoginModal: false,
+    }
+    this.showLoginModal = this.showLoginModal.bind(this)
+    this.hideLoginModal = this.hideLoginModal.bind(this)
+  }
+
+  showLoginModal() {
+    this.setState({
```
wlach: Didn't notice this before, but this can all just go on one line.

src/components/menu/header-messages.jsx
```diff
@@ -77,6 +101,11 @@ export class HeaderMessagesUnconnected extends React.Component {
         className="notebook-header-messages-container"
       >
         {content}
+        <LoginModal
+          visible={this.state.showLoginModal}
+          onClose={this.hideLoginModal}
+          login={this.props.login}
```
wlach: The login modal doesn't yet disappear when you click on login. I would create a helper function called `login` which hides the modal (or calls `this.hideLoginModal`) and then calls `this.props.login`.

src/components/menu/header-messages.jsx
iFlameing: we can not put in one line because it is giving linting error line exceeds the maximum line length of 100.

wlach: Ah I see, well then fall back to `&nbsp` I guess.

src/shared/user-menu.jsx
iFlameing: should I revert all the changes of user-menu.jsx

wlach: Yes. I don't think the changes have any functional impact, certainly they don't have anything to do with the issue at hand? Unless there was some reason you proposed them that I don't understand?

iFlameing: ok then I open another issue regarding this. when we are clicking the loginof header message, notebook header doesn't showing the avatar of user.

wlach: I'm not sure if I completely understand but definitely file an issue if you noticed another problem, perhaps with a screenshot demonstrating the issue.

iFlameing:
@wlach  can you now merge this ?

wlach:
Yes, this looks good. I tweaked the message jsx a small bit and updated the commit message. Thanks!

PR #1312, opened 2019-01-03
Disabling logout button for the user which uses oAuth for login

iFlameing:
#Fixes1205
This is  just for starting on this issue. To know whether I am in right direction or not.

### DIFF
--- a/src/shared/user-menu.jsx
+++ b/src/shared/user-menu.jsx
@@ -1,3 +1,5 @@
+/* global USE_OPENIDC_AUTH */
+
 import React from 'react'
 import { connect } from 'react-redux'
 import styled from 'react-emotion';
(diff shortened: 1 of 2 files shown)


src/shared/user-menu.jsx
```diff
@@ -128,9 +134,7 @@ export default class UserMenu extends React.Component {
                         Go to Profile
                       </MenuItem>
                       <MenuDivider />
-                      <MenuItem onClick={this.logout}>
-                        Log Out
-                      </MenuItem>
+                      {logout}
```
wlach: I'd prefer an approach like:

```jsx
{
  !USE_OPENIDC_AUTH && (
    <MenuItem onClick={this.logout}>
       Log Out
    </MenuItem>
  )
}
```

This is more direct, understandable, and compact.

PR #1318, opened 2019-01-04
Remove the wrapper element around the popover (fixes #1317)

### LINKED ISSUE #1317: Notebook options popup can be incorrectly positioned
Not sure if this is a new issue, or if it came from my changes in #1302, but I just noticed the popup for the notebook options on the user page gets incorrectly positioned if you scroll sufficiently far down:

image

This appears to be an issue with the fixed position wrapper element we have around the popover. I think the easiest fix is just to remove it, since it actually appears not to be required.

wlach:
We don't actually need it, and the use of `style: fixed` inside
of it caused it to be incorrectly positioned when it was triggered
by an element far down on a page which didn't fit into the viewport.


src/shared/components/popover.jsx
```diff
@@ -40,21 +40,7 @@ class OutsideClickBoundary extends React.Component {
...
 
   render() {
-    return (
-      <div
-        ref={(el) => { this.container = el }}
-        style={{
-          width: '100%',
-          height: '100%',
-          position: 'fixed',
-          zIndex: 1000,
```
hamilton: I'm with you that the position: fixed part is not necessary, but the z-index does need to be set to something like 1000. Otherwise we face this problem:

screen shot 2019-01-07 at 9 39 37 am

This can be fixed elsewhere, probably.

wlach: Ah ok, maybe we do need the wrapper element after all then. I think we can use `position: absolute` here to not disturb the positioning. That will also create a new stacking order, dispensing with the need for zIndex:

https://developer.mozilla.org/en-US/docs/Web/CSS/position

PR #1320, opened 2019-01-06
Refactoring the code related to authentication

### LINKED ISSUE #1313: Consolidate login code
We currently have two definitions of login/logout code inside iodide:

1. One initiated from the header bar in actions.js: 

https://github.com/iodide-project/iodide/blob/1718d634e47c08a987fa3e37586c6613b7b783d1/src/actions/actions.js#L270
https://github.com/iodide-project/iodide/blob/1718d634e47c08a987fa3e37586c6613b7b783d1/src/components/menu/header-messages.jsx#L105

2. Another initiated from the user menu:

https://github.com/iodide-project/iodide/blob/1718d634e47c08a987fa3e37586c6613b7b783d1/src/shared/user-menu.jsx#L55

It would be nice to consolidate the login/logout code inside a single file, maybe `src/shared/actions/. Obviously some parts of the business logic in the header and user menu would have to stay there, but the core pieces of interacting with the server should be shareable.

iFlameing:
Fixes #1313

### DIFF
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -3,6 +3,7 @@ import { getUrlParams, objectToQueryString } from '../tools/query-param-tools'
 
 import { getNotebookID } from '../tools/server-tools'
 import { clearAutosave, getAutosaveJsmd, updateAutosave } from '../tools/autosave'
+import { loginToServer, logoutFromServer } from '../tools/login'
 
 import { mirroredStateProperties } from '../state-schemas/mirrored-state-schema'
 
@@ -268,13 +269,9 @@ export function loginFailure() {
 }
 
 export function login(successCallback, failCallback) {
-  const url = '/oauth/login/github'
-  const name = 'github_login'
-  const specs = 'width=500,height=600'
-  const authWindow = window.open(url, name, specs)
-  authWindow.focus()
+  loginToServer()
   return (dispatch) => {
-    // Functions to be called by child window
+    // Functions to be called by child window set up by loginToServer()
     window.loginSuccess = (userData) => {
       dispatch(loginSuccess(userData))
       if (successCallback) successCallback(userData)
@@ -285,15 +282,19 @@ export function login(successCallback, failCallback) {
     }
   }
 }
+
+function logoutSuccess(dispatch) {
+  dispatch({ type: 'LOGOUT' })
+  dispatch(updateAppMessages({ message: 'Logged Out' }))
+}
+
+function logoutFailure(dispatch) {
+  dispatch(updateAppMessages({ message: 'Logout Failed' }))
+}
+
 export function logout() {
   return (dispatch) => {
-    fetch('/logout/')
-      .then((response) => {
-        if (response.ok) {
-          dispatch({ type: 'LOGOUT' })
-          dispatch(updateAppMessages({ message: 'Logged Out' }))
-        } else dispatch(updateAppMessages({ message: 'Logout Failed' }))
-      })
+    logoutFromServer(logoutSuccess, logoutFailure, dispatch)
   }
 }
 
--- a/src/shared/user-menu.jsx
+++ b/src/shared/user-menu.jsx
@@ -14,6 +14,7 @@ import Menu from './components/menu'
 import MenuItem from './components/menu-item'
 import MenuDivider from './components/menu-divider'
 import Popover from './components/popover'
+import { logoutFromServer, loginToServer } from '../tools/login'
 
 
 const AvatarButtonContainer = styled('div')`
@@ -35,6 +36,7 @@ export default class UserMenu extends React.Component {
     this.goToProfile = this.goToProfile.bind(this)
     this.login = this.login.bind(this)
     this.logout = this.logout.bind(this)
+    this.logoutSuccess = this.logoutSuccess.bind(this)
     this.showLoginModal = this.showLoginModal.bind(this)
     this.hideLoginModal = this.hideLoginModal.bind(this)
   }
@@ -50,6 +52,14 @@ export default class UserMenu extends React.Component {
 
   showLoginModal() {
     this.setState({ loginModalVisible: true })
+  }
+
+  logoutSuccess() {
+    if (this.props.refreshOnLoginLogout) {
+      window.location.reload()
+    }
+    // else...
+    this.setState({ isLoggedIn: false })
   }
 
   login() {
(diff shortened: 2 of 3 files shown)


iFlameing:
@wlach I refactor code related to authentication but one problem occurred during this is when we clicking to Go To Profile it giving store error I didn't understand why ?

src/shared/user-menu.jsx
```diff
@@ -173,7 +112,19 @@ function mapStateToProps(state) {
...
+}
+
+function mapDispatchToProps(dispatch) {
+  return {
+    login: () => {
+      dispatch(login())
+    },
+    logout: () => {
+      dispatch(logout())
+    },
```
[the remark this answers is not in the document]
iFlameing: @wlach  we can make the common code for login not for logout because in response of logout we are doing different things in src/actions/action.js (we are doing dispatch things) where as in user-menu.jsx( we are doing setState things ). Any suggestions regarding this.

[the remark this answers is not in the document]
wlach: I think by providing a callback function you can do what you need here. E.g.:

```js
function logout(successCallback, errorCallback) {
  fetch('/logout/').then(response => {
    if (response.ok) {
      successCallback()
      return
    } 
    return errorCallback()
  });
}
```

There might be an even better way of doing this, but I think this will work.

iFlameing:
@wlach  can you review this once again.

src/shared/user-menu.jsx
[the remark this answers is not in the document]
iFlameing: thanks :)

src/actions/actions.js
```diff
@@ -285,15 +282,19 @@ export function login(successCallback, failCallback) {
...
 export function logout() {
   return (dispatch) => {
-    fetch('/logout/')
-      .then((response) => {
-        if (response.ok) {
-          dispatch({ type: 'LOGOUT' })
-          dispatch(updateAppMessages({ message: 'Logged Out' }))
-        } else dispatch(updateAppMessages({ message: 'Logout Failed' }))
-      })
+    Logout(logoutSuccess(dispatch), logoutFailure(dispatch))
```
wlach: I'm pretty sure this doesn't work, it looks like you're actually calling the functions here. Instead, we want to pass a reference to the functions so the logout function can call them itself.

src/shared/user-menu.jsx
```diff
@@ -52,6 +53,20 @@ export default class UserMenu extends React.Component {
...
+  logoutSuccess() {
+    if (this.props.refreshOnLoginLogout) {
+      window.location.reload()
+    }
+    // else...
+    this.setState({ isLoggedIn: false })
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  logoutError(response) {
```
wlach: It is ok just to use an anonymous function when something is so short, instead of disabling eslint.

e.g.

```js
// FIXME: we should handle the logout failure case somehow (e.g. by popping up a notification)
logout(this.logoutSuccess, (response) => console.error('Logout unsuccessful', response))
```

src/tools/login.js
```diff
@@ -0,0 +1,17 @@
+export function Login() {
```
wlach: The first letter of a function name should not be capitalized.

iFlameing: I am calling Login functin in login therefore I capitalized the first letter.

wlach: I see. I think I'd prefer you named these functions something like `loginToServer` and `logoutFromServer` instead of capitalization (I'm not sure why that isn't triggering a linter error).

src/actions/actions.js
iFlameing: I am also thinking of . Somehow it doesn't give me error while I tested in my local environment . I should change this.

src/tools/login.js
iFlameing: ok

src/actions/actions.js
iFlameing: @wlach I amend some changes regarding this can you look into it ?

iFlameing:
@wlach I test it locally and it is working fine and I also check the action logout function using redux_verbose npm start .

src/shared/user-menu.jsx
```diff
@@ -52,6 +54,15 @@ export default class UserMenu extends React.Component {
     this.setState({ loginModalVisible: true })
   }
 
+  // eslint-disable-next-line no-unused-vars
```
wlach: Please don't disable eslint rules unless you have an extremely good reason for doing so. If you find that eslint is warning it's generally a sign that there is a problem with your code or it is more verbose than necessary-- in this case there is no need to have an unused dispatch parameter and you can just leave it out.

src/shared/user-menu.jsx
```diff
@@ -86,26 +92,15 @@ export default class UserMenu extends React.Component {
...
-            }
-            // else...
-            this.setState({ isLoggedIn: false })
-          } else {
-          // do something smart here (probably pop up a notification)
-            console.error('Login unsuccessful', response)
-          }
-        });
+      // eslint-disable-next-line no-unused-vars
+      logoutFromServer('', this.logoutSuccess, (response, dispatch) => console.error('Logout unsuccessful', response))
```
wlach: Same as above

src/actions/actions.js
```diff
@@ -268,11 +269,7 @@ export function loginFailure() {
...
 
 export function login(successCallback, failCallback) {
-  const url = '/oauth/login/github'
-  const name = 'github_login'
-  const specs = 'width=500,height=600'
-  const authWindow = window.open(url, name, specs)
-  authWindow.focus()
+  loginToServer()
   return (dispatch) => {
     // Functions to be called by child window
```
wlach: Modify the comment here to make it more clear that the child window is created by loginToServer.

```js
// Functions to be called by child window set up by loginToServer()
```

src/tools/login.js
```diff
@@ -0,0 +1,17 @@
+export function loginToServer() {
+  const url = '/oauth/login/github'
+  const name = 'github_login'
+  const specs = 'width=500,height=600'
+  const authWindow = window.open(url, name, specs)
+  authWindow.focus()
+}
+
+export function logoutFromServer(dispatch, successCallback, errorCallback) {
```
wlach: I would make `dispatch` the last parameter, that way you can just omit it in cases where it does not apply (i.e. `user-menu.jsx`)

src/shared/user-menu.jsx
iFlameing: ok

iFlameing:
@wlach can you merge this now ?

iFlameing:
build is failing due to circleci issue

PR #1327, opened 2019-01-08
Styling explore button

### LINKED ISSUE #984: "EXPLORE" button should be more prominent in report view (funnel optimization)
we want to make the "EXPLORE" button more prominent and visible so that it doesn't get overlooked during a user's first encounter with iodide

iFlameing:
Fixes #984 
Now explore button look like this.
screenshot from 2019-01-08 21-07-06

@bcolloran can you make comment on this ?

### DIFF
--- a/src/components/menu/view-mode-toggle-button.jsx
+++ b/src/components/menu/view-mode-toggle-button.jsx
@@ -33,7 +33,7 @@ export class ViewModeToggleButtonUnconnected extends React.Component {
         title={this.props.tooltipText}
       >
         <Button
-          style={{ color: this.props.textColor }}
+          style={{ color: this.props.textColor, ...this.props.backgroundColor }}
           onClick={this.toggleViewMode}
           variant="text"
           mini
@@ -53,6 +53,7 @@ export function mapStateToProps(state) {
     textColor: isReportView ? 'black' : '#fafafa',
     buttonText: isReportView ? 'Explore' : 'Report',
     tooltipText: isReportView ? 'Explore this notebook' : 'Go to Report view',
+    backgroundColor: isReportView ? { backgroundColor: '#eee', border: '1px solid #ccc' } : '',
   }
 }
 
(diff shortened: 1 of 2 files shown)


bcolloran:
fine by me -- @hamilton any strong feelings about the direction here? (we can revise later when we have more design cycles, but i think this accomplishes the funnel objective of making the explore button a slightly more visible call to action)

bcolloran:
@iFlameing please rebase this PR on the changes made in https://github.com/iodide-project/iodide/pull/1300 -- you should now be able to incorporate these styling tweaks in the mapStateToProps

iFlameing:
@bcolloran I rebased the pr as you said  and updated the test file also.

src/components/menu/view-mode-toggle-button.jsx
```diff
@@ -53,6 +53,7 @@ export function mapStateToProps(state) {
     textColor: isReportView ? 'black' : '#fafafa',
     buttonText: isReportView ? 'Explore' : 'Report',
     tooltipText: isReportView ? 'Explore this notebook' : 'Go to Report view',
+    backgroundColor: isReportView ? { backgroundColor: '#eee', border: '1px solid #ccc' } : '',
```
bcolloran: is `backgroundColor` an appropriate name for a prop that contains more that one style? Do you notice that we're passing we're now passing several props that style one element? What does that suggest about how the code could be simplified?

iFlameing: @bcolloran  we can remove both props textColor , backgroundColor and merge them into style props which take care how button is render in reportview and exploreview what you think ?

bcolloran: that sounds like a correct simplification @iFlameing -- try to pay attention for opportunities like that to simplify code and consolidate little things that are unnecessarily spread out. Also, always be careful about what you name variables, and make sure the name matches the data the variable holds. Paying careful attention to little details like this will help you to take your coding skills to the next level! :-)

One tip that I try to follow: when you are about to submit a PR, before hitting the 'submit' button, take a few minutes to review the diffs of everything you've changed in the "files changed" tab, and make sure that you haven't missed optimizations like this. Recheck variable names, make sure you don't have extra code you don't need, stuff like that. The diff view will help you to spot your changes and make sure they are what you want

iFlameing: These points are very helpful. Thanks for sharing these. I will careful from the next time and implement these points in my further pr. :)

PR #1332, opened 2019-01-08
Don't say we successfully saved the notebook when we didn't (fixes #1322)

### LINKED ISSUE #1322: When not logged in, saving the notebook (via meta-s) succeeds, even though it really failed

wlach:


wlach:
image

bcolloran:
hmmm, so the snackbar part of this works ok -- when i log out, it does show the snackbar with the error msg as expected... but it also shows "saved just now" in the header bar. i wonder if we want to change the routing of the events here such that `state.lastSaved` is only updated after a successful save on the server? i'm fine if you want to punt on that for now and land this PR to move the ball forward, but if you don't want to get that with this PR plz add an issue for that. thx!!

bcolloran:
oh also the code is simple and r+

wlach:
let's try and fix the saved state while we're here.

wlach:
Was actually quite easy to change it so the "last saved" state wouldn't be updated on failure. Going to land now.

PR #1360, opened 2019-01-09
Re-upgrade babel-loader to 8.0.5

wlach:


wlach:
This doesn't fix the problem

PR #1364, opened 2019-01-10
remove re-resizable, add a plugin that warns about unused files in src

bcolloran:
more repo hygiene:
1) i removed the re-resizable module (replaced with golden layout months ago)
2) added a plugin which warns about files that are not imported anywhere. it has so far turned up 20 files:
```
20 unused source files found.

● src
    • actions/__tests__/jsmd-parser-test.js
    • actions/scroll-helpers.js
    • components/controls/editor-button.jsx
    • components/menu/console-icon.js
    • components/menu/notebook-menu-divider.jsx
    • components/menu/notebook-menu-header.jsx
    • components/menu/notebook-menu-subsection.jsx
    • components/modals/__tests__/keyboard-shortcut-list-test.js
    • components/pane-layout/layout-config-explain.js
    • components/pane-layout/layout-config-explore.js
    • components/reps/__tests__/error-handler.jsx
    • components/reps/__tests__/promise-handler.test.jsx
    • eval-frame/components/controls/editor-link-button.jsx
    • eval-frame/html-template.js
    • eval-frame/tools/nb.js
    • iodide-examples.js
    • server/components/delete-element-button.jsx
    • server/style/ZillaSlab-Light.woff
    • server/user.jsx
    • tools/create-session-id.js
```
we can delete these in follow up PRs


package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "^7.0.0"
+        "@babel/highlight": "7.0.0"
```
wlach: I notice that we keep on switching back and forth between requiring "^VERSION" and "VERSION" for sub-dependencies. I don't think this makes much practical difference, but it feels like unnecessary churn.

What version of npm are you using? The latest version is 6.5.0, if `npm  -v` is giving you a lower one, try running `npm install npm@latest -g`

@hamilton @mdboom @iFlameing -- this would be a good idea for you to do too.

hamilton: yep, I'm on 6.5.0. Migrated to 6.x when I had to resolve a package-lock situation down the line. I think it'd be good to (1) get everyone on the same eveyrthing, and (2) make sure that the "same everything" is put in the contrib docs

hamilton: yep, I'm on 6.5.0. Migrated to 6.x when I had to resolve a package-lock situation down the line. I think it'd be good to (1) get everyone on the same eveyrthing, and (2) make sure that the "same everything" is put in the contrib docs

bcolloran: @wlach i think may also be because renovate actually tries to pin modules at specific versions (see PRs like eg https://github.com/iodide-project/iodide/pull/898 ), the theory being that if everything is pinned to a specific version, than different devs won't see different behavior if different modules within a range are pulled. And since renovate is watching for version changes and submitting PRs, that mechanism can be used to ensure we're getting minor releases rather than relying on npms `~` and `^`.

but in any case, def good to have everyone on the latest npm

PR #1375, opened 2019-01-11
Run jest tests sequentially (fixes #1374)

### LINKED ISSUE #1374: test-javascript job on circle is failing
For the last day or so the `test-javascript` job on circle has been failing intermittently (but more often than not). It appears to be timing out.

According to the resources below, it seems like the setup for running jest is optimized for workstations, not ci. Apparently in the latter environment is recommended that one run tests sequentially, not in parallel. We don't have all that many tests in iodide, so we might be able to just get away with making that the default:

https://discuss.circleci.com/t/babel-7-jest-results-in-timeout/25383
https://jestjs.io/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server

wlach:


wlach:
I think I'm just going to go ahead and merge this, as it's a bit of a no-brainer.

PR #1376, opened 2019-01-11
Use prettier's lint settings for code (fixes #1199)

### LINKED ISSUE #1199: Consider using prettier to lint/format code
Some of the spacing issues in #1197 suggest to me that using the airbnb linter isn't quite enough to catch all the nits that one might have about how a react file is formatted.

IMO we should just switch to using [prettier](https://prettier.io/), which you can configure to "auto format" your code so you don't even have to worry about violating our conventions/style. I believe it's mostly the same as the best practices we currently have, the one exception being that it requires/adds semi-colons to the end of lines. This is easy to disable with the the `--no-semi` option though (I used to feel like using semicolons was the one true way to write javascript, I've since softened my stance).

@hamilton @bcolloran @mdboom thoughts?

wlach:
This also updates all code to comply with the rules


wlach:
I opted just to use prettier's defaults, even where they differed from what we're doing currently. Some notable differences from the way we do things right now:

* Expressions are now always terminated by semicolons
* We now use double quotes instead of single quotes for most things
* Multi-line lists no longer have trailing commas

We could override all of these easily if desired, does anyone have strong feelings one way or the other? Looking at the new code, I don't really have an opinion one way or the other on whether these things matter. I'm mostly just looking forward to not having to critique formatting issues in PRs. :)

I would like to merge this by early Monday, if possible, so I don't have to rebase/rework things too many times.

bcolloran:
thanks for doing this will!

PR #1380, opened 2019-01-14
Cache results of pyodide compression when creating heroku buildpack (fixes #1338)

### LINKED ISSUE #1338: iodide server deployment is super slow
The iodide server takes quite a while to build on heroku (15-30 mins?), the main culprit seems to be compressing the scipy.data file from pyodide (which is 190 megs big) to brotli. This makes testing changes a bit annoying.

I think we should be able to cache the results of the whitenoise step using this technique:

https://devcenter.heroku.com/articles/buildpack-api#caching

... but we need to be careful to make sure we're not using stale assets. The best thing would probably be to just cache the pyodide assets specifically, and key that on the currently used version.

/cc @mdboom

wlach:
Have tested this a bunch on heroku, and it seems to work well.

### DIFF
--- a/bin/post_compile
+++ b/bin/post_compile
@@ -8,11 +8,30 @@ set -euo pipefail
 # Uses the node binaries/packages installed by the nodejs buildpack previously.
 npm install && npm run build
 
+# We compress files in `build/` below so they can be served more quickly
+# but we want to cache the results for pyodide as it changes relatively
+# infrequently and compressing it to brotli is sloooooow
+# see: https://devcenter.heroku.com/articles/buildpack-api#caching
+PYODIDE_CACHE_DIR=$CACHE_DIR/pyodide-$PYODIDE_VERSION
+if [ ! -d $PYODIDE_CACHE_DIR ]; then
+   echo "pyodide cache does not seem to exist yet, downloading and compressing"
+   rm -rf $CACHE_DIR
+   mkdir -p $CACHE_DIR
+   mv build/pyodide-$PYODIDE_VERSION $CACHE_DIR
+   python -m whitenoise.compress $PYODIDE_CACHE_DIR
+else
+   echo "Using cached version of pyodide assets"
+   rm -rf build/pyodide-$PYODIDE_VERSION
+fi
+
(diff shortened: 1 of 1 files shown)


bin/post_compile
```diff
@@ -8,12 +8,29 @@ set -euo pipefail
 # Uses the node binaries/packages installed by the nodejs buildpack previously.
 npm install && npm run build
 
+# We compress files in `build/` below so they can be served more quickly
+# but we want to cache the results for pyodide as it changes relatively
+# frequently and compressing it to brotli is sloooooow
```
mdboom: Do you mean "infrequently"?  (We only update the Pyodide version very few weeks...)

bin/post_compile
```diff
@@ -8,12 +8,29 @@ set -euo pipefail
...
+PYODIDE_DIR=$CACHE_DIR/pyodide-$PYODIDE_VERSION
```
mdboom: Where is `CACHE_DIR` defined?  Is that just something that Heroku defines?

mdboom:
> One thing to note is that this change requires the environment variable PYODIDE_VERSION to be specified on heroku. I think that's ok.

Which basically implies we have to run everything through the webpack script?  (Just a sanity to check to confirm we don't need to document that fact somewhere).

wlach:
> > One thing to note is that this change requires the environment variable PYODIDE_VERSION to be specified on heroku. I think that's ok.

> Which basically implies we have to run everything through the webpack script? (Just a sanity to 
> check to confirm we don't need to document that fact somewhere).

No, this qualification is applicable only for heroku. Maybe we should add a small section on our heroku deployment to the docs -- there's a few idiosyncracies to it which are probably worth writing down.

bin/post_compile
wlach: Yup. https://devcenter.heroku.com/articles/buildpack-api#caching 

I'll put a note about that in the comment

bin/post_compile
wlach: Yup, good catch, will fix.

PR #1389, opened 2019-01-15
Fix #1148: Document how to delete an account.

### LINKED ISSUE #1148: Add documentation on removing record of user's account from iodide site
We should add some way for users to delete their account from our site.

From Michael Feldman:

```
For GDPR compliance, we'll need a way to be able to delete our storage of user data if they ask for it. Are we storing any of the data on our servers, or is it all on Github, so that users have control over it?
```

I think this only makes sense on the public site, so this feature should be enabled only with a specific configuration flag.

mdboom:


src/shared/login-modal.jsx
```diff
@@ -37,6 +37,11 @@ export default class LoginModal extends React.Component {
             &nbsp; applies to your content. Your notebooks are associated with
             your account.
           </p>
+          <p>
+            Deleting a notebook permanently deletes its contents. If you wish to
```
bcolloran: Do we need to add some nuance here about other users forking copies of a notebook? What about revisions?

bcolloran: @wlach

mdboom: I thought about that, but then thought it was self-evident.  But maybe not.  "Deleting a notebook permanently deletes any copies and revisions in your account."?

mdboom:
Note from meeting: This should be moved to the docs.

bcolloran:
hey guys, i just threw together this:
https://github.com/iodide-project/iodide/blob/master/docs/policies.md

porting over @mdboom 's language (thanks mike!!) to a docs page. should be good to go, so i'll close this one.

PR #1394, opened 2019-01-17
Add redirects

mdboom:
This adds redirects on the server that go to GH pages, only when installed on `iodide.io` for the following:

- `dist/`: The old released versions of standalone (serverless) iodide, so as to not break existing notebooks on the internet
- `docs/`: This is location of the new markdown-based docs
- `pyodide-demo/`: Not strictly necessary, but it buys me some time to update links from blog posts and presentations

I'm not sure if the `SITE_URL` value is exactly correct, as I couldn't find where the configuration for our Heroku deployment is...  I did test it in a local server and changing it to `http://localhost:8000/` makes it work locally.

### DIFF
--- a/server/urls.py
+++ b/server/urls.py
@@ -1,10 +1,31 @@
+from django.conf import settings
 from django.conf.urls import include, url
 from django.contrib import admin
 from django.urls import path
+from django.views.generic.base import RedirectView
 
 import server.views
 
 admin.autodiscover()
+
+
+def parse_redirects(redirects):
+    """
+    Parses a string defining a set of redirects.
+
+    The string is in ;-delimited section, where each section is
+    if the form `$prefix=$dest`.
+
+    Yields Django url configurations to perform the redirect.
+    """
+    redirects = [x.strip() for x in redirects.split(';')]
+    for redirect in redirects:
+        parts = [x.strip() for x in redirect.split('=')]
+        if len(parts) != 2:
+            continue
+        prefix, dest = parts
+        yield url(f'^{prefix}.*', RedirectView.as_view(url=dest))
+
 
 urlpatterns = [
     # notebook stuff
@@ -26,7 +47,11 @@ urlpatterns = [
     path('admin/', admin.site.urls),
 
     url(r'^$', server.views.index, name='index'),
+]
 
+urlpatterns += list(parse_redirects(settings.IODIDE_REDIRECTS))
+
+urlpatterns += [
     # user urls
     # based on https://github.com/shinnn/github-username-regex
     url(r'^(?P<name>\w(?:\w|-|@|\.(?=\w)){0,38})/$', server.views.user, name='user'),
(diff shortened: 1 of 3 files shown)


server/urls.py
```diff
@@ -26,8 +28,21 @@
     path('admin/', admin.site.urls),
 
     url(r'^$', server.views.index, name='index'),
+]
+
+if settings.SITE_URL == 'https://iodide.io/':
```
wlach: Would this be extremely-alpha.iodide.io?

How about defining an environment variable with the following format:

`<pathspec>=<redirect>;<pathspec>=<redirect>;...`

in this case:

`dist/.*=https://iodide-project.github.io/dist/;docs/.*=https://iodide-project.github.io/docs/;pyodide-demo/.*=https://iodide-project.github.io/pyodide-demo/`

This is less iodide.io specific, and would allow us to configure/tweak the exact behaviour inside the heroku environment.

mdboom: > Would this be extremely-alpha.iodide.io?

This came out of a discussion of moving what's currently at `extremely-alpha.iodide.io` to `iodide.io` without blitzing what's currently at `iodide.io`.

> How about defining an environment variable with the following format.

Sure, that's easy enough.  I'll update the PR.

wlach:
Looks good now! Looks like there is a small flake8 error which should be addressed before this lands.

mdboom:
Where is the `flake8` error?  `make flake8` is passing for me...

mdboom:
Oh, cool.  The flake8 error should be resolved now.

server/urls.py
mdboom: Oh, that's a good idea.  But I'm not sure how we would implement it.  Currently Github pages handles `iodide.io`, and I don't think there's a way to set up dynamic redirects there.

But you're right -- this whole PR is on the assumption that `iodide.io` would eventually point to our server on Heroku, which may not have been the right assumption.

bcolloran:
@mdboom @wlach has this been superceded, or is it still relevant? please merge/close at your discretion, thanks!

mdboom:
I think it's no longer possible.  I dropped the ball on keeping `dist` around and I don't think we can do it (at least not in this way) anymore now that `iodide.io` is a redirect to `extremely-alpha.iodide.io`.  Oh well, forging on into the future we go!

rth:
If I understand correctly the pyodide-demo currently has issues loading (cf https://github.com/iodide-project/pyodide/issues/311) due to https://github.com/iodide-project/iodide/pull/1452 ?

Also there seem to be a double HTTPS -> HTTP -> HTTPS redirect (see first linked issue).

rth:
Nevermind, I see this hasn't been merged. I guess my previous comments apply to,

> now that iodide.io is a redirect to extremely-alpha.iodide.io

but I couldn't find the relevant issue...

mdboom:
Yes, the pyodide-owned demos are broken by the moving of URLs.  We'll have to either restore some URLs or update the demos to get them working again.

Issue #1330, opened 2019-01-08
there should be some way for non-logged in users to get to a fresh template notebook


bcolloran:

wlach:
Still not sure in my head exactly what the right technical implementation we want here is, but I think the behaviour we want is to have some big button on iodide's front page which leads the user directly to a blank notebook that's easy to start experimenting and playing with. To cover all the bases, maybe it's best to include both python *and* javascript in it. As we hinted at in our meeting, maybe the best thing is to make this some notebook on our server, and we can link to it by id in `server/settings.py` and render if present in the splash view that @hamilton is looking at.

@hamilton @bcolloran do you have any notebooks that would fit the bill here or should we create one?

wlach:
Rethought this a bit.

tentative new plan: 

* add a `/tryit/` endpoint which loads the template notebook for logged-out users (auto-redirect to /new/ for logged in users). This experience will be more impressive once @bcolloran lands #1384.
* while the user is editing, they will be warned that the document will not be saved until they log in
* if the user signs in, autosave the notebook from the client side.

May need a few more client-side bits to make this interaction smooth.

I have the endpoint working but the client side still needs a bit more work.

bcolloran:
#1384 is done

PR #1398, opened 2019-01-18
Refactor some saving code to reduce duplication

wlach:


wlach:
On reflection maybe I should have incorporated this into a larger PR for #1330, but I figured this would be a bit faster to review on its own.

bcolloran:
@wlach lgtm, merge when you wish!

Issue #1397, opened 2019-01-18
Allow option of "posting" to /new endpoint with a pre-filled jsmd document


wlach:
Right now we have a `/new` endpoint which will auto-create a notebook based on a template inside the iodide repository:

https://github.com/iodide-project/iodide/blob/e2ee77fdde96210eddb29ea756964aeb0af9afcd/server/notebooks/views.py#L103
https://github.com/iodide-project/iodide/blob/e2ee77fdde96210eddb29ea756964aeb0af9afcd/server/notebooks/templates/new_notebook_content.jsmd

What we want to do here is optionally allow posting to the above endpoint with a json payload containing something like `{ 'jsmd': <my cool iodide notebook> }`. If that's the case and the user is logged in, autocreate the notebook using that payload instead of the default. If the user is *not* logged in, we should redirect to the `/tryit/` endpoint with the same payload (which should do exactly the same thing, but not actually create the notebook) -- this depends on #1330 being implemented (hoping to get that done today).

So the initial use case for this is to facilitate exploration of redash queries in iodide (mozilla/redash#426), but there are probably other cases where this might be useful. 

/cc @jezdez @openjck

bcolloran:
Thanks for filing @wlach. @openjck so excited to have you joining in! :-) :100:

openjck:
Question about the JSON payload. Is there an existing format that it should adhere to or is defining that format part of this work?

wlach:
> Question about the JSON payload. Is there an existing format that it should adhere to or is defining that format part of this work?

No existing format. To start with, something simple like this should work:

```json
{
    "name": "My cool notebook",
    "jsmd": "%% md\nFoo bar\n..."
}
```

openjck:
Okay, great!

Another question as I think about this: does Iodide support private notebooks? Given that the data on Redash is private, do we want to implement that feature as part of this work, or we were thinking about addressing this some other way, like setting up a different instance of Iodide behind LDAP?

wlach:
> Another question as I think about this: does Iodide support private notebooks? Given that the data 
> on Redash is private, do we want to implement that feature as part of this work, or we were thinking > about addressing this some other way, like setting up a different instance of Iodide behind LDAP?

Well, I wouldn't say the data on redash is private *by definition*, we sometimes share portions of it with the public using the API (chutten's crash dashboard would be a previous example of this, there might be others). 

[...]

PR #1399, opened 2019-01-18
Add a "try it out" mode to iodide (fixes #1330)

### LINKED ISSUE #1330: there should be some way for non-logged in users to get to a fresh template notebook

wlach:
This is similar in behaviour to the `/new/` endpoint, except it doesn't
save the notebook. If the user later decides to log in, their changes
will automatically be saved to their account.

If a logged in user visits this endpoint, they will be redirected to
`/new/`.

### DIFF
--- a/src/tools/server-tools.js
+++ b/src/tools/server-tools.js
@@ -28,6 +28,10 @@ export function connectionModeIsServer(state) {
   return getConnectionMode(state) === "SERVER";
 }
 
+export function notebookIsATrial(state) {
+  return state.notebookInfo.tryItMode;
+}
+
 export function getNotebookID(state) {
   if (!connectionModeIsServer(state)) return undefined;
   const notebookID = state.notebookInfo.notebook_id;
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -111,3 +111,26 @@ def new_notebook_view(request):
             title='Untitled notebook'
         )
     return redirect(notebook)
+
+
+@ensure_csrf_cookie
+def tryit_view(request):
+    '''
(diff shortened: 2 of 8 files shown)


wlach:
So right now if you're not logged in and you click the "try it out" button, you'll get the current "new" notebook:

image

https://github.com/iodide-project/iodide/blob/master/server/notebooks/templates/new_notebook_content.jsmd

I feel like for new users, @hamilton's more fleshed out https://extremely-alpha.iodide.io/notebooks/154/ notebook (or something like it) might be a better default choice, at least for that particular option in the splash. But this would probably still be useful for something like #1397 for cases that users are "opening something in iodide".

wlach:
Another cool option might be to link to the more extended tutorial inside the default template notebook. I think I like that option best actually.

In any case, I think we can merge this for now and refine later.

src/tools/server-tools.js
```diff
@@ -28,6 +28,10 @@ export function connectionModeIsServer(state) {
   return getConnectionMode(state) === "SERVER";
 }
 
+export function notebookIsATrial(state) {
+  return state.notebookInfo.tryItMode;
```
bcolloran: `tryItMode` should be added to the state schema
https://github.com/iodide-project/iodide/blob/ae55936a5b9250215f41573cda2a69a68212253a/src/state-schemas/state-schema.js#L185

wlach:
@bcolloran it's not exactly the same, as the tryit endpoint uses the template notebook, not a created one.

My gut tells me a custom view is the right way here, but we can always revisit in the future. :) Good catch on the state schema, I added tryItMode to that. Will merge once ci passes.

PR #1409, opened 2019-01-22
Add a table of supported languages

mdboom:
This at least gets this started, and we can start filling in more options and details later.


bcolloran:
rad, thanks mike!

PR #1410, opened 2019-01-22
fetch cells no longer mysteriously block on error

### LINKED ISSUE #1372: fetch cell errors do not un-halt the evaluation queue, so nothing else can be evaluated after error
@hamilton plz, take a look at:
https://extremely-alpha.iodide.io/notebooks/210/

i reckon since you have the most experience with the fetch cell and evaluation machinery you should be able to make quick work of this one, but lmk if you want me to take a look

hamilton:
fixes #1372. Not sure how this was introduced as a bug, but something was preventing sending the evaluation response on syntax errors in fetch chunks.


bcolloran:
looks like a simple fix! thanks!

PR #1420, opened 2019-01-23
Make docs match new style of site

mdboom:
This updates the CSS of the docs to better match @hamilton's recent changes to the site and UI.

### DIFF
--- a/docs/extra.css
+++ b/docs/extra.css
@@ -17,13 +17,13 @@ body::before {
 }
 
 h1, h2, h3, h4, h5 {
-  font-family: 'Zilla Slab', sans-serif;
   font-weight: 300;
 }
 
 body {
-  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica,
-               Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
-               "Segoe UI Symbol";
+  --main-typeface: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica,
+                   Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
+                   "Segoe UI Symbol";
+  font-family: var(--main-typeface);
   font-weight: 400;
 }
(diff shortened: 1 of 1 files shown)


docs/extra.css
```diff
@@ -1,26 +1,29 @@
...
 
 .navbar-brand {
   font-family: 'Zilla Slab', serif;
+  font-size: 30px;
+  font-weight: 300;
 }
 
 h1, h2, h3, h4, h5 {
-  font-family: 'Open Sans', sans-serif;
+  font-family: 'Zilla Slab', sans-serif;
```
hamilton: __More Purple™__! This is awesome. Just a couple of small tweaks I'd recommend:

(1) set a family system font in body:
(2) use the common system fontset for the h* tags, while keeping the zilla slab font for just `navbar-brand`. generally I think this improves readability overall and keeps the font specifically for the logo itself.

All together, those changes would look like:

```css
body {
   --main-typeface: font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
   font-family: var(--main-typeface);
}

h1, h2, h3, h4, h5 {
    font-family: var(--main-typeface);
}
```

bcolloran: i disagree very strongly with using system fonts. the system fonts look nicer on macs than other platforms, and if we use system fonts, you guys _will not see_ what a big chunk of users see, and you'll have a mistaken impression of how good things look. We should design to one consistent standard, and have a non-system backup webfont for cases where zilla slab is not appropriate

hamilton: We use system fonts for all of the iodide properties, don't we? The point is to use the fonts that work best on all platforms, not just mac. Can you provide more information about where they fail? Happy to reassess this strategy but this is not my understanding.

mdboom:
I've updated this to not use Zilla Slab for the headers.  The question of what specifically to use for body text can probably be addressed as part of #1422, as the present PR just brings the docs inline with everything else.

hamilton:
@mdboom looks good, thanks! We can always revisit the system font discussion later and roll with this for now.

Issue #1424, opened 2019-01-23
re-enable "skipRunAll"


bcolloran:
need to re-enable this functionality. this should be an easy fix

PR #1431, opened 2019-01-25
skipRunAll restored

bcolloran:
@hamilton this should fix @teonbrooks skipRunAll issue, #1424  

~~had to refactor a bit to add some tests, but even as such i was not really able to add test cases to everything i would have liked. So please take a moment to kick the tires on this one and test it out locally.~~

~~(and if you feel like taking a crack at adding some tests/getting the non-working tests to run, that would be great too, but i don't think it's absolutely required to have more coverage if the code passes your review and it works as expected during your QA)~~

(did a little more refactoring and now i'm happy with the test coverage, standard review and QA should be fine)

### DIFF
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -228,7 +228,7 @@ const chunkNotSkipped = chunk =>
   );
 
 export function evaluateNotebook(evalQueueInstance = evalQueue) {
-  return () => (dispatch, getState) => {
+  return (dispatch, getState) => {
     const { jsmdChunks, kernelState } = getState();
 
     if (kernelState !== "KERNEL_BUSY") dispatch(setKernelState("KERNEL_BUSY"));
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -216,13 +216,24 @@ export function moveCursorToNextChunk() {
   };
 }
 
+const chunkIsRunnable = chunk =>
+  !["md", "css", "raw", ""].includes(chunk.chunkType);
+
+const chunkNotSkipped = chunk =>
+  !(
+    chunk.evalFlags.includes("skipRunAll") ||
+    chunk.evalFlags.includes("skiprunall")
+  );
+
 export function evaluateNotebook() {
   return (dispatch, getState) => {
     const { jsmdChunks, kernelState } = getState();
+
     if (kernelState !== "KERNEL_BUSY") dispatch(setKernelState("KERNEL_BUSY"));
+
     jsmdChunks.forEach(chunk => {
-      if (!["md", "css", "raw", ""].includes(chunk.chunkType)) {
-        evalQueue.evaluate(chunk, dispatch);
+      if (chunkIsRunnable(chunk) && chunkNotSkipped(chunk)) {
+        evalQueue.evaluate(chunk);
       }
     });
   };
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -227,7 +227,7 @@ const chunkNotSkipped = chunk =>
     chunk.evalFlags.includes("skiprunall")
   );
 
-export function evaluateNotebookUsingQueue(evalQueueInstance) {
+export function evaluateNotebook(evalQueueInstance = evalQueue) {
   return () => (dispatch, getState) => {
     const { jsmdChunks, kernelState } = getState();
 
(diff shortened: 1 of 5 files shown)


src/actions/evaluation-queue.js
```diff
@@ -6,9 +6,17 @@ export class EvaluationQueue {
     this.queueSize = 0;
     this.evaluationResolvers = {};
     this.idGenerator = new IdFactory();
+    this.reduxDispatch = null;
   }
 
-  evaluate(chunk, dispatch) {
+  connectToRedux(dispatch) {
```
hamilton: could this be `connectToStore`? "Redux" is kind of vague

src/actions/actions.js
```diff
@@ -216,18 +216,33 @@ export function moveCursorToNextChunk() {
...
-      if (!["md", "css", "raw", ""].includes(chunk.chunkType)) {
-        evalQueue.evaluate(chunk, dispatch);
+      if (chunkIsRunnable(chunk) && chunkNotSkipped(chunk)) {
+        evalQueueInstance.evaluate(chunk);
       }
     });
   };
 }
 
+export const evaluateNotebook = evaluateNotebookUsingQueue(evalQueue);
```
hamilton: is there a reason to have this actually be a separate function rather than

```javascript

export function evaluateNotebook(evalQueueInstance=evalQueue) {
...
}

```

This would let you still test it the same way without adding this additional layer (which I'm assuming is the point of this refactor)

src/actions/actions.js
```diff
@@ -216,18 +216,33 @@ export function moveCursorToNextChunk() {
...
+export function evaluateNotebookUsingQueue(evalQueueInstance) {
+  return () => (dispatch, getState) => {
     const { jsmdChunks, kernelState } = getState();
+
     if (kernelState !== "KERNEL_BUSY") dispatch(setKernelState("KERNEL_BUSY"));
+
     jsmdChunks.forEach(chunk => {
-      if (!["md", "css", "raw", ""].includes(chunk.chunkType)) {
-        evalQueue.evaluate(chunk, dispatch);
+      if (chunkIsRunnable(chunk) && chunkNotSkipped(chunk)) {
```
hamilton: There might be a way of doing this so we're not adding boilerplate to a function (we'll probably have more flags in the future) but I'm sure this comment will get YAGNI'd and we can always revisit down the line ;)

src/actions/evaluation-queue.js
```diff
@@ -6,9 +6,17 @@ export class EvaluationQueue {
...
+    this.reduxDispatch = dispatch;
+  }
+
+  evaluate(chunk) {
+    if (!this.reduxDispatch) {
+      throw new Error("must connect to redux before evaluating");
```
hamilton: `"must connect to a Redux store before evaluating"`

src/actions/evaluation-queue.js
hamilton: also, this is a great refactor in general

hamilton:
@bcolloran another thing worth thinking about (perhaps as a follow-on) - we should probably surface an iodide console warning (to be tackled when we decouple under the hood the inputs and outputs) saying that a flag was unrecognized.

src/actions/evaluation-queue.js
bcolloran: i'll do `connectDispatch` to match the pattern from review comments #1428

src/actions/actions.js
bcolloran: omfg yes that is so much better! it didn't even occur to me to use a default param!

bcolloran:
Thanks for the review @hamilton! Pushed those updates, ready for another pass

bcolloran:
oh, and agree that we should show a console warning for invalid flags. ideally we'll also surface that via highlighting or something in the editor.

hamilton:
@bcolloran dang, looks like I spoke too soon. I'm having trouble getting "run all" to work on your PR, but this could be a bigger problem w/ my current setup (which I've been neglecting to fix, now's the time). Give me a sec before merging.

bcolloran:
wow, thanks for catching that @hamilton!! i have no idea how i missed that in QA! :-/

should be fixed now. can you confirm+merge?

hamilton:
@bcolloran works great - merging now :tada: /cc @teonbrooks

bcolloran:
duh, i should have added docs!!! you even specifically called this out in the meeting, ack! tough to build new habits...

i'll add some docs now!

PR #1435, opened 2019-01-25
Simplify login code

wlach:
There was quite a bit of state and business logic that was unused
here, and could be eliminated.


src/tools/login.js
```diff
@@ -10,6 +16,7 @@ export function logoutFromServer(successCallback, errorCallback, dispatch) {
   fetch("/logout/").then(response => {
     if (response.ok) {
       successCallback(dispatch);
+      successCallback(dispatch);
```
bcolloran: typo here (duplicated line)

wlach: Oops, I'll fix this now

PR #1436, opened 2019-01-25
Add support for emoji in markdown view

wlach:
To capture the support of the elusive millenial/slack demographic,
we clearly need emoji support in markdown view to keep things :100:.


wlach:
image

PR #1463, opened 2019-01-31
Removing hamburger menu

### LINKED ISSUE #1458: Get rid of hamburger menu
I don't think there's much value to the iodide hamburger menu anymore and we should just get rid of it:

image

* Help already has its own button elsewhere on the toolbar
* New notebook could be included under the user icon
* Clear variables would probably better be placed inside the workspace pane

iFlameing:
Fixes #1458 .


iFlameing:
sorry to disturb you. I will take care from the next time.

wlach:
I think we decided not to do this for now, in the end. Sorry for taking your time with this @iFlameing!

iFlameing:
@wlach No problem :)

PR #1466, opened 2019-01-31
Set Python indent to 4 in .editorconfig

openjck:
Most, if not all, Python source files already use an indent of 4. Will
confirms that this is the intended indent size for Python files.


hamilton:
Thanks @openjck!

PR #1467, opened 2019-01-31
separate out must-have onload css from theme setting in client

hamilton:
parsed out some top level must-have styles before the JS is loaded and kept the non-vital theme elements in injectSGlobal.


bcolloran:
works great, code looks good. thanks @hamilton! CI is still running, whoever notices first can merge s soon as it completes

hamilton:
@bcolloran thanks for hunting down the cause of the flickering

PR #1470, opened 2019-02-01
Add a pull request template

mdboom:
As discussed at our last meeting, it might be time to have a PR checklist to remind us of processes we'd like to keep up on.

Consider this just a starting point -- we can discuss the details of what to put here in this PR.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+
+
+### Pull Request checklist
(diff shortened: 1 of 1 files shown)


bcolloran:
thank you mike, this is great. i'm actually going to merge this, because we can refine it as we go if we need to but tbh there's nothing i would change, and i can't imagine any changes/discussion/objection weighty enough that we shouldn't start using this now.

PR #1490, opened 2019-02-07
reimplements the plugin eval fixes in #1483

### LINKED ISSUE #1459: need to catch and surface plugin load errors
I had an error that #1452 will fix, but the issue that prompted the PR was only uncovered if I went to the network pane of the devtools console to figure out what happened to pyodide. If there's a 404 or similar issue w/ loading a plugin, we need to surface that in the console.

hamilton:
### Pull Request checklist

This re-implements a number of fixes on the plugin evaluators I found in #1483. I believe this should be a pretty quick review + merge. Fixes #1459


src/eval-frame/actions/language-actions.js
```diff
@@ -136,7 +152,14 @@ export function runCodeWithLanguage(language, code, messageCallback) {
   if (asyncEvaluator !== undefined) {
     const messageCb =
       messageCallback === undefined ? () => {} : messageCallback;
-    return window[module][asyncEvaluator](code, messageCb);
+    try {
+      return window[module][asyncEvaluator](code, messageCb);
+    } catch (e) {
+      if (e.message === "window[module] is undefined") {
```
hamilton: I think this is a more helpful message than `window[module] is undefined`. If some other error is present, then return that instead.

src/eval-frame/actions/language-actions.js
```diff
@@ -136,7 +152,14 @@ export function runCodeWithLanguage(language, code, messageCallback) {
...
+        return new Error(`eval type ${module} not not defined`);
```
bcolloran: i think we can pull `displayName` out of `language` on line 151 in order to make this message clearer for end-users, maybe somthing like:

`Error evaluating ${language.displayName}; evaluation module "${module}" not not defined`

src/eval-frame/actions/language-actions.js
```diff
@@ -94,13 +105,18 @@ export function evaluateLanguagePlugin(pluginText, evalId) {
...
+      })
+      .catch(err => {
+        sendStatusResponseToEditor("ERROR", evalId);
+        return err;
+      });
   };
 }
 
-export function ensureLanguageAvailable(languageId, state, dispatch) {
+export function ensureLanguageAvailable(languageId, state, evalId, dispatch) {
```
bcolloran: might be missing it, but it looks like `evalId` is not used in this function?

src/eval-frame/actions/actions.js
```diff
@@ -161,7 +161,7 @@ function evaluateCode(code, language, state, evalId) {
       );
     };
 
-    return ensureLanguageAvailable(language, state, dispatch)
+    return ensureLanguageAvailable(language, state, evalId, dispatch)
```
bcolloran: as below, might be missing it, but it looks like evalId is not yet used in this function?

src/eval-frame/actions/language-actions.js
hamilton: sure

src/eval-frame/actions/language-actions.js
hamilton: yep that's a mistake, thought I had removed the remaining ones

hamilton:
@bcolloran thanks for the review. I've updated this PR with your suggestions. I feel there's probably more untangling to be done here in simplifying how the language actions work, but that can be tackled down the line.

bcolloran:
looks great!

PR #1499, opened 2019-02-08
[WIP] Fix issue 1397: Allow clients to post JSMD to /new

openjck:
TODO:

- [ ] Documentation
- [ ] Changelog
- [ ] Tests

I can write these soon, but first wanted to see if someone could confirm that this works as expected.

### DIFF
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -1,12 +1,14 @@
 import urllib.parse
+import json
 
 from django.db import transaction
 from django.shortcuts import (get_object_or_404,
                               redirect,
                               render)
-from django.template.loader import get_template
+from django.template import loader
 from django.urls import reverse
 from django.views.decorators.csrf import ensure_csrf_cookie
+from django.http import HttpResponse
 
 from .models import Notebook, NotebookRevision
 from ..files.models import File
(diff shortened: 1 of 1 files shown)


openjck:
## Testing the */new/* endpoint while logged in

1. Open the *Network* tab in Firefox Developer Tools
2. Navigate to [http://localhost:8000/new/](http://localhost:8000/new/)
3. Right-click the very first request that is listed, then and click *Edit and Resend*
4. Change the method from `GET` to `POST`
5. Change the value of the `Accept` header to `application/json`
6. Add a `Content-Type` header with value `application/json`
7. Add an `X-CSRFToken` header with a value equal to the value of the `csrftoken` cookie
8. Put some JSMD in the body of the `POST`. For example, `{ "jsmd": "%% md\n# Markdown title"}`
9. Click *Send* to resend the request
10. You should see the URL of a new notebook in the *Network* tab. Navigate to that page to see the result

This should also work when logged out (except that Iodide should redirect to the */tryit/* page, but for some reason Firefox isn't persisting the cookie through to the redirect. I'll keep trying to figure that out.

server/notebooks/views.py
```diff
@@ -100,19 +102,29 @@ def notebook_revisions(request, pk):
...
-        NotebookRevision.objects.create(
-            notebook=notebook,
-            content=new_notebook_content_template.render(),
-            title='Untitled notebook'
-        )
-    return redirect(notebook)
+    # If the POST body includes JSMD, use it.
+    try:
+        post_body = json.loads(request.body.decode('utf-8'))
+        notebook_content = post_body['jsmd']
```
wlach: It seems like we may as well allow specifying a custom title here.

mdboom:
I just wanted to share a use case I have.  I don't know if this PR is the place to do it, and I certainly don't want to hold up this PR.

For Pyodide testing, it's really handy to have a link that creates a new notebook with custom content that loads a local copy of Pyodide.  (I currently use the old static standalone notebook feature of older versions of Iodide). This is not a notebook that we would ever want to save, so it shouldn't add an entry in the notebook database.  Would POSTing to `/tryit/` be a way to make that work?

wlach:
> For Pyodide testing, it's really handy to have a link that creates a new notebook with custom content that loads a local copy of Pyodide. (I currently use the old static standalone notebook feature of older versions of Iodide). This is not a notebook that we would ever want to save, so it shouldn't add an entry in the notebook database. Would POSTing to `/tryit/` be a way to make that work?

If we get rid of `/tryit/` as I propose, we could add a create parameter to the `/new/` endpoint like `create` which, if falsey (0) doesn't create and save a new notebook, but just pastes the content into a blank notebook at `/new/`. Probably best to tackle the details of this in another issue.

PR #1509, opened 2019-02-13
deprecates app info and sends app message to console

hamilton:
This is further work ported from #1483, made easier because of #1491. 

- deprecates the app message pane but keeps the app messages in the store (used in the snackbar, which is still here)
- each call to `updateAppMessages` now calls `APPEND_TO_EVAL_HISTORY` using `createHistoryItem`
- fixes a handful of app message related bugs that were hidden


hamilton:
This PR as of this comment reflects most of the work I did before. I can't help but wonder if there's an issue with this approach, generally. For one thing, any app-level errors aren't actually surfaced in the console (they weren't surfaced before in the pane either), so you'll just get "Error saving notebook" or other similarly vague issues. 

My hope is that this is tackled as a follow-on to this one, since the point of this one is merely to deprecate the app messages pane as it is now and give a sensible alternative in the console.

bcolloran:
> I can't help but wonder if there's an issue with this approach, generally.

I'm not sure i'm following -- do you think we should keep the appInfoPane and mot merge this PR? Or can you say more about what approach you prefer? We certainly don't need to remove the app info pane if you think that's the wrong direction.

hamilton:
@bcolloran it's a bit of a different issue. right now, I have one `historyType` for this - `APP_MESSAGE`. So the content actually is a key for a pre-made component here (for instance `NOTEBOOK_SAVED` gives a boilerplate "Notebook saved to server" message or something like that) when in fact I could actually just have the `content` here be the detail message. The point of doing this was to prevent the `dangerouslySetInnerHtml` stuff we were doing before and switch to a set of components.

The drawback of this PR's approach is that if the app has an error (for instance I lost my internet connection and I can't actually save) all we say is "error saving notebook", but don't provide other information (such as the error message itself, 404, etc.). I see this as a problem. fwiw we don't surface those errors anyway right now.

But I also think this can be fixed in a later PR. What I really want to do is provide (1) the app message text (including the error if I need to) and (2) the _level_ of the app message (defaults to `info` for most messages, maybe `error` if there is an app error).

So with this in mind, I think I can take two approaches with this work right now:

(1) move away from using `content` to be an app message type and have it just give a plain string detail (rendered in a `pre` tag) and add an `error` level to errors when I do the work around solidifying the history components, or
(2) approach this in a later PR after the rest of the console work has been done.

Either approach is fine to me.

hamilton:
Thinking about it a bit more, I think i'd be more in favor of merging this as-is (with feedback of course) instead of changing things. I suspect there will be some refactoring to get a lot of the server-requesting code to return the right errors and I don't want to unravel too much in this PR which was meant to scope out specifically the deprecation of the app message pane.

bcolloran:
I see. the other thing to do would just be to add an `APP_ERROR` msg type if that's the only thing that would have level===error. In any case, if you want to move ahead with this PR, i'll do a review now and we can work through that in later PRs

hamilton:
@bcolloran very good suggestion. Let's move ahead with this PR and I can re-address this later.

src/eval-frame/components/panes/console/app-message.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+const MESSAGES = {};
+MESSAGES.NOTEBOOK_SAVED = <React.Fragment>Notebook saved.</React.Fragment>;
+MESSAGES.ERROR_SAVING_NOTEBOOK = (
+  <React.Fragment>There was an error.</React.Fragment>
+);
+MESSAGES.LOGGED_IN = <React.Fragment>You are now logged in.</React.Fragment>;
+MESSAGES.LOGIN_FAILED = <React.Fragment>Login failed.</React.Fragment>;
+MESSAGES.LOGGED_OUT = <React.Fragment>You are now logged out.</React.Fragment>;
+MESSAGES.LOGOUT_FAILED = <React.Fragment>Logout failed.</React.Fragment>;
```
bcolloran: Two questions re `MESSAGES`:
(1) it looks like strings will suffice for these rather than `React.Fragment`s
(2) why not just use one object literal rather than adding properties after defining an empty object? seems more verbose.

PR #1511, opened 2019-02-14
separate out inputs & outputs, refactor history items to have consistent design

hamilton:
### Pull Request checklist

- [x] **Documentation**: If this feature has or requires documentation, the relevant docs have been updated. __*(no new documentation necessary)*__
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not. __*(tests added where appropriate)*__

This is the last PR to be broken off from #1483, focusing on the last piece - separating the inputs from outputs, and standardizing all the console entries to have a design that is a bit closer to what you might see on the devtools console. It provides a table setting for console interception as well through the `ConsoleMessage` component. While this is a somewhat big PR, it's probably about 30% the size of the parent PR it came from. Review will need to be thorough but that would be expected anyway, given this is a pretty systematic refactor of how the console works.

Re: console entry UX and aesthetics, everything is up for debate. I hope this might start the conversation around how we want to approach this, and I doubt I got everything right here, so feel free to offer your feedback.


src/components/reps/error-handler.jsx
```diff
@@ -132,9 +139,9 @@ export default class ErrorRenderer extends React.Component {
   render() {
     const stack = trimStack(this.props.error);
     return (
-      <div className="error-output">
+      <ErrorPrintout>
```
hamilton: this standardizes the rep to have the same spacing as all other reps + outputs.

src/components/reps/preformatted-text-items-handler.jsx
```diff
@@ -1,6 +1,12 @@
 import React from "react";
+import styled from "react-emotion";
 import PropTypes from "prop-types";
 
+const TextItems = styled("pre")`
```
hamilton: this also standardizes the rep to have the same margin as the other outputs.

src/eval-frame/actions/actions.js
```diff
@@ -60,35 +60,41 @@ export function setConsoleLanguage(language) {
...
     content,
     historyType,
-    historyId: historyOptions.historyId,
-    lastRan: historyOptions.lastRan
+    historyId,
+    level,
+    language
   });
-  historyAction.type = "APPEND_TO_EVAL_HISTORY";
+  historyAction.type = "ADD_TO_CONSOLE";
```
hamilton: should this be `"ADD_TO_CONSOLE_HISTORY"` for sake of clarity?

src/eval-frame/actions/actions.js
```diff
@@ -138,33 +144,55 @@ export function evalConsoleInput(consoleText) {
...
-    const messageCallback = msg => {
-      const messageHistoryId = generateRandomId();
-      dispatch(
-        appendToEvalHistory(null, msg, undefined, {
-          historyType: "CELL_EVAL_INFO",
-          historyId: messageHistoryId
-        })
-      );
-    };
+    // const messageCallback = msg => {
```
hamilton: FIXME

src/eval-frame/actions/actions.js
```diff
@@ -187,7 +215,7 @@ export function evaluateText(
     // if (!evalText || !evalType) { return undefined }
     // FIXME: we need to deprecate side effects ASAP. They don't serve a purpose
     // in the direct jsmd editing paradigm.
-    const historyId = generateRandomId();
+    // const historyId = generateRandomId();
```
hamilton: FIXME

src/reducers/eval-frame-reducer.js
```diff
@@ -7,17 +7,26 @@ export default function evalFrameActionReducer(state, action) {
...
+    case "ADD_TO_CONSOLE": {
       const actionCopy = Object.assign({}, action);
       delete actionCopy.type;
       const history = [...state.history, actionCopy];
       return Object.assign({}, state, { history });
     }
 
     case "UPDATE_VALUE_IN_HISTORY": {
-      const history = [...state.history];
-      // const historyEntry = history.find(h => h.historyId === action.historyId)
```
hamilton: FIXME

src/eval-frame/actions/actions.js
bcolloran: sure that sounds good

bcolloran: looking through the code a little more, i think this is a great idea actually, seems substantially clearer. likewise for the action creator `addToConsole` -> `addToConsoleHistory` above

src/eval-frame/components/panes/__tests__/history-item.test.js
```diff
@@ -7,7 +7,6 @@ import React from "react";
 import { store } from "../../../store"; /* eslint-disable-line no-unused-vars */
 
 import HistoryItem from "../history-item";
-import PaneContentButton from "../pane-content-button";
```
bcolloran: as mentioned elsewhere: we should revive these tests since we're updating this stuff

src/components/reps/preformatted-text-items-handler.jsx
bcolloran: is there anything that should be defined in THEMES and pulled from that source if something about this being standardized across components?

src/components/reps/error-handler.jsx
bcolloran: as mentioned elsewhere: is there anything that should be defined in THEMES and pulled from that source if something about this being standardized across components?

src/actions/__tests__/eval-frame-action-validator.test.js
```diff
@@ -28,7 +28,7 @@ describe("validateActionFromEvalFrame should throw errors as expected", () => {
     expect(() =>
       validateActionFromEvalFrame({
         type: "UPDATE_VALUE_IN_HISTORY",
-        historyId: 12424242
+        historyItem: 10
```
bcolloran: this test should probably be expanded a since `historyItem` is now a composite object -- we'll want to test that the things sent over within historyItem are of the correct type, not just that historyItem is an object and not an integer

bcolloran: i.e., in addition to this test case, we should add cases for subfields

src/actions/eval-frame-action-validator.js
```diff
@@ -92,11 +92,25 @@ const schemas = {
...
+      historyItem: {
+        type: "object",
+        properties: {
+          historyId: { type: "string" },
+          content: { type: "string" },
+          language: { type: "string" },
+          level: { type: "string" },
+          historyType: { type: "string" },
+          lastRan: { type: "number" }
+        },
```
bcolloran: do we need all of these element to be sent just to update a value? language, lastRan will never need updating, but i suspect we probably don't need a few more of these

src/actions/__tests__/eval-frame-action-validator.test.js
hamilton: yep good point, happy to add tests here

src/actions/eval-frame-action-validator.js
hamilton: I suppose not

src/components/reps/error-handler.jsx
hamilton: Ah, the "standardization" here is primarily around moving the default padding in a `pre` element so it aligns with everything else in our console (reps, messages, etc.). I'm not sure if adding anything to `src/shared/theme.js` would benefit us.

src/components/reps/preformatted-text-items-handler.jsx
hamilton: same response as one above

src/actions/__tests__/eval-frame-action-validator.test.js
hamilton: hmm, on second thought, this isn't really the point of this test suite, is it? It looks like it's meant to test the action validator, not the contents of the eval frame schema. I added one additional test to this but it doesn't really feel like it belongs.

hamilton:
@bcolloran thanks for the feedback. I updated everything as you suggested, and also fixed the console bugs you brought up. Turns out both of those were caused by the same issue - the full history was being examined, not just the inputs, so non-input content was there. The fix was pretty straightforward. 

Let me know if there's anything else to add or tweak here. If not, feel free to merge at your convenience.

src/eval-frame/components/panes/console/app-message.jsx
```diff
@@ -26,9 +20,9 @@ const MESSAGES = {
 const AppMessage = ({ messageType }) => {
   const message = MESSAGES[messageType];
   return (
-    <AppMessageContainer>
+    <ConsoleMessage level="log">
```
bcolloran: oops looks like this slipped thru, should be "LOG" now

bcolloran: hmm makes me wonder if even on our purely declarative components we should include tests to make sure that they at least fully render without crashing (even tho such components don't have logic that needs testing). what do you think? like literally just a single test that makes sure shallow rendering doesn't crash would help us catch thing like this in the case of more complex presentational components with e.g. lots of children or props.

hamilton: I agree with this. I like these declarative bits but we still need to find a way of failing tests if someone doesn't declare properly.

PR #1513, opened 2019-02-14
Fix #1493: Don't crash when the reps error out

### LINKED ISSUE #1493: iodide console errors out when returning a python module
This is related to the new reps. This code block breaks things:

```python
import numpy
numpy
```

mdboom:
There's sort of a weird way one handles exceptions in React components, but this seems to work.

It falls back on `String(obj)` if React inspector crashes while trying to represent something.

@hamilton: Maybe this needs some styling to indicate it's a fallback rep?

### Pull Request checklist

### DIFF
--- a/src/components/reps/default-handler.jsx
+++ b/src/components/reps/default-handler.jsx
@@ -8,19 +8,7 @@ export default class DefaultRenderer extends React.Component {
     value: PropTypes.any
   };
 
-  constructor(props) {
-    super(props);
-    this.state = { hasError: false };
-  }
-
-  componentDidCatch() {
-    this.setState({ hasError: true });
-  }
-
   render() {
-    if (this.state.hasError) {
-      return <pre>{String(this.props.value)}</pre>;
-    }
     return <Inspector data={this.props.value} shouldShowPlaceholder={false} />;
   }
 }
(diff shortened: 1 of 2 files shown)


bcolloran:
Oh I didn't know about error boundaries, very cool and useful! Looks like the correct way to handle this kind of thing to me, Mike.

I think that everything about this approach looks right, but that this error boundary should probably live in the ValueRenderer, which is upstream of the DefaultRenderer and handles other stuff (including custom renderers) that I could easily imagine causing the rep to crash.

Also, maybe we should show info about the error, like in this codepen:
https://codepen.io/gaearon/pen/wqvxGa

```
  componentDidCatch(error, errorInfo) {
    // Catch errors in any components below and re-render with error message
    this.setState({
      error: error,
      errorInfo: errorInfo
    })
    // You can also log error messages to an error reporting service here
  }
  
  render() {
    if (this.state.errorInfo) {
      // Error path
      return (
        <div>
          <h2>A value renderer encountered an error.</h2>
          <pre>
            {this.state.error && this.state.error.toString()}
            <br />
            {this.state.errorInfo.componentStack}
          </pre>
          <a href="https://github.com/iodide-project/iodide/issues/new">
            Please file a bug report.
          </a>
        </div>
      );
    }
    // Normal render path
  }  
```

Thanks Mike!

mdboom:
Great.  I've updated this to display the error, but also added a display of `value.toString()`, which at least gives the user some indication of the object they are dealing with.

PR #1524, opened 2019-02-20
JSS-based components (material-ui) now come before emotion styles (thus fixing the cascade issue)

### LINKED ISSUE #1515: due to recent webpack upgrades, css ordering has changed, causing style inconsistencies throughout the client
_EDIT:_ I was able to recreate this locally after dumping all the assets and starting fresh.

I noticed this while working on our internal staging server, but this appears to be an issue on definitely-staging as well.

~Here's how it looks locally on `master`~ here's how it looks on my two-week old version of `master`:

screen shot 2019-02-15 at 10 43 35 am

You can see the style cascade does what we want - `material-ui` has overridden styles from `emotion`:

screen shot 2019-02-15 at 10 45 42 am

On staging, svg icons not getting resized according to our `emotion` wrappers:

screen shot 2019-02-15 at 10 43 30 am

Here is the css for that element on the staging server. You can see the emotion-set widths are clobbered by the `material-ui` styling.

screen shot 2019-02-15 at 10 45 48 am

Pretty curious that the server build does one thing, and my local build does another here.

hamilton:
### Pull Request checklist

- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not - no way to test this aspect of the build.

This fixes #1515.

Perhaps the best way to test this is to run `npm run build-production` on this branch, and note that the icons in the console are not huge. If you run `npm run build-production` on `master`, however, you'll notice the issue.


bcolloran:
fix confirmed, and the code looks fine.

I'm with you in increasingly questioning the use of css-in-js for our project, and adding a second css in js  dependency to support it seems ugly given what it buys us -- i think the biggest problem that we needed to solve re: css was accomplished by the effort you put into consolidating a bunch of stuff in the `THEME`, but almost all of the other problems we would want to solve can be handled by inline styles (from what i've read, the only thing not supported by inline styles are special selectors like `:hover`, which we don't use very much).

let's merge this to get things working, but when material-ui 4.0 is ready, let's see if we can replace this and emotion with material's solution

ps reading material that i found
https://hackernoon.com/all-you-need-to-know-about-css-in-js-984a72d48ebc
https://mxstbr.blog/2016/11/inline-styles-vs-css-in-js/

hamilton:
I'll take a look at those links to see what you mean a bit more precisely, but my initial impression is that I'd rather remove material ui entirely, given all the bloat, confusion, and difficulties we've encountered with it. I have used styled components (emotion & styled-components specifically) on other projects since I brought emotion into iodide and I've never had a single problem with them on their own. I'm open to anything, but deprecating emotion would be _far_ more work than deprecating material-ui.

PR #1526, opened 2019-02-21
Switch away from pipenv to requirements files

robhudson:


robhudson:
A couple notes/questions:

1. I added pip hashes. Do we want those? Do we also want them for the `requirements/tests.txt` file? I would say they are less needed there since those are only for testing. And currently the build fails b/c flake8 has a dependency that isn't a hard-pinned dependency.

2. This currently builds with all the requirements including the test requirements for the Dockerfile. ATMO has a `Dockerfile.dev` that is used which builds with all dependencies for testing, and a regular docker file for minimal requirements for the production build. Do we want to follow the same pattern?

3. I didn't double check all requirements are the latest versions. I've updated some. I figured once this is merged we can configure pyup or dependabot or other tool to keep the requirements up to date.

wlach:
This is great, thanks @robhudson!

> A couple notes/questions:
> 
>     1. I added pip hashes. Do we want those? Do we also want them for the `requirements/tests.txt` file? I would say they are less needed there since those are only for testing. And currently the build fails b/c flake8 has a dependency that isn't a hard-pinned dependency.

I think the hashes are a good idea. We should probably pin all sub-dependencies as well.

>     2. This currently builds with all the requirements including the test requirements for the Dockerfile. ATMO has a `Dockerfile.dev` that is used which builds with all dependencies for testing, and a regular docker file for minimal requirements for the production build. Do we want to follow the same pattern?

I'm not crazy about the Dockerfile.dev approach, since it violates the "don't repeat yourself" rule. @jezdez do you have any thoughts on this? For a first pass, I think we can just install the dev dependencies for now.

>     3. I didn't double check all requirements are the latest versions. I've updated some. I figured once this is merged we can configure pyup or dependabot or other tool to keep the requirements up to date.

Sounds good! I believe we can use renovate here.

jezdez:
> This is great, thanks @robhudson!
> 
> > A couple notes/questions:
> > ```
> > 1. I added pip hashes. Do we want those? Do we also want them for the `requirements/tests.txt` file? I would say they are less needed there since those are only for testing. And currently the build fails b/c flake8 has a dependency that isn't a hard-pinned dependency.
> > ```
> 
> I think the hashes are a good idea. We should probably pin all sub-dependencies as well.
> 
> > ```
> > 2. This currently builds with all the requirements including the test requirements for the Dockerfile. ATMO has a `Dockerfile.dev` that is used which builds with all dependencies for testing, and a regular docker file for minimal requirements for the production build. Do we want to follow the same pattern?
> > ```
> 
> I'm not crazy about the Dockerfile.dev approach, since it violates the "don't repeat yourself" rule. @jezdez do you have any thoughts on this? For a first pass, I think we can just install the dev dependencies for now.

TBH nowadays I would probably use a different structure like defining an [`ARG` parameter](https://docs.docker.com/engine/reference/builder/#arg) in the Dockerfile for the requirements file to install during the build of the image. Then I'd specify a base requirements file (e.g. requirements/dev.txt and requirements/base.txt, where the first includes the latter) in the CI config in the deployment build process.

wlach:
This change broke heroku, because it detected python either via a top-level Pipfile or requirements.txt. Our CI wasn't set up to catch this, which I think is fine (this is something that should almost never change). Fixed in #1533.

PR #1535, opened 2019-02-22
Automatically determine pyodide version to cache on heroku (fixes #1426)

### LINKED ISSUE #1426: Don't define PYODIDE_VERSION on Heroku
We're still in a phase where the interface between Iodide and Pyodide is evolving, so it's necessary to track specific versions of Pyodide for use with specific versions of Iodide.  As we ran into earlier today, it's pretty easy to forget to update the `PYODIDE_VERSION` environment variable on Heroku when this happens.  I think if we remove the variable from the Heroku config, it will fall back on the version defined in `webpack.config.js` and we would just have a single place to update.  I think this a case where we don't necessarily need deployment specific configuration, since it's really the version of Iodide used that determines what version of Pyodide we want to use.

@wlach: What do you think?

wlach:


wlach:
Tested this on staging and it seems to work well.

PR #1544, opened 2019-02-26
Add a link to the user's github profile on the user page

wlach:


wlach:
On github-enabled iodide instances, clicking the little github icon in this area now takes you to the appropriate user's github profile:

image

On stage (where we don't have this information, currently) there should be no such icon.

server/views.py
```diff
@@ -8,6 +8,7 @@
 
 from .notebooks.models import Notebook
 from .base.models import User
+from django.conf import settings
```
robhudson: This should go up with the other django imports. Typically these are grouped by: stdlib, 3rd-party, then local imports.

wlach: Ah good point, we should probably figure out a way to check import order automatically. I think the cool kids might be using black for this...

robhudson: `isort` has a `--check-only` flag and can be used like `flake8` to check code. This could be added to CircleCI. Feel free to file an issue and I can add it in some time.

wlach: Makes sense, I just checked and black (which I think we should also consider using) doesn't in fact do this. I'll file some an issue to try to get both implemented as part of our ci step.

wlach: Filed #1552

src/server/pages/user-page.jsx
```diff
@@ -62,7 +66,14 @@ export default class UserPage extends React.Component {
                 alt={`${thisUser.name}'s avatar`}
               />
               <h1>{thisUser.full_name}</h1>
-              <h2>{thisUser.name}</h2>
+              <h2>
+                {thisUser.name}{" "}
+                {thisUser.github_url && (
+                  <a href={thisUser.github_url}>
```
hamilton: (1) mind if we reverse the order so it is icon then name? That'd require a right-ad on the icon of about 5px
(2) can we set this h2 `font-size` to `1em`? It feels a bit big, and I imagine down the line we may want to include other things like this
(3) could we have the anchor wrap the text as well, not just the icon?
(4) can the anchor come with a `target='_blank'`?

These are all a matter of taste, so feel free to take it or leave the feedback!

screen shot 2019-02-27 at 1 29 16 pm

wlach: this all sounds good to me, i'll try to change things as you suggest

PR #1548, opened 2019-02-27
Add Dockerfile ARG for controlling Python deps

robhudson:
TODO:

- [x] Update the CircleCI config to build the production docker for deployment but the testing build for running the tests.

The above may remove any benefits to caching of the docker artifacts.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,11 +133,11 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker
-      - restore_cache:
-          key: v1-{{.Branch}}
+      # We are not using the cached docker file here b/c we want the production
+      # build to use only the required Python files.
       - run:
-          name: Restore Docker image cache
-          command: docker load -i /tmp/docker.tar
+          name: Build Docker image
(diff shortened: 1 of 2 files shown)


wlach:
Hmm, yeah, breaking it down:

1. I think we definitely need/want the docker caching for tests, and ideally doing a release wouldn't force a rebuild of the cache.
2. Releases happen relatively infrequently (once a week or so), so I think it's ok if we don't use the docker caching in those cases (although it's obviously ok to do so where we can).

I feel like maybe all we need to do is set the right caching keys for circle, incorporating tags if needed (tags==releases)? Most of what you see in iodide's circle configuration is based on what @whd implemented for mission control, it would be good to get either him and/or @robotblake to review things?

https://github.com/iodide-project/iodide/blob/5cd5fefbf0a55ee1340ad38975206a32f7a5146d/.circleci/config.yml#L72
https://github.com/iodide-project/iodide/blob/5cd5fefbf0a55ee1340ad38975206a32f7a5146d/.circleci/config.yml#L137

Issue #1416, opened 2019-01-23
iomd is not HTML-escaped, allowing arbitrary code execution


bcolloran:
re #382 (which this supercedes) -- now that we're not focused on the single runnable file/bundle user story, it's really not necessary to patch the IOMD into that script tag. We should just send the IOMD down as a string in a IOMD object or script, or whatever mechanism we're using to send over eg userData. This will solve #382 and also maybe simplify our start up code a teeny bit.

wlach:
I'm not convinced that putting iomd in a script tag is inherently a problem, however we do need to escape the content of the iomd, to disallow arbitrary code execution as you can (currently) see in this example:

https://definitely-staging.iodide.io/notebooks/63/

wlach:
Ironically we are currently marking the iomd as "safe":

https://github.com/iodide-project/iodide/blob/c08375515cfe7d48259498c66f53bcaf63a0c9b4/server/notebooks/templates/notebook.html#L7

Likely we should change this from "safe" to "escape": 

https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#escape

We would have to add some code in the client to unescape this tag, but that shouldn't be too hard.

We should also add unit tests on both the client and the server to make sure this is working as expected.

bcolloran:
> I'm not convinced that putting iomd in a script tag is inherently a problem

ya, it's not a big problem, just a bit of code clean up -- it's not really necessary to pull the iomd out of the DOM at this point since we have not been optimizing for the single human-readable HTML file for quite some time.

wlach:
> > I'm not convinced that putting iomd in a script tag is inherently a problem
> 
> ya, it's not a big problem, just a bit of code clean up -- it's not really necessary to pull the iomd out of the DOM at this point since we have not been optimizing for the single human-readable HTML file for quite some time.

It is actually more efficient to embed the iomd in the document on the server side (saves an API round-trip), so my vote would be to keep things this way.

bcolloran:
oh i wasn't proposing that we stop embedding the iomd in the page, just that we not put it in the `<script type="iomd">` tag and then parse it out at as text. i thought we should do it more like in the example at https://facebook.github.io/create-react-app/docs/title-and-meta-tags#injecting-data-from-the-server-into-the-page -- by setting it as a variable directly on `window` in a script tag, because i thought that we'd already been doing that for e.g. notebookInfo and other stuff. But i now see that we're getting notebookInfo with

```
[...]

PR #1555, opened 2019-02-27
Don't show "new notebook" button on pages other than the current user's own

wlach:


src/server/pages/user-page.jsx
```diff
@@ -50,6 +50,8 @@ export const isLoggedIn = userInfo => "name" in userInfo;
 export default class UserPage extends React.Component {
   render() {
     const { thisUser, userInfo, notebookList } = this.props;
+    const isUserAccount =
+      isLoggedIn(userInfo) && thisUser.name === userInfo.name;
```
robhudson: Just out of curiosity, the `name` must be unique but what is the underlying data? Where does it come from?

wlach: So we use a kind of hacky strategy on the server of embedding this kind of metadata inside the html as a script element. Here's where it comes from in this case:

https://github.com/iodide-project/iodide/blob/0e06560240c21a6665edccee496265408f33e47c/server/views.py#L71

bcolloran: side note: i think we should change that hacky implementation detail at some point -- https://github.com/iodide-project/iodide/issues/1416

src/server/pages/user-page.jsx
```diff
@@ -50,6 +50,8 @@ export const isLoggedIn = userInfo => "name" in userInfo;
 export default class UserPage extends React.Component {
   render() {
     const { thisUser, userInfo, notebookList } = this.props;
```
hamilton: I've been very sloppy with these server components as I've added them, including a bit of logic in the render step, but hope we can apply the same code quality on these pages as we move forward. Is there any chance we can use this as an excuse to pull the logic out of `render` and put it in a function?

wlach: Sure, what did you have in mind? A class-specific helper function or... ?

wlach: As discussed, let's just merge this as-is. We can refactor this later.

PR #1557, opened 2019-02-27
Add proper page titles to all server-returned pages

wlach:
This makes pages appear much cleaner in the URL bar (we
already had client-side code to set the title for the notebook
view, but it would only be activated after the page had
fully loaded and its javascript code was interpreted/run)

### DIFF
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -50,6 +50,7 @@ def notebook_view(request, pk):
     else:
         notebook_info['forked_from'] = False
     return render(request, 'notebook.html', {
+        'title': revision.title,
         'user_info': _get_user_info_json(request.user),
         'notebook_info': notebook_info,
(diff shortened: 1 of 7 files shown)


server/notebooks/views.py
```diff
@@ -88,6 +89,7 @@ def notebook_revisions(request, pk):
         'date': revision.created.isoformat()}
         for revision in NotebookRevision.objects.filter(notebook_id=pk)])
     return render(request, '../templates/index.html', {
+            'title': f'Revisions - {nb.title}',
```
robhudson: I didn't see a matching test for this change?

wlach: No, not yet... we actually don't have a test for this view yet. I should probably do the hard thing and add one.

wlach: Done now

PR #1559, opened 2019-02-27
fix hard-coded links to point to newer versions of pyodide tour nb, changed link in iodide starter -> iodide.io/tryit

hamilton:
javascript starter now goes through https://iodide.io/tryit, python t…our is now NB 300, updated 'getting started with python' link to point to NB 300


bcolloran:
:+1:  thanks!

PR #1561, opened 2019-02-28
Add linting with black and isort

robhudson:
Without the commits to reformat the code using `black` and `isort`, this will break the build. Those changes will come in follow up commits. But first I wanted to test and make sure these checks will fail before committing the correcting changes.

Any comments on execution here is welcomed.


robhudson:
We may want to add some docs in the developer docs about conforming to the isort and black standards? Are there any other things we need to do here?

bcolloran:
> We may want to add some docs in the developer docs about conforming to the isort and black standards? 

Yeah, that would be great! I don't know of anything else to add, perhaps @wlach or @hamilton have ideas tho

Makefile
```diff
@@ -14,8 +14,11 @@ shell:
 up:
 	docker-compose up
 
-flake8:
-	docker-compose run server flake8 server/
+lint:
```
wlach: These new targets should be marked as `.PHONY` at the top of the file, as they don't depend on any files changing.

wlach:
Looks great, I will merge as soon as I confirm CI is passing

wlach:
Thanks so much!

PR #1565, opened 2019-02-28
implements a basic autosave feature

hamilton:
Every _k_ (_k_=10 in this PR) seconds, we compare the last-saved version to the current jsmd, and if different, save to server. If you lose your connection, surface a message saying you've lost the connection, that you'll keep trying, and that in the meantime your changes will be preserved locally.

Some minor extensions:

- [x] we can actually have it so that if there's a local autosave (vs. a server one) always apply it, since the server might eventually pick up the change & that's the behavior we want, for the most part.
- [x] this needs to pick up network error issues, 
- [x] and have a way of somewhat gracefully surfacing those network errors so they're obvious but not too unobtrusive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,4 @@
 # (Unreleased; add upcoming change notes here)
-# 0.4.0
 - iodide notebooks now automatically save to the server
 - removed `ctrl+s` as a keybinding
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # (Unreleased; add upcoming change notes here)
+# 0.4.0
+- iodide notebooks now automatically save to the server
+- removed `ctrl+s` as a keybinding
 
 # 0.3.0 (2019-03-06)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 # (Unreleased; add upcoming change notes here)
 - iodide notebooks now automatically save to the server
-- removed `ctrl+s` as a keybinding
 
 # 0.3.0 (2019-03-06)
 
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -332,7 +332,7 @@ export function createNewNotebookOnServer(options = { forkedFrom: undefined }) {
   };
 }
 
-export function saveNotebookToServer() {
+export function saveNotebookToServer(appMsg = true) {
   return (dispatch, getState) => {
     const state = getState();
     const notebookId = getNotebookID(state);
@@ -347,12 +347,14 @@ export function saveNotebookToServer() {
         .then(() => {
           const message = "Updated Notebook";
           updateAutosave(state, true);
-          dispatch(
-            updateAppMessages({
-              message,
-              messageType: "NOTEBOOK_SAVED"
-            })
-          );
+          if (appMsg) {
+            dispatch(
+              updateAppMessages({
+                message,
+                messageType: "NOTEBOOK_SAVED"
+              })
+            );
+          }
           dispatch({ type: "NOTEBOOK_SAVED" });
         })
         .catch(() => {
@@ -529,4 +531,14 @@ export function saveEnvironment(updateObj, update) {
     updateObj,
     update
   };
+}
+
+export function setPreviouslySavedJsmdToCurrentJsmd() {
+  return {
+    type: "SET_PREVIOUSLY_SAVED_JSMD_TO_CURRENT_JSMD"
+  };
+}
+
+export function changeMadeToJsmd(state) {
+  return state.previouslySavedJsmd !== state.jsmd;
 }
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -258,7 +258,7 @@ export function evaluateNotebook(evalQueueInstance = evalQueue) {
   };
 }
 
-function getNotebookSaveRequestOptions(state, options = undefined) {
+export function getNotebookSaveRequestOptions(state, options = undefined) {
   const data = {
     title: state.title,
     content: state.jsmd
@@ -274,8 +274,8 @@ function getNotebookSaveRequestOptions(state, options = undefined) {
 }
 
 export function saveNotebookRequest(url, postRequestOptions, dispatch) {
-  return fetchWithCSRFTokenAndJSONContent(url, postRequestOptions).then(
-    response => {
+  return fetchWithCSRFTokenAndJSONContent(url, postRequestOptions)
+    .then(response => {
       if (!response.ok) {
         return response
           .json()
@@ -303,8 +303,16 @@ export function saveNotebookRequest(url, postRequestOptions, dispatch) {
           });
       }
       return response.json();
-    }
-  );
+    })
+    .catch(err => {
+      dispatch(
+        updateAppMessages({
+          message: `Error Saving Notebook.`,
+          messageType: "ERROR_SAVING_NOTEBOOK"
+        })
+      );
+      throw new Error(err.message);
+    });
 }
 
 export function createNewNotebookOnServer(options = { forkedFrom: undefined }) {
@@ -313,7 +321,11 @@ export function createNewNotebookOnServer(options = { forkedFrom: undefined }) {
     const postRequestOptions = getNotebookSaveRequestOptions(state, {
       forkedFrom: options.forkedFrom
     });
-    saveNotebookRequest("/api/v1/notebooks/", postRequestOptions, dispatch)
+    return saveNotebookRequest(
+      "/api/v1/notebooks/",
+      postRequestOptions,
+      dispatch
+    )
       .then(json => {
         const message = "Notebook saved to server";
         dispatch(
@@ -339,7 +351,7 @@ export function saveNotebookToServer(appMsg = true) {
     const notebookInServer = Boolean(notebookId);
     if (notebookInServer) {
       // Update Exisiting Notebook
-      saveNotebookRequest(
+      return saveNotebookRequest(
         `/api/v1/notebooks/${notebookId}/revisions/`,
         getNotebookSaveRequestOptions(state),
         dispatch
@@ -357,12 +369,11 @@ export function saveNotebookToServer(appMsg = true) {
           }
           dispatch({ type: "NOTEBOOK_SAVED" });
         })
-        .catch(() => {
-          // do nothing here.
+        .catch(err => {
+          throw Error(err);
         });
-    } else {
-      createNewNotebookOnServer()(dispatch, getState);
     }
+    return createNewNotebookOnServer()(dispatch, getState);
   };
 }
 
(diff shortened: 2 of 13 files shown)


src/actions/actions.js
```diff
@@ -317,7 +317,7 @@ export function createNewNotebookOnServer(options = { forkedFrom: undefined }) {
   };
 }
 
-export function saveNotebookToServer() {
+export function saveNotebookToServer(appMsg = true) {
```
hamilton: here's the easiest way to turn off the snackbar (without `updateAppMessages` no snackbar)

src/actions/actions.js
```diff
@@ -515,3 +517,13 @@ export function saveEnvironment(updateObj, update) {
     update
   };
 }
+
+export function setPreviouslySavedJsmdToCurrentJsmd() {
+  return {
+    type: "SET_PREVIOUSLY_SAVED_JSMD_TO_CURRENT_JSMD"
```
bcolloran: could tighten and generalize this to something like `SET_MOST_RECENT_UPLOAD` or `SET_MOST_RECENT_SAVED_CONTENT` -- we may want to compare more than just JSMD (NB title, comments, who knows)

src/state-schemas/state-schema.js
```diff
@@ -128,6 +128,10 @@ export const stateProperties = {
     type: "string",
     default: ""
   },
+  previouslySavedJsmd: {
```
bcolloran: might generalize to eg
```
mostRecentUpload: {
jsmd: string,
title: string
}
```
?

src/tools/server-autosave.js
```diff
@@ -0,0 +1,18 @@
...
+export function changeMadeToJsmd(state) {
+  return state.previouslySavedJsmd !== state.jsmd;
+}
+export function checkForServerAutosave(store) {
+  store.dispatch(setPreviouslySavedJsmdToCurrentJsmd());
+  setInterval(() => {
+    const state = store.getState();
+    if (changeMadeToJsmd(state)) {
+      store.dispatch(saveNotebookToServer(false));
+      store.dispatch(setPreviouslySavedJsmdToCurrentJsmd());
```
bcolloran: edge case, but i wonder if perhaps `setPreviouslySavedJsmdToCurrentJsmd()` should be called within the `then` of the successful resolution of the `saveNotebookRequest` call _inside of_ `saveNotebookToServer(false)`.

this really is a small detail, but i think that maybe as written, the `store.dispatch(saveNotebookToServer(false));` could kick of it's server request, and we'd still dispatch the `store.dispatch(setPreviouslySavedJsmdToCurrentJsmd());` even if the request failed for some reason. (also possible very minor race condition if there are changes made while the server request is in flight)

not sure it matters, but it just seems more correct somehow to set the previously saved contents within the confirmation from the server that the save request was successful

src/tools/server-autosave.js
```diff
@@ -0,0 +1,18 @@
+import {
+  setPreviouslySavedJsmdToCurrentJsmd,
+  saveNotebookToServer
+} from "../actions/actions";
+
+export function changeMadeToJsmd(state) {
```
bcolloran: could be `notebookChangedSinceSave` or `notebookChangedSinceUpload` -- more generic to allow comparison of other notebook stuff

wlach:
Some questions to think about here after reviewing the patch:

* How does this behave if there is a network problem and we can't access the server? Will we still have an autosaved copy?
* How does this behave if the user is editing multiple copies of the same document in different tabs or on different machines?

In some ways I think this opens the door for iodide to misbehave in confusing ways, so we probably want to tread carefully here. Etherpad, for example, has a fairly sophisticated conflict resolution algorithm which accounts for the above cases and probably others I haven't thought of off the top of my head.

hamilton:
@wlach I agree with your concerns, but I think in this case we don't have too far to go from this PR to make sure it's robust enough for `extremely-alpha`.

(1) network problems: we can address these in this PR I believe. The local autosave still functions as-is, and I think the goal will be that if your network drops and you don't realize it and close the notebook tab, when you go back to the notebook, our local save should restore what you had automatically. If the network connection pops back on, then we can resume saving.
(2) to achieve (1) above, we'll need to make sure we have robust error catching in the event of network issues, which should not be hard to deal with in this PR, so we aren't accidentally saying "hey it saved!" when it in fact didn't. I just wanted to get the POC out the door so we could get the convo going, but having more robust error catching in the face of network issues is a must-have.
(3) if we were dealing with realtime collaboration, there are far more edge cases to handle, and I think a more sophisticated approach like what etherpad has would handle those cases. Thankfully we're just trying to keep people from losing their work, and even if this particular implementation won't get us to the other stuff, it will solve one of our biggest pain-points for the time being at a pretty low cost. Once we have a more sophisticated approach down the line, my hope is that to the end user there will be no real difference in UX re: autosave.

On the whole, I hope that this PR isn't going to be a ton of additional engineering effort, because if it is, we might as well do this the "right" way that sets us up for all sorts of other situations. But really we just need to keep people from losing their work.

hamilton:
oh and (4): fwiw we won't be "solving the multiple tabs of same notebook open" problem in this PR, unless there is an easy win where we use a ShareWorker to listen for other tabs, and if one exists, pop up a modal syaing "you already have a tab with this notebook open" or somethign to that effect.

src/tools/server-autosave.js
hamilton: I think this is the right idea. We need a Promise to be returned here so we can catch errors of various kinds.

hamilton:
Ok, made the so-far requested changes. Uncovered another bug w/ autosave referenced above, which will better unblock this.

At this juncture I'd like to figure out what to do w/ a failed network request. At the moment it will aggressively bring up app messages / console messages that we had an error in saving.

wlach:
> @wlach I agree with your concerns, but I think in this case we don't have too far to go from this PR to make sure it's robust enough for `extremely-alpha`.
> 
> (1) network problems: we can address these in this PR I believe. The local autosave still functions as-is, and I think the goal will be that if your network drops and you don't realize it and close the notebook tab, when you go back to the notebook, our local save should restore what you had automatically. If the network connection pops back on, then we can resume saving.

That sounds good to me, but we'll need to make that work. :) Currently if there are local changes and the page is refreshed, we pop up a "You have made changes to this notebook that are only saved locally. Restore or discard." header message. We should probably change the behaviour here to auto-restore local changes, and just get rid of that message (maybe in the case where a user isn't logged in or doesn't own the notebook, offer the option of reverting to the original copy?).

> (3) if we were dealing with realtime collaboration, there are far more edge cases to handle, and I think a more sophisticated approach like what etherpad has would handle those cases. Thankfully we're just trying to keep people from losing their work, and even if this particular implementation won't get us to the other stuff, it will solve one of our biggest pain-points for the time being at a pretty low cost. Once we have a more sophisticated approach down the line, my hope is that to the end user there will be no real difference in UX re: autosave.
>  
> On the whole, I hope that this PR isn't going to be a ton of additional engineering effort, because if it is, we might as well do this the "right" way that sets us up for all sorts of other situations. But really we just need to keep people from losing their work.

Yup, agreed. I think this incremental approach is the right way to solve the problem and thanks for getting the ball rolling on it, I just want to make sure that things work predictably and reasonably robustly (and that we have a good foundation for the next few steps). I think if it takes a few more days/PR cycles to get there, that should be ok. It's worth being *a little* careful for such a core piece of functionality. :)

> At this juncture I'd like to figure out what to do w/ a failed network request. At the moment it will aggressively bring up app messages / console messages that we had an error in saving.

Let's add a header message notifying the user that there has been an error saving to the server, and that we will try again when the user next connects.

src/actions/task-definitions.js
```diff
@@ -92,14 +92,13 @@ tasks.newNotebook = new ExternalLinkTask({
...
+// this overrides the browser default's ctrl+s but otherwise does nothing.
 tasks.saveNotebook = new UserTask({
   title: "Save Notebook",
   keybindings: ["ctrl+s", "meta+s"],
   displayKeybinding: `${commandKey}+s`,
   preventDefaultKeybinding: true,
-  callback() {
-    dispatcher.saveNotebookToServer();
-  }
+  callback() {}
```
wlach: Glitch pops up a little dialog saying "pssst, glitch autosaves!" when you use this shortcut. Might be something to think about for the future.

hamilton: coincidentally, I brought up this idea w/ @bcolloran yesterday afternoon! Agree we could account for that.

hamilton:
@wlach @bcolloran this PR is ready for another review. Might be good to dogfood this a bit on `definitely-staging` before merging / releasing.

bcolloran:
code looks fine on a quick review, but i honestly haven't looked deeply b/c i know you two have covered this pretty thoroughly. Now that we have an extra day to maneuver, it would be nice to see if we can figure out how to get some tests around some of this material, but i wouldn't block on that in this case, b/c i think we all agree that we really want to see this in a release by monday \o/ :-)

src/tools/server-autosave.js
```diff
@@ -0,0 +1,43 @@
...
+      if (validAutosave) {
+        store.dispatch(setMostRecentSavedContent());
+        // remove previous autosave if we've successfully saved.
+        clearAutosave(state);
+        if (state.notebookInfo.connectionStatus !== "CONNECTION_ACTIVE") {
+          store.dispatch(setConnectionStatus("CONNECTION_ACTIVE"));
+        }
+      }
+    }
+  }, 10000);
```
wlach: So I think we should set this to 30 seconds. Rationale:

1. It will make the history viewer more manageable
2. It will reduce the size of the database dramatically
3. Even if the user accidentally closes their tab, autosave should restore it if the user re-opens the window.

Also, could you make this a constant and put it at the top of the file? Something like `AUTOSAVE_INTERVAL` would be fine.

src/actions/task-definitions.js
wlach: Filed #1586

src/actions/actions.js
wlach: Is this part relevant anymore? i.e. is there any situation where we'd pop up a snackbar? I guess it might be useful for #1586, in which case we could leave it in for now (along with the keybinding).

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # (Unreleased; add upcoming change notes here)
+# 0.4.0
+- iodide notebooks now automatically save to the server
+- removed `ctrl+s` as a keybinding
```
wlach: I think you can skip the last bullet. Also, it doesn't matter so much, but I don't think it's necessary to pre-emptively fill in the next version number. Whoever bumps the changelog can assign their own.

wlach:
Opened #1587 for the revision viewer issues. That can land either before or after this one.

CHANGELOG.md
hamilton: np, wasn't really thinking deeply about that :)

hamilton:
Thanks for the reviews, y'all.

PR #1572, opened 2019-03-02
if there is no title provided to REPLACE_NOTEBOOK_CONTENT, default to state.title

### LINKED ISSUE #1571: restoring a locally-saved set of changes with the "restore" link causes a "bad request" error when saving
I noticed this while attempting to bring harmony to both local and server autosaving on #1565, realizing that perhaps this particular bug was not brought about by changes in that branch.

on `extremely-alpha`, if I make a change to a notebook and then refresh the page (without saving), our autosave functionality kicks in, and I have the option of restoring my unsaved changes. If I click "restore", however, all further save attempts in that page load (through `ctrl+s`) throw an error. It appears I'm getting a `400 BAD REQUEST` error as a result. If I attempt to just edit the document further and save after the restore, I still get the 400 error.

If I ignore the "restore" / "discard" banner entirely and start editing my notebook and save through `ctrl+s`, there is no issue. So it seems to be something about our local autosave functionality that causes the issue. I'll have to dig into this when I have a bit more time, but I suspect something about our autosave implementation is perhaps transforming the saved CSRF token.

@wlach any other thoughts why this might be the case?

hamilton:
This PR fixes #1571, and opens the door to unblocking #1565.

### DIFF
--- a/src/reducers/notebook-reducer.js
+++ b/src/reducers/notebook-reducer.js
@@ -93,7 +93,7 @@ const notebookReducer = (state = newNotebook(), action) => {
       return Object.assign({}, state, {
         jsmd: action.jsmd,
         jsmdChunks: action.jsmdChunks,
-        title: action.title
+        title: action.title || state.title
       });
     }
 


bcolloran:
looks like a simple and sensible fix, thanks @hamilton! merging now

Issue #1567, opened 2019-03-01
visiting the user page of user without human name shows "username ()" in the tab


bcolloran:
for example, i've never entered my human name in my GH account, so the name displayed in the tab and in the browser window title bar when looking at my user page is "bcolloran ()":

https://extremely-alpha.iodide.io/bcolloran/

FarooqAR:
I want to take this issue.

PR #1576, opened 2019-03-04
fix title for user page

FarooqAR:
fixes [#1567](https://github.com/iodide-project/iodide/issues/1567)

### DIFF
--- a/server/tests/test_server_pages.py
+++ b/server/tests/test_server_pages.py
@@ -54,12 +54,20 @@ def test_index_view(client, two_test_notebooks, fake_user, logged_in):
     }
 
 
-@pytest.mark.parametrize("username", ["testuser", "test-user", "testuser@foo.com"])
-def test_user_view_with_different_names(transactional_db, client, username):
+@pytest.mark.parametrize(
+    ("username", "first_name", "last_name"),
+    [
+        ("testuser", "User", "McUserTons"),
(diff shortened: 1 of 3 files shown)


FarooqAR:
Thanks for the feedback :) I'll surely try my best to add the changes as requested.

server/tests/test_server_pages.py
```diff
@@ -54,12 +54,19 @@ def test_index_view(client, two_test_notebooks, fake_user, logged_in):
...
 
 
-@pytest.mark.parametrize("username", ["testuser", "test-user", "testuser@foo.com"])
-def test_user_view_with_different_names(transactional_db, client, username):
+@pytest.mark.parametrize(
+    ("username", "first_name", "last_name"),
+    [
+        ("testuser", "User", "McUserTons"),
+        ("testuser@foo.com", "Test", ""),
+        ("test-user", "", "User"),
```
wlach: You need to add a test case where get_full_name() returns "". I believe that would be:

`("test-user", "", "")`

I'm pretty sure that will just work thanks to your changes, but try it out and see!

wlach:
Also, for the future, please give your pull request / main commit a description of what was actually changed, instead of just "fix XXX". In this case, something like "Improve title for user page when no first or last name is available (fixes #1567)" would be good.

PR #1579, opened 2019-03-04
Don't allow the creation of empty revisions with no changes

wlach:
I tested this on the client and it works well. Currently it still says "Notebook updated" while silently not updating anything if you save with no changes. An obvious improvement would be to say "No changes detected" instead but I feel it's better to just leave the client side as-is for now while we work on the auto-save implementation. There are definitely worse usability gotchas in iodide as things stand, and I'd rather not write code/tests that are going to be deleted/massively refactored within the next few weeks.

@robhudson you're probably not too familiar with the surrounding context here, but I thought it would be good to get your opinion on the code changes. Basically we want to prevent people from accidentally creating tons of no-op revisions by pressing Meta-S repeatedly (or when we land an implementation of auto-save, accidentally doing the same automatically). 

### Pull Request checklist

### DIFF
--- a/server/notebooks/serializers.py
+++ b/server/notebooks/serializers.py
@@ -51,6 +51,14 @@ class NotebookRevisionDetailSerializer(serializers.ModelSerializer):
     Details of a revision for a notebook (includes content)
     """
 
+    def validate(self, attrs):
+        last_revision = NotebookRevision.objects.filter(
+            notebook_id=self.context["notebook_id"]
+        ).first()
+        if attrs["title"] == last_revision.title and attrs["content"] == last_revision.content:
+            raise serializers.ValidationError("Revision unchanged from previous")
+        return serializers.ModelSerializer.validate(self, attrs)
+
     class Meta:
         model = NotebookRevision
         fields = ("id", "title", "created", "content")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - set default document position to (row 1, column 1) on startup (#1568)
 - displays owner name in header message when viewing a notebook you do not own
 - randomly generate initial notebook names (based on a list of iodide compounds) (#1541)
+- don't allow the creation of "empty" revisions on the server
 
 # 0.2.0 (2019-02-28)
 
--- a/server/tests/test_notebook_revision_api.py
+++ b/server/tests/test_notebook_revision_api.py
@@ -144,6 +144,44 @@ def test_create_notebook_revision(fake_user, test_notebook, client):
     test_notebook.refresh_from_db()
     assert test_notebook.title == post_blob["title"]
 
+    # also validate that the response is what we expect
+    assert resp.json() == {
+        "content": post_blob["content"],
+        "created": get_rest_framework_time_string(new_notebook_revision.created),
+        "id": new_notebook_revision.id,
+        "title": post_blob["title"],
+    }
+
+
+def test_dont_create_unmodified_notebook_revision(fake_user, test_notebook, client):
+    client.force_login(user=fake_user)
+
+    # get latest (only) revision for test notebook
+    notebook_revision = NotebookRevision.objects.first()
+    original_creation_time = notebook_revision.created
+    original_id = notebook_revision.id
+
+    # unmodified post
+    unmodified_post_blob = {"title": notebook_revision.title, "content": notebook_revision.content}
+
+    resp = client.post(
+        reverse("notebook-revisions-list", kwargs={"notebook_id": test_notebook.id}),
+        unmodified_post_blob,
+    )
+    assert resp.status_code == 400
+
+    # no new notebook revisions should be created, and the existing (only)
+    # revision should be unchanged
+    assert NotebookRevision.objects.count() == 1
+    notebook_revision = NotebookRevision.objects.first()
+    assert notebook_revision.title == unmodified_post_blob["title"]
+    assert notebook_revision.content == unmodified_post_blob["content"]
+    assert notebook_revision.created == original_creation_time
+    assert notebook_revision.id == original_id
+
+    # be sure that we get the expected error
+    assert resp.json() == {"non_field_errors": ["Revision unchanged from previous"]}
+
 
 def test_delete_notebook_revision_not_logged_in(test_notebook, client):
     # should not be able to delete a notebook revision if not logged in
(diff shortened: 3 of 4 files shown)


server/notebooks/api_views.py
```diff
@@ -80,9 +81,29 @@ def perform_destroy(self, instance):
 
     def perform_create(self, serializer):
         ctx = self.get_serializer_context()
-        notebook_owner_id = Notebook.objects.values_list("owner", flat=True).get(
-            id=ctx["notebook_id"]
-        )
-        if self.request.user.id != notebook_owner_id:
+        notebook = Notebook.objects.get(id=ctx["notebook_id"])
```
robhudson: Just in case, should we `get_object_or_404` here?

server/notebooks/api_views.py
```diff
@@ -80,9 +81,29 @@ def perform_destroy(self, instance):
...
+
+        # check if user has permission to create a revision
+        if self.request.user.id != notebook.owner_id:
             raise PermissionDenied
```
robhudson: This seems like something that would be used in a lot of places perhaps? If so, this could be pulled out into a view decorator. Similar to Django's `@staff_member_required` but `@notebook_owner_required` or something similar.

robhudson: Looking closer this isn't a simple Django view, it's part of the `ModelViewSet`, but maybe DRF has hooks to define something similar as a permission class that can be re-used?

server/notebooks/api_views.py
```diff
@@ -80,9 +81,29 @@ def perform_destroy(self, instance):
...
+
+        # check that this revision has genuinely new content
+        last_revision = NotebookRevision.objects.filter(notebook_id=notebook.id).first()
+        if (
+            serializer.validated_data["title"] == last_revision.title
+            and serializer.validated_data["content"] == last_revision.content
+        ):
+            return (False, NotebookRevisionDetailSerializer(last_revision).data)
```
robhudson: This smells like something that could live in the validation step of the serializer to me? I'm imagining there's no case where we'd want to save a revision that has no change. Also as a validation step the API can return a non-200 response (a 4xx) that can then be used to display something in the UI. In other words, 200 is a successful save but a 400 is not.

server/notebooks/api_views.py
wlach: Hmm, I'm not sure how to do this, given that it requires the lookup of a relate piece of metadata? I agree this would be nice to define in something like a decorator, were that possible...

server/notebooks/api_views.py
wlach: Hmm, you might be right about this now that I think of it. When I wrote this I figured we would simply "accept" revisions that were unchanged and do nothing with them, but perhaps a 400 code would be better. Let me investigate that approach, we may need to make some minor modifications on the client for this...

wlach: Your proposed approach has the additional advantage of not requiring so much overloading of the view.

server/notebooks/api_views.py
robhudson: In a typical Django view, the decorator would get the same `*args` and `**kwargs` as the view so it would also have to do the `Notebook` query to get the `owner_id`. But as a DRF `ModelViewSet` I'm not sure either.

server/notebooks/api_views.py
wlach: I wound up reverting the changes here, I think it's reasonably ok to just leave things be. If you like we could do a followup with this check.

server/notebooks/api_views.py
wlach: Yes, this is better! Pushed a new revision with the suggested change.

wlach:
Ok, I think this is ready for another look.

@hamilton I know the frontend change is kind of a hack, but I'd rather not extensively change code we're about to refactor in other ways soon.

server/notebooks/serializers.py
```diff
@@ -51,6 +51,14 @@ class NotebookRevisionDetailSerializer(serializers.ModelSerializer):
     Details of a revision for a notebook (includes content)
     """
 
+    def validate(self, attrs):
+        last_revision = NotebookRevision.objects.filter(
+            notebook_id=self.context["notebook_id"]
+        ).first()
+        if attrs["title"] == last_revision.title and attrs["content"] == last_revision.content:
+            raise serializers.ValidationError("Revision unchanged from previous")
+        return serializers.ModelSerializer.validate(self, attrs)
```
robhudson: 👍

hamilton:
The code looks good to me and works as expected! I agree that we shouldn't put too much into the UX, but I do think saying "error saving notebook" might mislead users to think something is broken at a critical juncture. I would suggest in `app-message.jsx` including a `NO_CHANGES_TO_NOTEBOOK` item, and sending that as the `messageType` in `actions.js` if there are in fact no changes to the notebook. Below is how it might look:

In `actions.js`:

```javascript

function saveNotebookRequest(url, postRequestOptions, dispatch) {
  return fetchWithCSRFTokenAndJSONContent(url, postRequestOptions)
    .then(response => {
      if (!response.ok) {
        throw response;
      }
      return response.json();
    })
    .catch(err => {
      // this is a horrible and obvious hack that will go away
      // when we don't provide an easy interface for the user
      // to double save a revision (i.e. we have provided
      // continuous autosave)
      err.json().then(json => {
        const [message, messageType] =
          json.non_field_errors && json.non_field_errors.length
            ? [json.non_field_errors.join(" "), "NO_CHANGES_TO_NOTEBOOK"]
            : ["Error Saving Notebook.", "ERROR_SAVING_NOTEBOOK"];
        dispatch(
          updateAppMessages({
            message,
            messageType
          })
        );
      });
      throw new Error(err);
    });
}
```

in `app-message.jsx`:

```javascript
const MESSAGES = {
  NOTEBOOK_SAVED: "Notebook saved.",
  NO_CHANGES_TO_NOTEBOOK: "No changes to notebook.",
  ERROR_SAVING_NOTEBOOK: "Error saving notebook.",
  LOGGED_IN: "You are now logged in.",
  LOGGED_OUT: "You are now logged out.",
  LOGIN_FAILED: "Login failed.",
  LOGOUT_FAILED: "Logout failed."
};
```

Feel free to take or leave this bit of feedback! I don't want to add a new perceived rough edge for users if we don't have to.

hamilton:
fwiw I don't believe the new autosave functionality will force this situation too often, since it won't autosave if there are no changes, but it'll be worth testing these intersecting features.

wlach:
On reflection, I don't think we want to add a "Notebook unsaved" status, since that will so rarely come up in an autosave world (and if it does, it's not something we would want to notify the user about).

Instead, I updated the patch to just silently not error out if we have no changes to save (i.e. it claims to have saved something, even though nothing happened). Seems to be working well locally, I'm going to land this patch so we can move on to the next phase of client/server improvements around autosave

hamilton:
@wlach great points

PR #1583, opened 2019-03-05
Randomly generate initial notebook name (fixes #1541)

### LINKED ISSUE #1541: Randomly generate default notebook names?
We're seeing a lot of "Untitled Notebook" notebooks appear on extremely-alpha.iodide.io, now that people are slowly starting to discover the site. We probably should add a "published" field to the notebook so we can not index these kinds of test notebooks, but I wonder if it might also be useful/fun to randomly generate a name for a new notebook? Going with the iodide/scientific theme, I thought it might be fun to generate some random names based on molecules. Here's a simple notebook demonstrating the idea:

https://extremely-alpha.iodide.io/notebooks/302/

Sample output:

```
0: "dihydrogen triphosphide"
1: "trioxygen triargide"
2: "pentargon dihydride"
3: "pentargon tetrahydride"
4: "tetraoxygen pentargide"
5: "dihydrogen pentaphosphide"
6: "dioxygen dioxide"
7: "pentargon hexahydride"
8: "trioxygen tetraphosphide"
9: "hexaoxygen pentargide"
```

Thoughts?

wlach:
### Pull Request checklist


server/tests/test_notebook_view.py
```diff
@@ -31,6 +33,7 @@ def test_notebook_view(client, test_notebook):
 
 @pytest.mark.parametrize("logged_in", [True, False])
 def test_new_notebook_view(client, fake_user, logged_in):
+    random.seed(0)
```
wlach: I searched a bit for a pytest fixture that would do this for me, but didn't see anything, at least nothing simple.

server/notebooks/names.py
```diff
@@ -0,0 +1,26 @@
...
+    ("argon", "argide"),
+    ("phosphorous", "phosphide"),
+]
+PREFIXES = ["di", "tri", "tetra", "penta", "hexa"]
+
+
+def _get_elements(num, elements):
+    i = int(random.random() * len(elements))
+    element_part = elements[i][num]
+    prefix_part = PREFIXES[int(random.random() * len(PREFIXES))]
```
robhudson: I don't know if you like it better but there's also `random.choice(PREFIXES)`

server/notebooks/names.py
```diff
@@ -0,0 +1,26 @@
+import random
+
+ELEMENTS = [
+    ("oxygen", "oxide"),
+    ("hydrogen", "hydride"),
+    ("argon", "argide"),
+    ("phosphorous", "phosphide"),
```
robhudson: We should add `("carbon", "carbide")` so we can get some tetracarbon notebooks. ;)

server/notebooks/names.py
```diff
@@ -0,0 +1,26 @@
...
+    i = int(random.random() * len(elements))
+    element_part = elements[i][num]
+    prefix_part = PREFIXES[int(random.random() * len(PREFIXES))]
+    if element_part[0] == prefix_part[-1]:
+        val = prefix_part[:-1] + element_part
+    else:
+        val = prefix_part + element_part
+    if num == 0:
+        return val
+    return _get_elements(num - 1, elements[0 : i - 1] + elements[i + 1 : -1]) + " " + val
```
robhudson: I think a test around the recursiveness of this would be good too.

robhudson:
Very cool, I like it! 👍

bcolloran:
So this might be too dorky, or it might be fun...

from this glossary of chemical compounds -- https://en.wikipedia.org/wiki/Glossary_of_chemical_formulas -- i grepped out all of the iodine compounds:

https://gist.github.com/bcolloran/4b7428209681d471ae69d931411793bf

we could use that, or indeed we could sample from the whole list of compounds :-D
https://gist.github.com/bcolloran/6f3bfca829cae8c656e4f45577bc4351

wlach:
Sounds good @bcolloran, just using a flat set of names is simpler. I think I'll go with the iodine ones for now-- the second set seems a little unfiltered and has some that don't look/sound so great (e.g. `urea`). :)

server/notebooks/names.py
wlach: I didn't know about that, thanks!

server/notebooks/names.py
wlach: Would be a good idea, but I think we'll go with @bcolloran's plan of using a random set of iodine compounds. :)

PR #1589, opened 2019-03-07
Just alpha

bcolloran:
@wlach and @hamilton -- removing the explicit `extremely-alpha` from our urls, b/c the redirect to the subdomain works fine, and hey let's say it loud and proud: we're just regular alpha at this point :-)

Also one string change in here. My method was: find-in-all-files for "extremely". Only found hits in 3 files.

Any thing i'm not thinking of here?

### Pull Request checklist

tests and docs not needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # (Unreleased; add upcoming change notes here)
+
 - iodide notebooks now automatically save to the server
 - fix bug where we should incorrect revision save times in history viewer,
   also show revision save time down to the second level
+- alpha now less extreme
 
 # 0.3.0 (2019-03-06)
 
--- a/src/eval-frame/components/panes/onboarding-content.jsx
+++ b/src/eval-frame/components/panes/onboarding-content.jsx
@@ -92,10 +92,10 @@ export default class OnboardingContent extends React.Component {
           <Element key="examples">
             <ElementTitle>Start with a Template</ElementTitle>
             <ElementBody>
-              <ElementBlockLink href="https://extremely-alpha.iodide.io/tryit">
+              <ElementBlockLink href="https://iodide.io/tryit">
                 Javascript starter
               </ElementBlockLink>
-              <ElementBlockLink href="https://extremely-alpha.iodide.io/notebooks/222/">
+              <ElementBlockLink href="https://iodide.io/notebooks/222/">
                 Pyodide starter
               </ElementBlockLink>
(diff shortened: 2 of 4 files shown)


src/eval-frame/components/panes/onboarding-content.jsx
```diff
@@ -92,21 +92,21 @@ export default class OnboardingContent extends React.Component {
           <Element key="examples">
             <ElementTitle>Start with a Template</ElementTitle>
             <ElementBody>
-              <ElementBlockLink href="https://extremely-alpha.iodide.io/tryit">
+              <ElementBlockLink href="https://iodide.io/tryit">
```
wlach: We should just not include the URLs here at all -- we can just use a relative path to the resource. e.g. `<ElementBlockLink href="/tryit">` should work just fine.

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # (Unreleased; add upcoming change notes here)
+
 - iodide notebooks now automatically save to the server
 - fix bug where we should incorrect revision save times in history viewer,
   also show revision save time down to the second level
+- alpha now less extreme
```
wlach: If you follow my suggestion below, you can just say something like "stop hardcoding server name in demo notebooks"

src/eval-frame/components/panes/onboarding-content.jsx
bcolloran: ah thank you! yes, much better idea, i'll make those changes

CHANGELOG.md
bcolloran: i thought this would be a cute in-joke for us :-)

src/eval-frame/components/panes/onboarding-content.jsx
bcolloran: hmmm, this actually may not be the best way to do this in the short term: it will break the demo links on eg the internal mozilla instance. i think we should keep the links to `iodide.io` for now, and we can revisit this when we eventually get around to a better system for doing featured notebooks.

CHANGELOG.md
wlach: Ok leave it in then :)

wlach:
Actually could you make the `/tryit/` link relative? That should not be instance-specific.

bcolloran:
oops this window had not refreshed when i pressed the merge button, i missed that last request Will.

Issue #1594, opened 2019-03-07
move info about cursor position to redux store, and generally decouple the app from codemirror


bcolloran:
we should move the current cursor position into the redux store and fire an action to update it whenever the cursor moves, instead of just checking the cursor position inside of the evaluateText action. this will more cleanly decouple the actions from the internals of the editor, which will (1) push this into our state tree, where we have more kinds of sanity checking and thus bugs like this are easier to to catch in advance, and (2) make it easier to migrate to Monaco down the line. (I don't know why i didn't do this to begin with -- this was my bad during the jsmd migration)

hamilton:
@bcolloran can we close this issue or repurpose it?

bcolloran:
ah yes, this is done via #1597, thanks for catching that hamilton!

PR #1596, opened 2019-03-09
Cursor position in state

bcolloran:
part 1 of #1594 fix, moves stuff about cursor position to redux.

there is one notably less than ideal part of this, which is `editorCursorForceUpdate`. this is a weird thing to have in the redux store, but this was actually the least side-effecty way i could find to do force the cursor to update in the case of move triggered outside of code mirror (this whole thing is required because codeMirror is not a fully controlled component, so setting cursor position with props doesn't work other than at initialization, and there is a race condition between when the "UPDATE_CURSOR" actions get back to the component from redux and the position already set internally  within codemirror. (i suspect the same will be true in monaco.))

@hamilton as with all code editing things, please kick the tires on this one throughly -- do all the editing and entry operations you can think of, use all the eval shortcuts and other editor shortcuts, etc.

### Pull Request checklist

no doc or changelog entry needed, no user-facing changes

### DIFF
--- a/src/actions/__tests__/cursor-actions.test.js
+++ b/src/actions/__tests__/cursor-actions.test.js
@@ -1,0 +1,79 @@
+import configureMockStore from "redux-mock-store";
+import thunk from "redux-thunk";
+import { moveCursorToNextChunk, updateEditorCursor } from "../actions";
+
+const middlewares = [thunk];
+const mockStore = configureMockStore(middlewares);
+
+describe("updateEditorCursor - returns correct action", () => {
+  it("should create an action to update the cursor (NO force param)", () => {
+    const [editorCursorLine, editorCursorChar] = [1, 2];
+    const expectedAction = {
+      type: "UPDATE_CURSOR",
+      editorCursorLine,
+      editorCursorChar,
+      editorCursorForceUpdate: false
+    };
+    expect(updateEditorCursor(editorCursorLine, editorCursorChar)).toEqual(
+      expectedAction
+    );
+  });
+
+  it("should create an action to update the cursor (with force param)", () => {
+    const [editorCursorLine, editorCursorChar, editorCursorForceUpdate] = [
+      1,
+      2,
+      true
+    ];
+    const expectedAction = {
+      type: "UPDATE_CURSOR",
+      editorCursorLine,
+      editorCursorChar,
+      editorCursorForceUpdate
+    };
+    expect(
+      updateEditorCursor(
+        editorCursorLine,
+        editorCursorChar,
+        editorCursorForceUpdate
+      )
+    ).toEqual(expectedAction);
+  });
+});
+
+// /*  eslint-disable no-unused-vars */
+// const window = {
+//   /*  eslint-enable no-unused-vars */
+//   getDoc: () => ({ somethingSelected: () => false })
+// };
+window.ACTIVE_CODEMIRROR = {
+  getDoc: () => ({ somethingSelected: () => false })
+};
+
+describe("moveCursorToNextChunk dispatches correct actions", () => {
+  it("no selection, and one chunk", () => {
+    const [editorCursorLine, editorCursorChar] = [1, 2];
+
+    const store = mockStore({
+      editorCursorLine,
+      editorCursorChar,
+      jsmdChunks: [{ startLine: 0, endLine: 10 }]
+    });
+
+    const expectedActions = [
+      {
+        type: "UPDATE_CURSOR",
+        editorCursorLine: 11,
+        editorCursorChar: 0,
+        editorCursorForceUpdate: true
+      }
+    ];
+
+    // return store.dispatch(moveCursorToNextChunk()).then(() => {
+    //   // return of async actions
+    //   expect(store.getActions()).toEqual(expectedActions);
+    // });
+    store.dispatch(moveCursorToNextChunk());
+    expect(store.getActions()).toEqual(expectedActions);
+  });
+});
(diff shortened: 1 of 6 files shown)


src/actions/actions.js
```diff
@@ -188,15 +188,30 @@ export function addLanguage(languageDefinition) {
...
     const { jsmdChunks, kernelState } = getState();
     if (kernelState !== "KERNEL_BUSY") dispatch(setKernelState("KERNEL_BUSY"));
     const cm = window.ACTIVE_CODEMIRROR;
     const doc = cm.getDoc();
     if (!doc.somethingSelected()) {
-      const { line } = doc.getCursor();
-      const activeChunk = getChunkContainingLine(jsmdChunks, line);
+      const activeChunk = getChunkContainingLine(
+        jsmdChunks,
+        getState().editorCursorLine
```
hamilton: we are already calling `getState` / destructuring that above - can we use that same destructuring for `editorCursorLine`?

src/actions/actions.js
```diff
@@ -216,7 +231,7 @@ export function moveCursorToNextChunk() {
     let targetLine;
 
     if (!doc.somethingSelected()) {
-      targetLine = doc.getCursor().line;
+      targetLine = getState().editorCursorLine;
```
hamilton: same point as above

src/actions/__tests__/cursor-actions.test.js
```diff
@@ -0,0 +1,96 @@
...
+    };
+    expect(
+      updateEditorCursor(
+        editorCursorLine,
+        editorCursorChar,
+        editorCursorForceUpdate
+      )
+    ).toEqual(expectedAction);
+  });
+});
```
wlach: I would personally err on taking out the above tests, I don't think testing object creation in this way provides much value: this is *very* unlikely to break in an unexpected way and the same functionality is implicitly tested by the subsequent tests.

src/components/jsmd-editor.jsx
```diff
@@ -61,6 +70,11 @@ class JsmdEditorUnconnected extends React.Component {
     this.props.actions.updateJsmdContent(content);
   }
 
+  updateCursor(editor) {
+    const { line, ch } = editor.getCursor();
+    this.props.actions.updateEditorCursor(line, ch);
```
hamilton: I know this is an older pattern and this file has used `bindActionCreators` for a while, but it might be better to explicitly pick which actions we want in `mapDispatchToProps`.

src/components/jsmd-editor.jsx
```diff
@@ -138,14 +152,15 @@ class JsmdEditorUnconnected extends React.Component {
     //   },
```
hamilton: let's delete this commented-out code block above if we're not going to use it

src/state-schemas/state-schema.js
```diff
@@ -102,6 +102,20 @@ export const stateProperties = {
...
+  editorCursorChar: {
+    type: "integer",
+    default: 0
+  },
+  editorCursorForceUpdate: {
+    // if this is true when the editor recieves it as props,
+    // then the editor must reposition the cursor using internal editor APIs
+    type: "boolean",
+    default: false
+  },
```
wlach: I wonder if it would be better to just have an `editor` object that we can stuff properties into.

src/actions/__tests__/cursor-actions.test.js
```diff
@@ -0,0 +1,96 @@
+import configureMockStore from "redux-mock-store";
+import thunk from "redux-thunk";
+import { moveCursorToNextChunk, updateEditorCursor } from "../actions";
+
+const middlewares = [thunk];
+const mockStore = configureMockStore(middlewares);
+
+describe("updateEditorCursor - returns correct action", () => {
```
hamilton: let's make adding action tests a requirement for any additions going forward.

src/components/jsmd-editor.jsx
```diff
@@ -169,10 +184,13 @@ function mapStateToProps(state) {
...
   }
-
+  const { editorCursorLine, editorCursorChar, editorCursorForceUpdate } = state;
   return {
     content: state.jsmd,
-    editorOptions
+    editorOptions,
+    editorCursorLine,
+    editorCursorChar,
+    editorCursorForceUpdate
```
wlach: Is there any way you could add some tests for the jsmd editor here? Perhaps experiment with modifying the state and that it updates as expected? That seems much more likely to break that the redux action/store stuff that is currently being tested.

src/components/jsmd-editor.jsx
```diff
@@ -138,14 +152,15 @@ class JsmdEditorUnconnected extends React.Component {
     //   },
     // )
 
-    // FIXME: should set cursor position in redux store (see: https://github.com/iodide-project/iodide/issues/1568)
+    const { editorCursorLine, editorCursorChar } = this.props;
     return (
       <ReactCodeMirror
         editorDidMount={this.storeEditorInstance}
-        cursor={{ line: 1, ch: 1 }}
```
hamilton: the schema sets these values to 0 by default. does this matter? Should it have been 0 originally?

src/actions/__tests__/cursor-actions.test.js
hamilton: I think I agree with @wlach's point above but still think that having test coverage is a positive signal. It's not always just about utility.

src/components/jsmd-editor.jsx
hamilton: good point

src/components/jsmd-editor.jsx
bcolloran: it should have been 0 originally

bcolloran:
thanks for the feedback guys! all the comments seemed to be about what i'd call stylistic issues rather than logical issues, so i'll merge this as is, but refer to these suggested changes as i revise pt 2. i think that will make it easier for you guys to review pt2 overall b/c it will be a smaller changeset.

PR #1603, opened 2019-03-12
Move version.json generation to deploy job

jasonthomas:
In #1548 a change was made that stopped using the cached docker image generated in the `build` job. In the `build` job there is a task that generates a version.json file which is used for image verification as part of deployment into our environments. Since this file is no longer available in the docker image the deployments are failing. I've moved this task into the `deploy` job which should fix this issue.

https://bugzilla.mozilla.org/show_bug.cgi?id=1534646


wlach:
Thanks for the speedy fix!

PR #1604, opened 2019-03-12
Fix the npm built assets for deploy

robhudson:


wlach:
Let's try this, it should fix the internal staging server.

PR #1597, opened 2019-03-10
Cursor in state pt 2

bcolloran:
follow up from #1596 -- this is builds on the changes in that PR; @hamilton if that looked ok, take a look at this next.

this will get us almost all the way there for #1594 

### Pull Request checklist

- [ ] **Documentation**: If this feature has or requires documentation, the relevant docs have been updated.
- [ ] **Changelog**: This PR updates the [changelog](../CHANGELOG.md) with any user-visible changes.
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not

no user-facing changes, no doc or changelog entry needed

PR #1625, opened 2019-03-14
Cursor in state 2 rebase

bcolloran:
### Pull Request checklist

### DIFF
--- a/src/components/jsmd-editor.jsx
+++ b/src/components/jsmd-editor.jsx
@@ -53,26 +53,11 @@ class JsmdEditorUnconnected extends React.Component {
   }
 
   componentDidUpdate() {
-    // need to
-    // const cm = this.editor;
-    // const { editorCursorLine, editorCursorChar } = this.props;
-    // if (
-    //   cm.getCursor().line !== editorCursorLine ||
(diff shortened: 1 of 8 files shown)


src/components/jsmd-editor.jsx
```diff
@@ -141,6 +152,7 @@ class JsmdEditorUnconnected extends React.Component {
   };
 
   render() {
+    // FIXME: restore autocomplete to working order
```
bcolloran: @hamilton i added a fixme for the commented out block of code having to do with autocomplete. if we stick with codemirror, this will be helpful in reimplementing that; if we switch, it's irrelevant anyway

bcolloran:
@wlach and @hamilton this is ready for review (replaces #1597)

there is some churn in this PR around renaming:
editorCursorLine -> line,
      editorCursorChar -> col,
      editorCursorForceUpdate -> forceUpdate, and moving them to "editorCursor" in state (per will's suggestion), and a bit of other code churn around  some of the other fixes i put in from your last review, but hopefully that won't be too distracting. the conceptual thing to focus on here is that in this PR i move info about the editor selection(s) to the redux state. pretty straightforward really. lmk what you think, thanks!

PR #1633, opened 2019-03-15
Uncircle imports

bcolloran:
@hamilton this should be another quick one. fixes an import cycle by having the editor use same messagePasser intermediary that we already set up in the evalframe a while back.

(also renamed `messagePasser` -> `messagePasserEval` for ctrl-f later on, you can easily ignore that)

### Pull Request checklist

this is quite a minor change -- no user facing anythng, no truly new code, just clean up

### DIFF
--- a/src/port-to-eval-frame.js
+++ b/src/port-to-eval-frame.js
@@ -35,7 +35,7 @@ function receiveMessage(event) {
     const { messageType, message } = event.data;
     switch (messageType) {
       case "ADD_TO_EVALUATION_QUEUE": {
-        evalQueue.evaluate(message, store.dispatch);
+        evalQueue.evaluate(message);
         store.dispatch(setKernelState("KERNEL_BUSY"));
         break;
       }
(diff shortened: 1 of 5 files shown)


src/port-to-eval-frame.js
```diff
@@ -1,11 +1,11 @@
 /* global IODIDE_EVAL_FRAME_ORIGIN  */
 
 import Mousetrap from "mousetrap";
-import { store } from "./store";
 import { addLanguage, setKernelState } from "./actions/actions";
 import { genericFetch as fetchFileFromServer } from "./tools/fetch-tools";
 import evalQueue from "./actions/evaluation-queue";
 import validateActionFromEvalFrame from "./actions/eval-frame-action-validator";
+import messagePasserEd from "./redux-to-port-message-passer";
```
hamilton: can this be `messagePasserEditor` or something? the `Ed` took me a few seconds, and would probably be confusing to a new contributor

bcolloran: yep that seems sensible -- i'll make the string change and merge

PR #1640, opened 2019-03-15
Remove heroku-run-build-script directive

openjck:
According to the Heroku blog post which announced this feature, this
directive is not needed after March 11, 2019.

https://devcenter.heroku.com/changelog-items/1557


wlach:
Ah good catch! thanks.

PR #1642, opened 2019-03-15
Add token-based auth and allow authorized users to create notebooks on another's behalf

wlach:
This should be enough server-side functionality to support mozilla/redash#426.

### Pull Request checklist


wlach:
Needs some cleanup before I ask for review, but this is looking pretty promising so far.

server/notebooks/api_views.py
```diff
@@ -32,12 +34,22 @@ def perform_destroy(self, instance):
 
     def perform_create(self, serializer):
         with transaction.atomic():
+            print(self.request.data)
```
openjck: 👆 😄

server/settings.py
```diff
@@ -92,6 +100,7 @@
     "django.middleware.common.CommonMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
+    # "server.auth_delegator.middleware.AuthDelegatorMiddleware",
```
openjck: ☝️ 😄

openjck:
Thanks for picking this up for me. I really would have struggled to figure this out.

How would this be used? It sounds like there would be a special user, which has an API token and which is allowed to create notebooks on behalf of other users and Redash would only ever need to know the token of that one user. Is that correct?

wlach:
> Thanks for picking this up for me. I really would have struggled to figure this out.

I had several false starts on this myself today, the DRF documentation on how authentication works (and cooperates with the rest of Django) is not that straightforward.

> How would this be used? It sounds like there would be a special user which has an API token and which is allowed to create notebooks on behalf of other users; Redash would only ever need to know the token of that one user. Is that correct?

Yes, exactly.

wlach:
Going to try creating an admin view for managing this stuff. This may also include a facility for manually creating users, which could be useful in other contexts as well.

wlach:
> Going to try creating an admin view for managing this stuff. This may also include a facility for manually creating users, which could be useful in other contexts as well.

Actually this was mostly done, we just need to be sure to give users who need to do this the `is_staff` permission.

server/notebooks/api_views.py
wlach: oops! good eye, thanks

wlach:
hey @openjck / @robhudson -- would love your opinions on this, as I think this is basically working now. The details of how to use this new functionality are in `docs/` -- hopefully things are clear enough there.

bcolloran:
nice job, thanks guys! :clap:

docs/server-admin.md
```diff
@@ -0,0 +1,37 @@
+# The administration panel
+
+Iodide has an administration view under `/admin`, based on the [Django
+admin](https://docs.djangoproject.com/en/2.1/ref/contrib/admin/) interface.
+
+To log into this interface, a user must either have the `is_staff` or `is_superuser` permission set. Most operations (including the one currently
+described here) require `is_superuser`.
+
+An existing super user can grant old users super-user permissions via
+the panel. Or, if no super user is currently configured, you can perform this operation manually via a SQL session (`./manage.py dbshell` from the console):
```
openjck: Not a big deal, but a *make dbshell* command would be handy. :smiley:

docs/server-admin.md
```diff
@@ -0,0 +1,37 @@
+# The administration panel
+
+Iodide has an administration view under `/admin`, based on the [Django
+admin](https://docs.djangoproject.com/en/2.1/ref/contrib/admin/) interface.
+
+To log into this interface, a user must either have the `is_staff` or `is_superuser` permission set. Most operations (including the one currently
```
openjck: This wasn't true for me. I'm not allowed to view the */admin* page when only `is_superuser` is set.

Here's a complete breakdown of what I'm seeing: 

|State|Result|
|---|---|
|Only `is_superuser` set to true|Error when trying to access the */admin* page|
|Only `is_staff` set to true|The */admin* page loads and shows only a *Notebooks* link|
|Both set to true|The */admin* page loads with the following links: *Groups*, *Notebooks*, *Associations*, *Nonces*, *User social auths*|

docs/server-admin.md
```diff
@@ -0,0 +1,37 @@
...
+described here) require `is_superuser`.
+
+An existing super user can grant old users super-user permissions via
+the panel. Or, if no super user is currently configured, you can perform this operation manually via a SQL session (`./manage.py dbshell` from the console):
+
+```sql
+update base_user set is_superuser='t' where username=MY_USER_NAME;
```
openjck: If my above comment is correct this will need to be updated as well.

docs/server-admin.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Authentication tokens
+
+It may, in some cases, by desirable to create special users which can
+create notebooks on behalf of others (e.g. bots). For this task, we
+have configured the iodide server to allow for the notebook creation API to be accessed using django-rest-framework's [token authentication])(https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication)
+
+To take advantage of this:
+
+1. Go to the server's admin panel (as a user with root permissions, see above) and create a new user with the "can create on behalf of
```
openjck: Even with `is_staff` and `is_superuser` permissions, I don't see a way to create new users.

Screenshot showing what the admin interface looks like to me

openjck: I created this account / log into it with GitHub, if that matters.

docs/server-admin.md
openjck: Er... I was on the wrong branch. 🙄

openjck: This wasn't true for me. I get a *Permission Denied* error when only `is_superuser` is set.

Here's what I'm seeing in more detail:

|State|Result|
|---|---|
|Only `is_superuser` is set to true|*You are authenticated as openjck, but are not authorized to access this page. Would you like to login to a different account?*|
|Only `is_staff` is set to true|I can see the admin interface and the following links are present: *Users*, *Notebooks*|
|Both are set to true|I can see the admin interface and the following links are present: *Tokens*, *Users*, *Notebooks*, *Associations*, *Nonces*, *User social auths*|

openjck:
I added two line comments, which are visible above. (I also added a few line comments before that, but later realized I was on the wrong branch at the time. You may get email notifications about them but they've been deleted and should no longer be visible here.)

The documentation makes sense to me and this behaves as expected, too. Using [HTTPie](https://httpie.org/), I can create a notebook with this command:

`http POST http://localhost:8000/api/v1/notebooks/ 'Authorization: Token [value]' owner=[username] title='API-generated notebook' content='Text here'`

I notice that `\n` isn't converted to a linebreak, so when the POST body contains `content='Text\nhere'`, the `\n` shows up literally in the resulting notebook. When there is a true linebreak in the value of `content`, that linebreak is persisted in the resulting notebook, but JSON values cannot contain linebreaks.

docs/server-admin.md
wlach: Feel free to send in a PR :)

docs/server-admin.md
wlach: Ah ok, I guess you need is_staff permission to use the admin interface at all. I'll update the docs.

wlach:
@openjck I think that comes from the command line tool that you're using. If i post to the endpoint via the rest framework view (http://localhost:8000/api/v1/notebooks/), it works fine

wlach:
Noticed some issues while testing (users should be able to set "owner" to themselves, or set "owner" to null when creating a notebook), fixed those along with doc updates.

openjck:
> @openjck I think that comes from the command line tool that you're using. If i post to the endpoint via the rest framework view (http://localhost:8000/api/v1/notebooks/), it works fine

Gotcha. That must be it. Thanks for looking into that.

PR #1644, opened 2019-03-17
(WIP) Simplify eval code, remove dispatch from eval frame

bcolloran:
builds on  #1643 (superset of those changes)

this substantially simplifies the eval-frame's evaluation code, mostly by removing its use of dispatch in almost all cases. This is done by de-thunking a number of functions having to do with evaluation. doing this clears the way for removing the rest of the instances of dispatch from other parts of the eval frame.

the main file with substantive logical changes is src/eval-frame/actions/actions.js. a bunch of the stuff at the top that appears to have been cut has really just moved to other files, but `evaluateCode` and `evaluateText` have major changes.

i've tried to leave many comments to guide you through this review. i think it may not actually be too too bad, at least i hope not.

there are some more architecture changes that need to be in place before test writing begins, but this PR already gets us very close to the longstanding dream of shrinking our use of redux in the eval frame to almost nothing, and finally removes the need for the heinous hack of having a special reducer that forwards actions from the eval frame to the editor

### Pull Request checklist

### DIFF
--- a/src/actions/__tests__/eval-actions.test.js
+++ b/src/actions/__tests__/eval-actions.test.js
@@ -1,7 +1,8 @@
 import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
 
-import { evaluateNotebook, nonRunnableChunkType } from "../actions";
+import { evaluateNotebook } from "../eval-actions";
+import { NONCODE_EVAL_TYPES } from "../../state-schemas/state-schema";
 
 const mockStore = configureMockStore([thunk]);
 
@@ -72,7 +73,7 @@ describe("evaluateNotebook - correct evaluations", () => {
     expect(mockEvalQueue.evaluate.mock.calls.length).toBe(2);
   });
 
-  nonRunnableChunkType.forEach(chunkType => {
+  NONCODE_EVAL_TYPES.forEach(chunkType => {
     it(`doesn't call evalQueueInstance.evaluate for chunktype: ${chunkType}`, async () => {
       testState.jsmdChunks[0].chunkType = chunkType;
       store = mockStore(testState);
--- a/src/actions/__tests__/verify-action-store-consistency.test.js
+++ b/src/actions/__tests__/verify-action-store-consistency.test.js
@@ -1,6 +1,6 @@
 import { store } from "../../store";
 import * as actions from "../actions";
-import { evaluateText, evaluateNotebook } from "../eval-actions";
+import { evaluateNotebook } from "../eval-actions";
 
 import { stateProperties } from "../../state-schemas/state-schema";
 import { SchemaValidationError } from "../../reducers/create-validated-reducer";
--- a/src/actions/__tests__/verify-action-store-consistency.test.js
+++ b/src/actions/__tests__/verify-action-store-consistency.test.js
@@ -1,5 +1,7 @@
 import { store } from "../../store";
 import * as actions from "../actions";
+import { evaluateText, evaluateNotebook } from "../eval-actions";
+
 import { stateProperties } from "../../state-schemas/state-schema";
 import { SchemaValidationError } from "../../reducers/create-validated-reducer";
 import { languageDefinitions } from "../../state-schemas/language-definitions";
@@ -97,7 +99,7 @@ describe("make sure action creators leave store in a consitent state", () => {
   });
 
   it("evaluateNotebook", () => {
-    expect(() => store.dispatch(actions.evaluateNotebook())).not.toThrow();
+    expect(() => store.dispatch(evaluateNotebook())).not.toThrow();
   });
 
   it("loginSuccess", () => {
--- a/src/actions/actions.js
+++ b/src/actions/actions.js
@@ -21,13 +21,7 @@ import { loginToServer, logoutFromServer } from "../tools/login";
 import { fetchWithCSRFTokenAndJSONContent } from "./../shared/fetch-with-csrf-token";
 
 import { jsmdParser } from "./jsmd-parser";
-import {
-  getChunkContainingLine,
-  selectionToChunks,
-  removeDuplicatePluginChunksInSelectionSet
-} from "./jsmd-selection";
-
-import evalQueue from "./evaluation-queue";
+import { getChunkContainingLine } from "./jsmd-selection";
 
 import createHistoryItem from "../tools/create-history-item";
 
@@ -198,36 +192,6 @@ export function updateEditorSelections(selections) {
   };
 }
 
-export function evaluateText() {
-  return (dispatch, getState) => {
-    const {
-      jsmdChunks,
-      kernelState,
-      editorSelections,
-      editorCursor
-    } = getState();
-
-    if (kernelState !== "KERNEL_BUSY") dispatch(setKernelState("KERNEL_BUSY"));
-
-    if (editorSelections.length === 0) {
-      const activeChunk = getChunkContainingLine(jsmdChunks, editorCursor.line);
-      evalQueue.evaluate(activeChunk, dispatch);
-    } else {
-      const selectionChunkSet = editorSelections
-        .map(selection => ({
-          start: selection.start.line,
-          end: selection.end.line,
-          selectedText: selection.selectedText
-        }))
-        .map(selection => selectionToChunks(selection, jsmdChunks))
-        .map(removeDuplicatePluginChunksInSelectionSet());
-      selectionChunkSet.forEach(selection => {
-        selection.forEach(chunk => evalQueue.evaluate(chunk, dispatch));
-      });
-    }
-  };
-}
-
 export function moveCursorToNextChunk() {
   return (dispatch, getState) => {
     const {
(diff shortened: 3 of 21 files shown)


src/actions/__tests__/eval-actions.test.js
```diff
@@ -1,7 +1,8 @@
 import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
 
-import { evaluateNotebook, nonRunnableChunkType } from "../actions";
+import { evaluateNotebook } from "../eval-actions";
+import { NONCODE_EVAL_TYPES } from "../../state-schemas/state-schema";
```
bcolloran: just removing a duplicate def in favor of using the def in the state-schema

src/actions/__tests__/verify-action-store-consistency.test.js
```diff
@@ -1,5 +1,7 @@
 import { store } from "../../store";
 import * as actions from "../actions";
+import { evaluateText, evaluateNotebook } from "../eval-actions";
```
bcolloran: no real change, just moved the eval functions to a separate file

src/actions/actions.js
```diff
@@ -198,36 +192,6 @@ export function updateEditorSelections(selections) {
   };
 }
 
-export function evaluateText() {
```
bcolloran: no real changes in this file, just moved these fns to a separate file

src/actions/console-actions.js
```diff
@@ -0,0 +1,23 @@
+import evalQueue from "./evaluation-queue";
+
+export function evalConsoleInput(consoleText) {
```
bcolloran: moved this logic over from the eval frame; this is now triggered by a message from the eval frame

src/actions/eval-actions.js
```diff
@@ -0,0 +1,61 @@
+import {
```
bcolloran: just moved these functions to a new file, no real changes here

src/actions/task-definitions.js
```diff
@@ -2,6 +2,7 @@ import UserTask from "./user-task";
 import ExternalLinkTask from "./external-link-task";
 import { store } from "../store";
 import * as actions from "./actions";
+import { evaluateText, evaluateNotebook } from "./eval-actions";
```
bcolloran: no real updates in this file, just replaced ref b/c of file move.

src/eval-frame/actions/editor-message-senders.js
```diff
@@ -0,0 +1,9 @@
+import messagePasserEval from "../../redux-to-port-message-passer";
+
+export function sendStatusResponseToEditor(status, evalId) {
```
bcolloran: moved this fn over from eval-frame/actions

src/eval-frame/actions/editor-message-senders.js
```diff
@@ -0,0 +1,9 @@
...
+  messagePasserEval.postMessage("EVALUATION_RESPONSE", { status, evalId });
+}
+
+export function sendActionToEditor(action) {
+  messagePasserEval.postMessage("REDUX_ACTION", action);
+}
```
bcolloran: this is just a little convenience fn that made it easy to migrate from `dispatch(actionCreatorFnc(...))` to `sendActionToEditor(actionCreatorFnc(...))`. may not keep this in the long run, but it made the migration flow tractable

src/eval-frame/components/panes/console/console-input-field.jsx
```diff
@@ -154,7 +156,9 @@ export const ConsoleInputMessagePasser = connectMessagePassers(
       postActionToEditor({
         type: "CONSOLE_HISTORY_MOVE",
         consoleCursorDelta
-      })
+      }),
+    evalConsoleInput: consoleText =>
+      postMessageToEditor("CONSOLE_NEEDS_EVALUATION", consoleText)
```
bcolloran: the substantive change in this file is that this component no longer dispatches, but rather sends a message  to the editor to trigger a console eval

src/eval-frame/components/panes/console/console-language-menu.jsx
```diff
@@ -83,7 +83,8 @@ const ConsoleLanguageMenuUnconnected = ({
             <MenuItem
               key={language.languageId}
               onClick={onMenuClickCreator(
-                setConsoleLanguage,
+                languageId =>
+                  sendActionToEditor(setConsoleLanguage(languageId)),
```
bcolloran: as above, the substantive change in this file is that this component no longer dispatches, but rather sends a action to the editor to set the console language

src/eval-frame/initialize-user-variables.js
```diff
@@ -1,5 +1,6 @@
 import { updateUserVariables } from "./actions/actions";
+import { sendActionToEditor } from "./actions/editor-message-senders";
 
-export default store => {
-  store.dispatch(updateUserVariables());
+export default () => {
+  sendActionToEditor(updateUserVariables());
```
bcolloran: no longer depends on store or dispatches to store, just sends the editor a action with the list of variables

src/eval-frame/index.jsx
```diff
@@ -48,3 +47,5 @@ render(
   </Provider>,
   document.getElementById("view-mode-styles")
 );
+
+initializeUserVariables();
```
bcolloran: remove `store` param, and move after render (since this is not needed immediately when the window opens)

src/eval-frame/store.js
```diff
@@ -29,6 +29,4 @@ const initialState = evalFrameStateSelector(newNotebook());
 
 const store = createStore(reducer, initialState, enhancer);
 
-const { dispatch } = store;
-
-export { store, dispatch };
```
bcolloran: `dispatch` export not used anywhere

src/eval-frame/tools/fetch-file-from-parent-context.js
```diff
@@ -30,6 +30,6 @@ export default async function fetchFileFromParentContext(path, fetchType) {
     // resolve and reject are handled in port-to-editor.js when
     // the file is received by the editor.
     addResolvers(path, resolve, reject);
-    postMessageToEditor("REQUEST_FETCH", { path, fetchType });
+    messagePasserEval.postMessage("REQUEST_FETCH", { path, fetchType });
```
bcolloran: changing this to messagePasserEval resolved an import cycle

src/port-to-eval-frame.js
```diff
@@ -38,6 +39,10 @@ function receiveMessage(event) {
         messagePasserEditor.dispatch(setKernelState("KERNEL_BUSY"));
         break;
       }
+      case "CONSOLE_NEEDS_EVALUATION": {
+        messagePasserEditor.dispatch(evalConsoleInput(message));
+        break;
+      }
```
bcolloran: this adds a listener on the editor side that triggers eval of the console

src/eval-frame/actions/fetch-cell-eval-actions.js
```diff
@@ -99,71 +102,57 @@ export async function handleFetch(fetchInfo) {
 }
 
 export function evaluateFetchText(fetchText, evalId) {
-  return dispatch => {
```
bcolloran: there is actually not much changed in this file, but the diff is garbage. the main thing is that this is no longer a redux-thunk, because it no longer needs to be dispatched -- it's just a plain function. the diff looks like a lot has changed b/c the indents of all these lines changed

src/eval-frame/actions/language-actions.js
```diff
@@ -103,75 +103,43 @@ function loadLanguagePlugin(pluginData, dispatch) {
 }
 
 export function evaluateLanguagePlugin(pluginText, evalId) {
-  return dispatch => {
```
bcolloran: as above, this is just a bad diff -- the important change is that this is not a thunk, so it doesn't return a function

src/eval-frame/reducers/reducer.js
```diff
@@ -1,29 +1,3 @@
...
-function replaceStateFromEditor(state, action) {
+export default function reducer(state, action) {
   return action.type === "REPLACE_STATE" ? action.state : state;
 }
-
-// this function forwards actions to the editor
-// FIXME THIS IS CRITICAL
-// for securty reasons, any actions forwarded must be whitelisted
-// and verified against schema on the editor side.
-const actionForwarder = (state, action) => {
```
bcolloran: hooray! finally done with the hideous action forwarder!!!

src/eval-frame/reducers/reducer.js
```diff
@@ -1,29 +1,3 @@
-// /*  global IODIDE_BUILD_MODE */
-import { postActionToEditor } from "../port-to-editor";
-
-function reduceReducers(...reducers) {
-  return (previous, current) =>
-    reducers.reduce((p, r) => r(p, current), previous);
-}
-
-function replaceStateFromEditor(state, action) {
+export default function reducer(state, action) {
```
hamilton: this is the real story

src/eval-frame/actions/actions.js
```diff
@@ -22,10 +22,6 @@ initialVariables.add("CodeMirror");
 initialVariables.add("FETCH_RESOLVERS");
```
hamilton: We should probably break the `initialVariables` obj + associated function into a separate file as well, either in this PR or elsewhere. Mind adding a `FIXME` if you don't want to tackle this now?

src/eval-frame/components/panes/console/console-input-field.jsx
hamilton: Is there a reason we're not going through `sendActionToEditor` at this point instead? Seems like it would make it easier to follow along vs. forcing me to look up the diffs b/t `postMessageToEditor` and `sendActionToEditor`.

bcolloran:
spoke to hamilton-- 100% agree with these points; will add them to the next phase of this refactor. Merging this now to reduce merge conflict risk with other PRs, and to simplify review of future PRs, and because there were no concerns to the overall direction, just a desire to keep pushing father (so it seems reasonable to land this incremental step)

src/eval-frame/actions/language-actions.js
```diff
@@ -103,75 +103,43 @@ function loadLanguagePlugin(pluginData, dispatch) {
...
-      dispatch
-    ).then(() => state.languageDefinitions[languageId]);
-  }
-  // There is neither a loaded language or a predefined definition that matches...
-  // FIXME: It would be hard to get here in the UX, but with direct JSMD
-  // editing you could...
-  return new Promise((resolve, reject) => reject());
+  return loadLanguagePlugin(pluginData)
+    .then(() => {
+      sendStatusResponseToEditor("SUCCESS", evalId);
```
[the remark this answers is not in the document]
bcolloran: Ah, i think there is a small confusion here: the status response it not a redux action that is being sent up, it is it's own message type that the evalQueue object responds to in order to resolve its promise chain, which is why it still needs to be separate

src/eval-frame/components/panes/console/console-input-field.jsx
bcolloran: as in the case above, the message sent here is not a redux action that is being sent up, it is it's own message type that causes the message receiver in the editor to dispatch on the editor side.

the fact that this stuff is so confusing indicates the need for more refactor!

PR #1651, opened 2019-03-19
PROPOSAL: iodide.file API

hamilton:
### Pull Request checklist

This is an API proposal for `iodide.file`, which gives programmatic file read-write access in our "virtual" file environment (the server). It proposes `iodide.file.save`, `iodide.file.load`, `iodide.file.delete`, `iodide.file.list`, `iodide.file.exists`, and `iodide.file.lastUpdated`.

The `iodide.file.load` API specifically is meant to mirror the fetch chunk API, generally, but only for uploaded files. The goal is **not** to replace the fetch chunks - rather, it's about unlocking the kind of file-ish workflows a lot of users might be expecting. For most use cases we should still be pushing users towards using the fetch chunk, because it is substantially more straightforward than using this API for basic uses.

I thought I'd try writing some docs around this proposed API to get the conversation going before actually implementing, so others can see what I had in mind before I did the actual work / suggest changes.

Some notes:

- this is likely blocked a bit by the file uploader branch that @openjck is working on, mostly because I think we will want to have each notebook have all the file names / metadata for files loaded in the store from the server, and he'll likely get to that in his work before I do.
- I think this PR also is blocked by fixing #597 so we can make sure to block further chunk evaluation on the Promises returned by `iodide.file.load`, `iodide.file.save`, and `iodide.file.delete`. After migrating to the cell-less flat-text world, I see more merit to this approach than I did before, and I think that if programmatic file access is what we get from making sequential these async return values, then so be it.
- @mdboom I'm especially curious to hear your thoughts and how this might interop with language plugins. I want to make it easy to use this stuff from within pyodide and other languages without introducing too much overhead.
- I would encourage everyone to also spend some time thinking about the implications of this direction. This means I can actually do a bunch of expensive computations in my notebook and cache them for others pretty easily. It also might open the door to do a "server-side" evaluation of an expensive notebook, especially if that could lead to generating cached data that makes notebooks load more quickly.
- (EDIT) I have not really considered what will happen if a user does not own the notebook and they are attempting to use `iodide.file.save` or `iodide.file.delete`. Perhaps there is an option we can include in those to ignore such calls, or just throw an error. Thoughts here would be appreciated.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -57,7 +57,7 @@ Returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guid
 
 `fileType` is the file type to handle. These are the same as the file types in [fetch chunks](https://iodide-project.github.io/docs/jsmd/#fetch-chunks-fetch): `js` (load this file as javascript, eg a library), `css` (load this file as a stylesheet, applying the styles to the report), `json` (load this file as json and parse into a javascript object), `text` (load this file as text), `blob` (load this file as binary).
 
-`variableName` (required for `json`, `text`, and `blob` file types, otherwise not applicable): the variable name in which to load the data.
+`variableName` (required for `json`, `text`, and `blob` file types, otherwise not applicable): the variable name in which to load the data, available in the browser `window` namespace.
 
 
 #### `iodide.file.load` patterns
--- a/docs/api.md
+++ b/docs/api.md
@@ -133,7 +133,7 @@ if (iodide.file.exists(FILENAME)) {
   iodide.file.load(FILENAME, 'text', 'evictionsData')
     .then(() => {
       evictionsData = d3.parseCsv(evictionsData)
-    }) ...
+    })
 } else {
   // if we don't have the cached file, let's go ahead
   // and download the bigger one, manipulate it with some
@@ -145,7 +145,7 @@ if (iodide.file.exists(FILENAME)) {
     .then(processDataAndReduceItsSize)
     .then((finalData) => {
       evictionsData = finalData
-      iodide.file.save(evictionsData, 'dataset.csv', d3.csvFormat)
+      return iodide.file.save(evictionsData, 'dataset.csv', d3.csvFormat)
     })
 }
 ```
--- a/docs/api.md
+++ b/docs/api.md
@@ -10,22 +10,28 @@ tracker](https://github.com/iodide-project/iodide/issues/new).
 
 ## `iodide.file`
 
-The `iodide.file` API provides convenience functions around working with files uploaded to the Iodide server in your notebook.
+The `iodide.file` API provides convenience functions around working with files
+uploaded to the Iodide server in your notebook.
 
 
 
 ### `iodide.file.save(data, fileName[, saveOptions])`
 
-Returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), which, when resolved, will signal that `data` was uploaded to the server under the filename `fileName`.
+Returns a
+[Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises),
+which, when resolved, will signal that `data` was uploaded to the server under
+the filename `fileName`.
 
-`data` (required) is any object or variable in the eval name space. It can be anything, really - an array of data, or a csv, or a binary blob.
+`data` (required) is any object or variable in the eval name space. It can be
+anything, really - an array of data, or a csv, or a binary blob.
 
-`filename` (required) is a string that represents the file name.
+`fileName` (required) is a string that represents the file name.
 
 The optional argument `saveOptions` has the following keys:
 
-- `overwrite` (optional, default `false`): if `true`, will overwrite whatever is at `fileName` with `data`. If `false`, will throw an error and halt the subsequent enqueued evaluation (similar to a syntax error).
-- `serializer` (optional): a function that transforms `data` to its final form.
+- `overwrite` (optional, default `false`): if `true`, will overwrite whatever is
+  at `fileName` with `data`. If `false`, will throw an error and halt the
+  subsequent enqueued evaluation (similar to a syntax error).
 
 
 ##### `iodide.file.save` patterns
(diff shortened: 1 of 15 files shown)


docs/api.md
```diff
@@ -6,6 +6,171 @@ challenges presented by JavaScript and web browsers.
...
+  // and download the bigger one, manipulate it with some
+  // function processDataAndReduceItsSize, then save it to the server
+  // so the next time we run this code chunk, we'll just load the 
+  // cached (smaller) version.
+  fetch('http://...')
+    .then((r) => r.json())
+    .then(processDataAndReduceItsSize)
+    .then((finalData) => {
+      evictionsData = finalData
+      return iodide.file.save(evictionsData, 'dataset.csv', d3.csvFormat)
```
hamilton: hmm, I need to think more about this example, because if I don't own this notebook, I'm not going to be saving the data.

bcolloran:
@hamilton i think this nails it. I would not change a single thing about this.

i've changed my mind since we talked this afternoon -- i think you're right that it's probably better if `file.save` just warns if the person evaling does not own a copy, b/c in the case when a notebook author leaves in a `save` call, we don't want it to crash their report view.

Also: i really love this approach of writing the docs _first_! What a phenomenal idea! We should consider doing this more often! :+1:

docs/api.md
```diff
@@ -6,6 +6,171 @@ challenges presented by JavaScript and web browsers.
...
+Returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), which, when resolved, will signal that `data` was uploaded to the server under the filename `fileName`.
+
+`data` (required) is any object or variable in the eval name space. It can be anything, really - an array of data, or a csv, or a binary blob.
+
+`filename` (required) is a string that represents the file name.
+
+The optional argument `saveOptions` has the following keys:
+
+- `overwrite` (optional, default `false`): if `true`, will overwrite whatever is at `fileName` with `data`. If `false`, will throw an error and halt the subsequent enqueued evaluation (similar to a syntax error).
+- `serializer` (optional): a function that transforms `data` to its final form.
```
hamilton: This may not be necessary, vs. just serializing `data` before passing into `iodide.file.save`

docs/api.md
```diff
@@ -6,6 +6,171 @@ challenges presented by JavaScript and web browsers.
...
+
+// iodide.file.save works with ArrayBuffers as well.
+iodide.file.save(new ArrayBuffer(...), 'model-output.bin')
+```
+
+
+
+### `iodide.file.load(fileName, fileType[, variableName])`
+
+Returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises) that, when resolved, loads into the notebook the save file `fileName`. For each notebook can find a list of uploaded files in the [upload modal](#link-required) or on your notebook's [Revisions Page](#needs-link). You can also access any uploaded file through the fetch chunk, [following this pattern](https://iodide-project.github.io/docs/workflows/#uploading-data-to-an-iodide-notebook). For most use cases using [fetch chunks](https://iodide-project.github.io/docs/jsmd/#fetch-chunks-fetch) is preferable and more straightforward. In more dynamic programmatic cases, however, `iodide.file.load` can provide more nuanced workflows.
```
openjck: This is definitely a nitpick, but is there any way this could be wrapped to 80 characters? That might make future diffs easier to read. I have <kbd>F5</kbd> set to `gqip` in my *.vimrc* for this kind of thing. :smiley:

docs/api.md
```diff
@@ -6,6 +6,171 @@ challenges presented by JavaScript and web browsers.
...
+
+
+
+### `iodide.file.save(data, fileName[, saveOptions])`
+
+Returns a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises), which, when resolved, will signal that `data` was uploaded to the server under the filename `fileName`.
+
+`data` (required) is any object or variable in the eval name space. It can be anything, really - an array of data, or a csv, or a binary blob.
+
+`filename` (required) is a string that represents the file name.
```
openjck: This argument is written in CamelCase (i.e., `fileName`) above.

docs/api.md
```diff
@@ -6,6 +6,171 @@ challenges presented by JavaScript and web browsers.
...
+
+`fileName` is the name of the file uploaded to the Iodide server. 
+
+`fileType` is the file type to handle. These are the same as the file types in [fetch chunks](https://iodide-project.github.io/docs/jsmd/#fetch-chunks-fetch): `js` (load this file as javascript, eg a library), `css` (load this file as a stylesheet, applying the styles to the report), `json` (load this file as json and parse into a javascript object), `text` (load this file as text), `blob` (load this file as binary).
```
wlach: I wonder if we should make this optional and have some sort of sensible default behaviour where not specified: maybe load the file as json if it's json, load the file as a blob otherwise?

docs/api.md
wlach: Maybe we could not specify it for now, and add the argument later if it turns out be helpful

wlach:
One note: I'd like to distinguish files uploaded via notebook content vs. user-uploaded files, somehow. Can probably just add an additional column to the files table/model.

docs/api.md
hamilton: Not a nitpick at all!

docs/api.md
hamilton: Good idea.

docs/api.md
hamilton: Yeah, I'm of two minds about this. We make it so you have to explicitly say what you want in the fetch chunk, and I think in general that's a good policy because it reduces magic a bit and lets a user know exactly what's going on at a pretty minimal cost. My hope would be to keep this API as similar to fetch chunks as possible. Blobs are not particularly gentle for people just trying to load a CSV, I think, so it may be better to just default to `text` if we go this direction. 

It could be that we need `iodide.file.load` to have other conveniences like `iodide.file.loadText` and `iodide.file.loadJSON` or whatever, which just wraps `iodide.file.load`, and leave this function as-is.

For now, I'll keep this as-is, and we can discuss this further during our Thursday meeting.

docs/api.md
```diff
@@ -6,6 +6,202 @@ challenges presented by JavaScript and web browsers.
...
+// iodide.file.save works with ArrayBuffers as well.
+iodide.file.save(new ArrayBuffer(...), 'model-output.bin')
+```
+
+
+
+### `iodide.file.load(fileName, fileType[, variableName])`
+
+Returns a
+[Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises)
```
mdboom: Does the promise return the content as a parameter?

docs/api.md
```diff
@@ -6,6 +6,202 @@ challenges presented by JavaScript and web browsers.
...
+// delete the various pngs originally saved to this notebook
+iodide.file.list()
+  .filter(filename => !filename.includes('.png'))
+  .forEach(iodide.file.delete)
+```
+
+
+### `iodide.file.exists(fileName)`
+
+Returns `true` if the file `fileName` is available to the notebook, and `false`
```
mdboom: Would this need to be async, and thus return a `Promise`?

docs/api.md
```diff
@@ -6,6 +6,202 @@ challenges presented by JavaScript and web browsers.
...
+    .then((finalData) => {
+      evictionsData = finalData
+      return iodide.file.save(evictionsData, 'dataset.csv', d3.csvFormat)
+    })
+}
+```
+
+
+
+### `iodide.file.lastUpdated(fileName)`
```
mdboom: Also async, I think?

docs/api.md
hamilton: It would not need to be async because we will (soon) have all the file metadata brought into the notebook from the server upon load.

docs/api.md
hamilton: same point as above

docs/api.md
mdboom: Oh, that's very cool, actually.

docs/api.md
hamilton: Good question – this thought crossed my mind. My thinking is (1) it will add it to `window` as `variableName` (as the fetch chunk) but I think it should also return the file for use in other languages. Otherwise there would have to be an additional import step if you were to use this API in python.

I should clarify this particular point and make sure to add it to the docs + the patterns section for `iodide.file.load`.

@wlach @bcolloran any thoughts on this in particular?

hamilton:
Thanks for the feedback y'all. At this point I'm going to implement these functions. Once this PR is ready for a first round of reviews, I'll move it out of draft mode. As of this comment I've pushed a number of changes / implemented parts of the API but expecting to refactor some parts of the implementation as I get more comfortable with the idea. In the meantime, you're free to peruse this code, but just know that it's likely subject to change.

hamilton:
I'm going to close this particular PR and start fresh one with the same exact API but different approach than these subsequent commits. This branch has gotten just far enough away from `master` at this point that it would be inefficient to build off of what I have here.

PR #1663, opened 2019-03-21
brings file metadata into the notebook client

hamilton:
This brings in associated files into the notebook, which seeks to give a common basis to work on #1651 and the in-notebook file uploader (which has not been filed as of this PR).

I noticed that `make lintfix` fixed a few files that were in our repository that I did not touch. Apologies - it seems like those probably should have gotten linted before, I'm guessing?

@openjck plz let me know if this has everything you need for your work.

### DIFF
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -33,6 +33,13 @@ def notebook_view(request, pk):
         revision = get_object_or_404(NotebookRevision, pk=int(request.GET["revision"]))
     else:
         revision = notebook.revisions.first()
+    files = [
+        {
+            'filename': file.filename,
+            'id': file.id,
+            'lastUpdated': file.last_updated.isoformat()
+        }
+        for file in File.objects.filter(notebook_id=pk).order_by("-last_updated")]
     notebook_info = {
         "username": notebook.owner.username,
         "user_can_save": notebook.owner_id == request.user.id,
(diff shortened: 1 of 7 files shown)


openjck:
I don't have merge access but I'd say this is merge-able.

server/notebooks/views.py
```diff
@@ -33,13 +33,18 @@ def notebook_view(request, pk):
         revision = get_object_or_404(NotebookRevision, pk=int(request.GET["revision"]))
     else:
         revision = notebook.revisions.first()
+    files = [
+        {"filename": file.filename, "id": file.id, "lastUpdated": file.last_updated.isoformat()}
+        for file in File.objects.filter(notebook_id=pk).order_by("-last_updated")
+    ]
```
bcolloran: man i miss list comprehensions...

bcolloran:
merging, thanks @hamilton and @openjck

server/notebooks/views.py
wlach: Ideally the file model would be ordered by last_updated, which would make the last clause unnecessary.

We already do this for e.g. NotebookRevision:

https://github.com/iodide-project/iodide/blob/f7699e90da6ad52e4f36c44d765afdae70c7050c/server/notebooks/models.py#L59

(this can be done in a followup, if you prefer)

PR #1668, opened 2019-03-23
Temporary fork markdown-it katex plugin

jaylinski:
## Description
The markdown-it plugin for katex hasn't been updated since 2016, so
the katex version was out of date (https://github.com/waylonflinn/markdown-it-katex/blob/v2.0.0/package.json). By temporarily forking the plugin, we can use the latest katex package, which includes lots of fixes.

This commit should be reversed as soon as the markdown-it katex plugin
is **actively maintained** again.

### Pull Request checklist

### Tests

No tests have been added, because this commit is only a package update.


package.json
```diff
@@ -51,12 +51,12 @@
     "golden-layout": "1.5.9",
     "iodide-react-inspector": "2.3.2",
     "jss": "9.8.7",
+    "katex": "^0.10.1",
```
jaylinski: I didn't commit the `package-lock.json`, because npm fucks everything up in the lock-file as soon as I remove or install a package. (I guess the npm-version has to match...)

src/eval-frame/markdown-it-katex.js
```diff
@@ -0,0 +1,226 @@
+/*
+Forked from https://github.com/waylonflinn/markdown-it-katex, because of inactivity.
+This file can be replaced by `markdown-it-katex` as soon as it is using the latest katex package.
+*/
+
+/* eslint-disable no-param-reassign, no-cond-assign */
```
jaylinski: I was too lazy to fix all eslint issues...

wlach:
Hi @jaylinski -- I'm not sure if vendoring the package is the right solution here. Maybe we could use this user fork of the package instead? 

https://www.npmjs.com/package/@iktakahiro/markdown-it-katex

I don't know if @iktakahiro wants to take on the responsibility, but they might be worth considering taking over the package if the original maintainer no longer wants to maintain it. The procedure for this is pretty straightforward and I've had good experiences with it in the past:

https://wrla.ch/blog/2017/07/taking-over-an-npm-package-sanity-prevails/

jaylinski:
@wlach My PR is not the right solution at all, that's why it should only be temporary.

This PR fixes the issue right now, waiting for someone else could take weeks/months/years.
I'm fine if you want to wait for someone to update the package or provide an alternative.

Iodide is awesome by the way! :heart: :smiley:

wlach:
@jaylinski right, but @itakahiro's fork of the package seems to do everything this vendored version does, without the burden of us maintaining it. Try this:

```bash
npm uninstall markdown-it-katex
npm install @iktakahiro/markdown-it-katex
```

Then change:

`import MarkdownItKatex from "markdown-it-katex";`

to

`import MarkdownItKatex from "@iktakahiro/markdown-it-katex";`

and verify that the results are what you'd expect when editing an iodide notebook. If all looks good, feel free to submit another PR :)

iktakahiro:
Hi, Thank you for using my package @iktakahiro/markdown-it-katex , I will continue maintenance if necessary. 👍

PR #1676, opened 2019-03-26
Add improved file upload UI

openjck:
### Pull Request checklist

### DIFF
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -43,11 +43,11 @@ syntax --
 TYPE: VAR_NAME = files/FILE_NAME
 ```
 
--- where `TYPE` is the type of the file ("text", "json", or "blob"), `VARNAME`
+-- where `TYPE` is the type of the file ("text", "json", "arraybuffer", or "blob"), `VARNAME`
 is the variable into which you want to load the data, and `FILE_NAME` is the
 name of the file you uploaded.
 
-This notebook [shows how to load and use an image blob](https://alpha.iodide.io/notebooks/127/)
+This notebook [shows how to load and use an image blob](https://alpha.iodide.io/notebooks/127/).
 
 ## Manipulating the DOM (for e.g adding plots)
 
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -10,7 +10,7 @@ It's possible to download data from a remote server using a browser APIs or a fe
 
 #### CORS problems
 
-One of the most common problems people encounter when attempting to load data into Iodide from a remote server are issues with [CORS settings](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which can restrict the ability of Iodide to download data from some other websites. Unfortunately, for security reasons browsers make it [impossible to surface CORS errors in a web app](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Functional_overview), so Iodide cannot give you the guidance you need directly, but if you have trouble loading a data file from a third party server, this is very likely to be the reason why. To check, you can open your browser's developer tools (press `ctrl+shift+i`); in case of a CORS issue, you will see a message like: 
+One of the most common problems people encounter when attempting to load data into Iodide from a remote server are issues with [CORS settings](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which can restrict the ability of Iodide to download data from some other websites. Unfortunately, for security reasons browsers make it [impossible to surface CORS errors in a web app](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Functional_overview), so Iodide cannot give you the guidance you need directly, but if you have trouble loading a data file from a third party server, this is very likely to be the reason why. To check, you can open your browser's developer tools (press `ctrl+shift+i`); in case of a CORS issue, you will see a message like:
 
 ```
 Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at {URI of resource you tried to load}.
@@ -21,20 +21,31 @@ To work around CORS errors, you need to upload your data to a server that does n
 
 ### Uploading data to an Iodide notebook
 
-_Note (2019-03-11): we are aware that this workflow is currently quite awkward. It is under active development and will be substantially improved very soon._
-
 _Note (2019-03-11): during our alpha period, files larger than 10MB cannot be uploaded._
 
-If you want to work with data stored on your local machine, or if you are encountering CORS problems while attempting to use a dataset from a remote serve, you may upload data directly to Iodide.
+If you want to work with data stored on your local machine or if you are
+encountering CORS problems while attempting to use a dataset from a remote
+server, you may upload data directly to Iodide.
 
-Data is uploaded on a per-notebook basis -- rather than uploading data to your user account, you upload it to a specific notebook. To do so, go to the "revisions" page of your notebook, which is available at `iodide.io/{path to your notebook}/revisions`. On this page, click the gear icon and select "Upload a File". Once you've uploaded the file, it will be listed under the "Files" heading on this page.
+Data is uploaded on a per-notebook basis&mdash;rather than uploading data to
+your user account, you upload it to a specific notebook. To do so, click on the
+hamburger menu icon near the top-left of the page, then click _Manage Files_.
+The file management modal will appear. Click _Add Files_ and select any number
+of files to upload. Once the files finished uploading, they will appear in a
+list alongside other files associated with that notebook. Click _Delete_ at any
+time to delete a file.
 
-To access this file from your notebook, you would use a fetch cell with the syntax --
+To access this file from your notebook, you would use a fetch cell with the
+syntax --
+
 ```
 %% fetch
 TYPE: VAR_NAME = files/FILE_NAME
 ```
--- where `TYPE` is the type of the file ("text", "json", or "blob"), `VARNAME` is the variable into which you want to load the data, and `FILE_NAME` is the name of the file you uploaded.
+
+-- where `TYPE` is the type of the file ("text", "json", or "blob"), `VARNAME`
+is the variable into which you want to load the data, and `FILE_NAME` is the
+name of the file you uploaded.
 
 This notebook [shows how to load and use an image blob](https://alpha.iodide.io/notebooks/127/)
 
@@ -54,4 +65,3 @@ d3.select("#plot-target") ... // d3
 ```
 
 You can always manipulate the DOM in Markdown cells as if your notebook were a static web page. Anything defined in a Markdown cell is fair game for DOM manipulation using the browser's [document API methods](https://developer.mozilla.org/en-US/docs/Web/API/Document).
-
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -9,7 +9,7 @@ from django.views.decorators.csrf import ensure_csrf_cookie
 
 from ..base.models import User
 from ..files.models import File
-from ..settings import APP_VERSION_STRING, EVAL_FRAME_ORIGIN
+from ..settings import APP_VERSION_STRING, EVAL_FRAME_ORIGIN, MAX_FILE_SIZE, MAX_FILENAME_LENGTH
 from ..views import get_user_info_dict
 from .models import Notebook, NotebookRevision
 from .names import get_random_compound
@@ -54,6 +54,8 @@ def notebook_view(request, pk):
         "connectionMode": "SERVER",
         "title": revision.title,
         "files": files,
+        "max_filename_length": MAX_FILENAME_LENGTH,
+        "max_file_size": MAX_FILE_SIZE,
     }
     if notebook.forked_from is not None:
         notebook_info["forked_from"] = notebook.forked_from.id
--- a/src/editor/components/menu/editor-toolbar-menu.jsx
+++ b/src/editor/components/menu/editor-toolbar-menu.jsx
@@ -9,7 +9,8 @@ import { connectionModeIsServer } from "../../tools/server-tools";
 
 export class EditorToolbarMenuUnconnected extends React.Component {
   static propTypes = {
-    isServer: PropTypes.bool.isRequired
+    isServer: PropTypes.bool.isRequired,
+    canUpload: PropTypes.bool
   };
 
   render() {
@@ -18,6 +19,9 @@ export class EditorToolbarMenuUnconnected extends React.Component {
         {this.props.isServer && <NotebookMenuItem task={tasks.newNotebook} />}
         {this.props.isServer && (
           <NotebookMenuItem task={tasks.toggleHistoryModal} />
+        )}
+        {this.props.canUpload && (
+          <NotebookMenuItem task={tasks.toggleFileModal} />
         )}
         <NotebookMenuItem task={tasks.clearVariables} />
         <NotebookMenuItem task={tasks.toggleHelpModal} />
@@ -28,8 +32,10 @@ export class EditorToolbarMenuUnconnected extends React.Component {
 
 export function mapStateToProps(state) {
   const isServer = connectionModeIsServer(state);
+
   return {
-    isServer
+    isServer,
+    canUpload: isServer && Boolean(state.userData.name)
   };
 }
 
--- a/src/editor/components/menu/editor-toolbar-menu.jsx
+++ b/src/editor/components/menu/editor-toolbar-menu.jsx
@@ -35,7 +35,7 @@ export function mapStateToProps(state) {
 
   return {
     isServer,
-    canUpload: isServer && Boolean(state.userData.name)
+    canUpload: isServer && Boolean(state.notebookInfo.user_can_save)
   };
 }
 
--- a/src/editor/components/modals/file-modal/add-button.jsx
+++ b/src/editor/components/modals/file-modal/add-button.jsx
@@ -1,0 +1,45 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "react-emotion";
+
+import { ContainedButton } from "../../../../shared/components/buttons";
+
+const AddButton = styled(ContainedButton)`
+  justify-self: start;
+  margin: 0;
+`;
+
+const HiddenInput = styled.input`
+  display: none;
+`;
+
+export default class extends React.Component {
+  static propTypes = {
+    onFileSelection: PropTypes.func.isRequired
+  };
+
+  constructor(props) {
+    super(props);
+    this.fileInputRef = React.createRef();
+  }
+
+  /**
+   * Click the hidden file <input>, which will cause the browser's file
+   * selection dialog to appear.
+   */
+  openSelectDialog = () => {
+    this.fileInputRef.current.click();
+  };
+
+  render = () => (
+    <React.Fragment>
+      <AddButton onClick={this.openSelectDialog}>Add files</AddButton>
+      <HiddenInput
+        type="file"
+        onChange={this.props.onFileSelection}
+        innerRef={this.fileInputRef}
+        multiple
+      />
+    </React.Fragment>
+  );
+}
--- a/src/editor/components/modals/file-modal/body.jsx
+++ b/src/editor/components/modals/file-modal/body.jsx
@@ -1,0 +1,65 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "react-emotion";
+
+import AddButton from "./add-button";
+import NoFilesNotice from "./no-files-notice";
+import FileList from "./file-list";
+import Footer from "./footer";
+import { fileShape } from "./propShapes";
+import { defaultSpacing } from "./styles";
+
+const bodyPadding = defaultSpacing;
+
+// Padding is not scrollable in Firefox. Wrapping it fixes that.
+// https://stackoverflow.com/a/29987556/4297741
+const BodyWrapper = styled.div`
+  display: grid;
+  height: 100%;
+  overflow: auto;
+`;
+
+const Body = styled.div`
+  align-items: start;
+  display: grid;
+  grid-gap: ${defaultSpacing};
+  grid-template-rows: min-content auto;
+  padding: ${bodyPadding};
+`;
+
+export default class extends React.Component {
+  static propTypes = {
+    files: PropTypes.objectOf(PropTypes.shape(fileShape)).isRequired,
+    onFileSelection: PropTypes.func.isRequired,
+    confirmDelete: PropTypes.func.isRequired
+  };
+
+  hasVisibleFiles = () => {
+    return (
+      Object.values(this.props.files).filter(file => {
+        return file.status !== "deleted";
+      }).length > 0
+    );
+  };
+
+  render = () => (
+    <React.Fragment>
+      <BodyWrapper>
+        <Body>
+          <AddButton onFileSelection={this.props.onFileSelection} />
+          {this.hasVisibleFiles() ? (
+            <React.Fragment>
+              <FileList
+                files={this.props.files}
+                confirmDelete={this.props.confirmDelete}
+              />
+              <Footer />
+            </React.Fragment>
+          ) : (
+            <NoFilesNotice />
+          )}
+        </Body>
+      </BodyWrapper>
+    </React.Fragment>
+  );
+}
--- a/src/editor/components/modals/file-modal/file-list-item.jsx
+++ b/src/editor/components/modals/file-modal/file-list-item.jsx
@@ -1,0 +1,221 @@
+import React from "react";
+import PropTypes from "prop-types";
+import CircularProgress from "@material-ui/core/CircularProgress";
+import styled from "react-emotion";
+
+import { TextButton } from "../../../../shared/components/buttons";
+import {
+  ListItem,
+  LIST_BORDER,
+  LIST_BORDER_RADIUS
+} from "../../../../server/components/list";
+import THEME from "../../../../shared/theme";
+import { fileShape } from "./propShapes";
+import { defaultSpacing } from "./styles";
+
+const circularProgressSize = "20px";
+
+const FileListItem = styled(ListItem)`
+  align-content: center;
+  align-items: center;
+  display: grid;
+  grid-auto-flow: column;
+  grid-gap: ${defaultSpacing};
+  justify-content: space-between;
+  min-height: ${circularProgressSize};
+
+  &.error {
+    background-color: #f4cccc;
+  }
+
+  &.deleted {
+    border: 0;
+    height: 0;
+    min-height: 0;
+    opacity: 0;
+    padding-bottom: 0;
+    padding-top: 0;
+    transform: scaleY(0);
+
+    transition: border 500ms, height 500ms, min-height 500ms, opacity 500ms,
+      padding-bottom 500ms, padding-top 500ms, transform 500ms;
+  }
+
+  /* Give the first and last visible list items rounded borders even if they are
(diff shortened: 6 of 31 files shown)


openjck:
@hamilton Updated with current progress. :smiley:

I still need to show some feedback, like a progress bar or a spinner, while a file is being uploaded. I also need to warn if the file size is above the specified limit.

src/shared/components/modal.jsx
```diff
@@ -22,7 +22,6 @@ const Backdrop = styled("div")`
   bottom: 0;
   left: 0;
   right: 0;
-  z-index: 1300;
```
hamilton: Couple things here:

(1) I would keep the `z-index` declaration in the styled component itself. You could easily add it like this:

```
const Backdrop = styled('div')`
...
z-index: ${props => props.zIndex || 1300};
...
`
```
(2) it might be a better and more tractable thing to not force me to understand the inner workings of the z-index in our modals by just having a keyword that says it needs to be on top of the other z-indices. Perhaps there is just a `nested` (true/false) which make the `z-index` 1400, otherwise the `z-index` is 1300. I think we probably won't ever have more than the 3 levels of z-index we have here so I think this should work nicely in theory.

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,230 @@
...
+    fileReader.readAsArrayBuffer(file);
+
+    fileReader.onload = () => {
+      saveFileToServer(this.notebookId, fileReader.result, file.name)
+        .then(savedFile => {
+          this.setState(state => {
+            state.files.push(savedFile);
+            return state;
+          });
+          this.props.updateAppMessages({
```
hamilton: these don't 100% work properly, I think. While it definitely shows a message in the snackbar, it does not show anything in the console. I recognize this is a failing we have with moving from dangerously set content for app messages in the console -> boilerplate messages, but we'll have to figure out something here before this merges.

Otherwise they all appear blank in the console:

Screen Shot 2019-04-23 at 3 11 28 PM

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,230 @@
...
+    return (
+      <React.Fragment>
+        <DeleteModal
+          visible={this.state.deleteModalVisible}
+          title={`Delete file "${this.state.deletedFileName}?"`}
+          onClose={this.hideDeleteModal}
+          onDelete={this.deleteFile}
+          elementID={this.state.deletedFileID}
+        />
+        <ModalContainer style={{ width: "90%" }} tabIndex="-1">
```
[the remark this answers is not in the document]
openjck: Good call. This was based on the help modal for visual consistency, but `90%` is too much.
2.

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,230 @@
...
+    };
+  };
+
+  addFile = () => {
+    this.fileInputRef.current.click();
+  };
+
+  copyFilePath = fileID => {
+    const pathElement = document.querySelector(`#file-${fileID} .file-path`);
+    pathElement.select();
```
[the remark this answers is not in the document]
openjck: I'm not sure if I understand. Can the second form (http://...) not be used?

src/shared/components/modal.jsx
openjck: I like this. Good suggestion.

openjck: Also, keeping the `z-index` declaration in the style component is going to require moving `Backdrop` into the constructor where `props` is in scope. Just FYI if you see some big whitespace changes.

src/editor/components/modals/file-modal.jsx
openjck: Huh. I didn't notice that! Are messages supposed to be simul-printed to the console?

openjck: Is there an example of an `updateAppMessages` message that is printed to the console correctly?

openjck:
This is pretty much done. I'll give it one more pass tomorrow before asking for a final review.

openjck:
This is ready for final review. :smiley:

src/server/components/notebook-actions-menu.jsx
```diff
@@ -57,7 +57,7 @@ export default class NotebookActionsMenu extends React.Component {
       let filename;
```
bcolloran: lol @openjck you should not have drawn attention to this method, now i'm going to ask you to refactor it ;-)

- we're trying to move toward async/await rather than `.then` -- can you convert this to an async method and await `selectFileAndFormatMetadata`?
- these funny `catch (err) { throw err; }` blocks keep showing up in our code. please be proactive in destroying them wherever you find the :-)
- but in this case, it looks like we might refactor this function using `await` an differently structured try/catch to both simplify this code AND to get rid of the `// eslint-disable-line prefer-destructuring` flag. how about somthing like:

```
try{
// try to do whatever we'd want id the file does exist...
const formdata = await selectFileAndFormatMetadata(...)
...etc code...
} catch {
// do whatever we need to do if the file does not exist
...etc code...
}
```

it looks like that is the actual logical structure here, and that we can get rid of the if/else below also

(cc @hamilton who knows this part of the code better than i do)

src/shared/components/modal.jsx
bcolloran: oh, plz add anything relating to z-indexes as constants in "src/editor/style/z-index-styles". we do some funky things with the layout manager, so i want to make sure that we can have an overview of the z-index twiddles in one place.

(@hamilton if you want to merge these constants into `THEME` that is fine with me, but all the Z stuff should be in one place; either this file or THEME.layers or something)

src/server/components/delete-modal.jsx
```diff
@@ -19,8 +19,10 @@ export default class DeleteModal extends React.Component {
       PropTypes.element,
       PropTypes.string,
       PropTypes.array
-    ])
+    ]),
+    aboveOtherModals: PropTypes.bool
```
bcolloran: should `aboveOtherModals`  be marked as `.isRequired`? @hamilton i suspsect a bunch of these proptypes should also be required. thoughts?

I don't know the details of DeleteModal, but it really looks like it needs to be refactored - every place it is invoked we hand in a ton of function props that deal with setting. it looks like we hand in a ton of redundant props each time is invoked.  i'm pretty sure that `onClose===onCancel` in each instance.

@openjck + @hamilton can you take a quick look at this and see if we can just get rid of onClose altogether?

src/server/components/notebook-actions-menu.jsx
openjck: Sure! How do you feel about doing this in a follow-up? This PR has gotten long in the tooth so I'd prefer not to block on this.

src/server/components/delete-modal.jsx
openjck: > should `aboveOtherModals` be marked as `.isRequired`? @hamilton i suspsect a bunch of these proptypes should also be required. thoughts?

I don't think so because *delete-modal.jsx* just passes the value to *modal.jsx*, where it is required.

src/shared/components/modal.jsx
```diff
@@ -129,7 +126,9 @@ export default class Modal extends React.Component {
       <Backdrop
         onClick={e => {
           e.stopPropagation();
-          this.props.onClose();
+          if (this.props.onClose) {
```
bcolloran: if this is a required prop, we should not have to guard against it being missing

src/shared/components/modal.jsx
```diff
@@ -147,5 +146,16 @@ export default class Modal extends React.Component {
 
 Modal.propTypes = {
   visible: PropTypes.bool.isRequired,
-  onClose: PropTypes.func
+  children: PropTypes.element,
```
bcolloran: as above, let's figure out whic of these props are actually required and which are not.

also, please move all these proptypes within the class definition

src/shared/components/modal.jsx
```diff
@@ -147,5 +146,16 @@ export default class Modal extends React.Component {
...
+  onClose: PropTypes.func,
+
+  // eslint thinks that this prop isn't being used, even though it's used in a
+  // react-emotion style.
+  //
+  // eslint-disable-next-line react/no-unused-prop-types
+  aboveOtherModals: PropTypes.bool
+};
+
+Modal.defaultProps = {
```
bcolloran: please move within the class definition

src/server/components/delete-modal.jsx
bcolloran: (or get rid of onCancel, or consolidate them as onCloseOrCancel -- for a modal of this sort, hitting cancel is always the same as closing the modal, right? the cancel button is just decoration for a no-op)

src/shared/components/modal.jsx
openjck: That's a good point. I'm not sure it should be required, though. I don't use it in *file-modal.jsx*. It may need to be removed entirely as you say elsewhere.

src/server/components/delete-modal.jsx
```diff
@@ -30,7 +32,9 @@ export default class DeleteModal extends React.Component {
     // add delete here
     try {
       await this.props.deleteFunction(this.props.elementID);
-      this.props.onClose();
+      if (this.props.onClose) {
+        this.props.onClose();
```
bcolloran: if this prop (or onCloseOrCancel) is always required, we should not have to make this defensive check

src/shared/components/modal.jsx
openjck: Sure! Is `static propTypes` preferred? If so, why? I noticed both are used throughout the codebase.

src/server/components/delete-modal.jsx
openjck: I just did a quick search through the codebase. We never send `DeleteModal` an `onClose` which is different from an `onCancel`. Unless Hamilton objects, I'll consolidate them into one.

src/shared/components/modal.jsx
bcolloran: static proptypes is preffered whenever we have a component class, just because it's tidier -- it's clear immediately what the proptypes are for the class, as they are right next to the constructor. for functional components, the `X.PropTypes` notation is required.

TBH, i have no idea how this `Modal.propTypes` got through review in the first place :-~

openjck:
> main thing: this PR def needs docs that write up how this file upload flow works and how one loads files. once the docs are in place, i'll do a proper QA pass in which i actually test out the flow, but i want to do that by following along with the docs. (so far i've only done code review)

Sure! I'll update the uploading section in the *workflows.md* doc.

openjck:
Updated the *workflows.md* docs.

docs/workflows.md
```diff
@@ -10,7 +10,7 @@ It's possible to download data from a remote server using a browser APIs or a fe
...
+One of the most common problems people encounter when attempting to load data into Iodide from a remote server are issues with [CORS settings](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), which can restrict the ability of Iodide to download data from some other websites. Unfortunately, for security reasons browsers make it [impossible to surface CORS errors in a web app](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Functional_overview), so Iodide cannot give you the guidance you need directly, but if you have trouble loading a data file from a third party server, this is very likely to be the reason why. To check, you can open your browser's developer tools (press `ctrl+shift+i`); in case of a CORS issue, you will see a message like:
```
openjck: ???

openjck: Oh, it removed a trailing space after *like:*. Weird that GitHub didn't highlight that.

src/editor/components/modals/file-modal.jsx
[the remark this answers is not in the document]
openjck: Sure thing. Just to confirm, you mean we should get rid of the *Copy path* button?

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+                      className={fileMeta.status}
+                    >
+                      {
+                        // This is something like a switch statement. Only one
+                        // will render.
+                        {
+                          uploading: (
+                            <React.Fragment>
+                              <FileName>{fileName}</FileName>
+                              <FileCircularProgress
```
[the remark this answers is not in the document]
openjck: Yeah, I think that's a good idea. Thanks!

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+            </AddFilesButton>
+            <HiddenInput // Hidden because this.openSelectDialog will trigger the click
+              type="file"
+              onChange={this.receiveFiles}
+              innerRef={this.fileInputRef}
+              multiple
+            />
+            {Object.keys(this.state.files).length > 0 && (
+              <FileList>
+                {Object.keys(this.state.files).map(key => {
```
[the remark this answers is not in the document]
openjck: Another good call. I agree!

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+                  file: savedFile
+                };
+                return newState;
+              });
+            })
+            .catch(() => {
+              this.setState(state => {
+                const newState = state;
+                newState.files[key].status = "error";
+                newState.files[key].errorMessage = "Upload error";
```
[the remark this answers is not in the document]
openjck: Agreed. I assumed the server just couldn't handle duplicate files but the replace option makes sense. When you say mirroring what the server does, do you mean what the revision page does?

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+            } characters`,
+            file: unsavedFile
+          };
+          return newState;
+        });
+      } else {
+        // Update state immediately to reflect that this file has begun
+        // uploading. This allows us to show a spinner in its row for the
+        // duration of the upload.
+        this.setState(state => {
```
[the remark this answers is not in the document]
openjck: For sure. You're right that objects like `state` are passed by reference. I used this approach because it satisfies the *airbnb* style guidelines, but you're right that it satisfies the guidelines in name only. I'll use `Object.assign`.

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+                                    className="file-action file-action-delete"
+                                    onClick={() => {
+                                      this.showDeleteConfirmationModal(
+                                        key,
+                                        fileMeta.file.id,
+                                        fileName
+                                      );
+                                    }}
+                                  >
+                                    delete
```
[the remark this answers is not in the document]
openjck: Agreed. I played around with fading the `FileItem` out, but it ended up being trickier than expected because CSS transitions cannot be applied to elements that are removed from the DOM. I'll take another look at this. My latest refactor, which gave each file a status, may make this easier.

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+            status: "uploading",
+            file: unsavedFile
+          };
+          return newState;
+        });
+
+        // Attempt to upload the file and update state accordingly
+        const fileReader = new FileReader();
+        fileReader.readAsArrayBuffer(unsavedFile);
+        fileReader.onload = () => {
```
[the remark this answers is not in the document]
openjck: I thought about that. Would that not make file uploads sequential? The file modal allows multiple files to be selected at one time. At the moment, all selected files are uploaded simultaneously. If we used `await`, I think they would be uploaded one-by-one.

I could be mistaken about that, though. I'll double-check.

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+          onCloseOrCancel={this.hideDeleteConfirmationModal}
+          onDelete={this.executePendingDelete}
+          deleteFunction={deleteFileOnServer}
+          elementID={this.state.pendingDelete.id}
+          aboveOtherModals
+        />
+        <ModalContainer tabIndex="-1">
+          <FileModalAppBar position="static">
+            <Toolbar>
+              <FileModalTitle variant="title">Available Files</FileModalTitle>
```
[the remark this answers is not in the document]
openjck: 👍

src/editor/components/modals/file-modal.jsx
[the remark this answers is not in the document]
openjck: Great tip, by the way. I didn't consider that the `Array.prototype.slice` cloning method doesn't copy inner objects by value. Interesting!

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+  };
+
+  /**
+   * Delete the file that the user confirmed they wanted to delete.
+   */
+  executePendingDelete = () => {
+    // setState enqueues state changes; they might not take effect immediately.
+    // To ensure that the delete modal disappears at the same time that the file
+    // disappears from the list, it's important to update "files" and
+    // "pendingDelete" in the same batch of setState updates.
```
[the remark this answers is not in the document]
openjck: > can't both of these changes be enqueued with two calls to `setState` within `executePendingDelete`? They both should get enqueued directly after each other and then update accordingly.

I'm trying to think... if a user begins uploading a file, then deletes a file, which would cause this method to execute, isn't it possible that the file will finish uploading (causing a `setState` call) between the two `setState` calls in this method?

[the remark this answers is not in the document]
openjck: Or are callbacks never executed in the middle of the execution of another method? I guess I've never thought about that.

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+    } else {
+      this.lastFileKey += 1;
+    }
+    return this.lastFileKey;
+  };
+
+  /**
+   * Upload all files that the user selected.
+   */
+  receiveFiles = e => {
```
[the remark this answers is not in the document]
openjck: Good call. This deserves some refactoring.

openjck:
Digging into this a bit deeper, I realize I'm also open to suggestions for simplifying `render`. It's pretty big right now. I can't decide if it would be better to flatten things out by breaking some things into sub-components (like *help-modal.jsx* does) or keep everything in one big, albeit contained, `render` function (like *revisions-page.jsx* does). Suggestions welcome. Feel free to point me to any relevant design patterns that you recommend that could make `render` or any other part of this file easier to work with. :smiley:

src/editor/components/modals/file-modal.jsx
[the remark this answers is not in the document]
openjck: Thanks for catching this. I feel a lot better after cleaning this up.

src/editor/components/modals/file-modal.jsx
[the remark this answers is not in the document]
openjck: I'm mistaken. `forEach` doesn't need to wait for the `await` to resolve before moving onto the next element. Thanks!

openjck:
I should also refactor `render`. It's pretty big right now. I could split the sub-components (`FileList`, `FileListItem`, etc.) into separate files, but I tend not to do that when they are unlikely to be reused. What do you think?

hamilton:
@openjck awesome. I don't have a strong opinion re: rendering, neither in this case nor philosophically. My hunch, however, is that we will be reusing the entire component on the server, if possible.

src/editor/components/modals/file-modal.jsx
```diff
@@ -0,0 +1,472 @@
...
+            file: unsavedFile
+          };
+          return newState;
+        });
+
+        // Attempt to upload the file and update state accordingly
+        const fileReader = new FileReader();
+        fileReader.readAsArrayBuffer(unsavedFile);
+        fileReader.onload = () => {
+          saveFileToServer(
```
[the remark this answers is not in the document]
openjck: Thanks for catching this. Would it be better to use `saveFile` from *file-request-actions.js*? It seems to do some checks that would be useful, like checking that the file doesn't already exist. I probably should have been using that from the start, huh?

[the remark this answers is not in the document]
openjck: I could move all of my local state into Redux, really. Not sure if that would be overkill.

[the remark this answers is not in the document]
openjck: I ended up using `saveFile`; it seems like the right way to go.

openjck:
Getting closer.

@hamilton, do you know why I'm getting `TypeError: fileRequests[fileRequestID] is undefined` when uploading or deleting a file? I don't think I'm doing anything wrong in my code, but I could be misunderstanding something. I'm going to keep digging, but thought you might have a clue.

openjck:
This should be done minus tests. 😃

src/editor/components/modals/file-modal/file-list-item.jsx
```diff
@@ -0,0 +1,219 @@
...
+  &.error {
+    background-color: #f4cccc;
+  }
+
+  &.deleted {
+    border: 0;
+    height: 0;
+    min-height: 0;
+    opacity: 0;
+    padding: 0;
```
hamilton:
```suggestion
    padding-top: 0;
    padding-bottom: 0;
    transform: scale(1, 0);
```

src/editor/components/modals/file-modal/file-list-item.jsx
```diff
@@ -0,0 +1,219 @@
...
+
+    transition: border 250ms step-end, height 250ms step-end,
```
hamilton:
```suggestion
    transition: border 250ms step-end, height 250ms,
```

src/editor/components/modals/file-modal/file-list-item.jsx
```diff
@@ -0,0 +1,219 @@
...
+      min-height 250ms step-end, padding 250ms step-end, opacity 250ms;
```
hamilton:
```suggestion
      min-height 250ms step-end, padding-top 250ms, padding-bottom 250ms, transform 250ms,
      opacity 250ms;
```

src/editor/components/modals/file-modal/file-list-item.jsx
```diff
@@ -0,0 +1,219 @@
...
+  }
```
hamilton: Hey @openjck – I love this approach you started with this kind of transition. I'm not 100% sold on the idea of keeping the elements (vs. unmounting), but I don't think in practice it matters much. That this isn't easy is to me a real pain when using React. At any rate, the above changes here actually also animate the collapsing of the element, so it's a bit easier for the eye to track what happens when you delete. Try these suggestions out and let me know what you think.

src/editor/components/modals/file-modal/title.jsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
```
hamilton: Is this `title.jsx` element not something that we can generalize? We have other modals that have the same style. Any chance we can have one component in all places? I'm sure at this point we're not using one title component for all the other cases, but we probably should!

src/editor/components/modals/file-modal/add-button.jsx
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  render = () => (
+    <React.Fragment>
+      <AddButton onClick={this.openSelectDialog}>Add files</AddButton>
+      <HiddenInput
+        type="file"
+        onChange={this.props.onFileSelection}
+        innerRef={this.fileInputRef}
+        multiple
```
hamilton: in theory I'm a big fan of `multiple` but we don't ask for duplicates on each selected file. I dunno. Maybe we keep this + add a `FIXME` about this. I don't really want to block on it since it's more useful to have the feature than to not have it + avoid the bug.

src/editor/components/modals/file-modal/body.jsx
```diff
@@ -0,0 +1,64 @@
...
+`;
+
+export default class extends React.Component {
+  static propTypes = {
+    files: PropTypes.objectOf(PropTypes.shape(fileShape)).isRequired,
+    onFileSelection: PropTypes.func.isRequired,
+    confirmDelete: PropTypes.func.isRequired
+  };
+
+  hasVisibleFiles = memoize(files => {
```
hamilton: is `memoize-one` necessary here? This function is probably very fast to perform, right? Even in hairy cases (hundreds of files) this function probably wouldn't slow things down if it were run a bunch of times.

openjck: True, fair enough. I'll remove memoize-one.

src/editor/components/modals/file-modal/file-list-item.jsx
openjck: I don't think it's possible to animate them without keeping them in the DOM. Transitions can't be applied to elements removed from the DOM or set to `display: none`. I'll test out your animations and let you know what I think!

src/editor/components/modals/file-modal/add-button.jsx
openjck: Good catch. I'll fix this now. Shouldn't be too hard.

src/editor/components/modals/file-modal/file-list-item.jsx
openjck: I like this. I made a couple of small tweaks but nothing huge.

openjck:
Updated to address review comments and add tests. This should be ready for final review. :smiley:

openjck:
Fixed the lint issue. Would you mind double-checking for me? I had trouble reproducing it at first.

openjck:
I also notice that tests are failing but it doesn't seem (?) like that has anything to do with my code.

openjck:
Updated. Would you mind double-checking the linting issues? I'm seeing `make lint` fail, but it's also failing on master. I've asked about that on Gitter.

server/notebooks/views.py
```diff
@@ -7,12 +7,17 @@
 from django.urls import reverse
 from django.views.decorators.csrf import ensure_csrf_cookie
 
+from .models import Notebook, NotebookRevision
+from .names import get_random_compound
```
wlach: If you re-run the latest version of isort against this file, it should fix the issue.

openjck: Great! Working on some Docker surgery right now, then I'll take care of that. :smiley:

openjck:
Updated to apply isort and appease black. I needed to update the isort config to fix a typo and apply some other options that are recommended in the black docs. Can I bother you to double-check that I got this right, @hamilton? :smiley:

openjck:
@hamilton Sorry to give you one more thing to review, but I just fixed a bug where users could open the file modal on notebooks that they did not own and *attempt* to upload files to it. The upload would be rejected by *file-request-actions.js*, but it would look like it had succeeded.

This has been fixed in the new commit. The file modal cannot be opened in an unowned notebook and if the user does somehow get it open and attempts to upload a file, they will see an error.

src/editor/components/menu/editor-toolbar-menu.jsx
```diff
@@ -28,8 +32,10 @@ export class EditorToolbarMenuUnconnected extends React.Component {
 
 export function mapStateToProps(state) {
   const isServer = connectionModeIsServer(state);
+
   return {
-    isServer
+    isServer,
+    canUpload: isServer && Boolean(state.notebookInfo.user_can_save)
```
hamilton: Good catch here. One additional problem, though – if I fork the notebook, we need to make sure to clear the `files` object here. This is a more general problem in forking not specific to this. So let's file a follow-up issue after this is merged and we can address this problem.

PR #1672, opened 2019-03-25
Switch to maintained markdown-it-katex plugin

jaylinski:
## Description

The markdown-it plugin for katex hasn't been updated since 2016, so
the katex version was out of date.

Superseeds #1668.

### Pull Request checklist

- [ ] **Documentation**: If this feature has or requires documentation, the relevant docs have been updated.
- [x] **Changelog**: This PR updates the [changelog](../CHANGELOG.md) with any user-visible changes.
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not

### Tests

No tests have been added, because this commit is only a package update.

PR #1677, opened 2019-03-26
Switch to maintained markdown-it-katex plugin

wlach:
This is just #1672 with some unwanted package-lock.json changes reverted


wlach:
Thanks for the patch!

PR #1685, opened 2019-03-28
Allow the microphone in the iframe

mdboom:
This change is required to make access to the microphone through the Web Audio API work.

See this example notebook (currently broken on `alpha.iodide.io` obviously):

https://alpha.iodide.io/notebooks/1440/

I am not a security expert, so feel free to reject this on security grounds.  However, my understanding is that the default that iframe's don't have access is based on the fact that iframes are often given over to third-parties (ad servers) to provide.  In this case, we control both the main and iframes, and we already allow all kinds of arbitrary code to run in the iframe.  Additionally, there is still an explicit permission pop-up, a flashing red microphone icon and a tray indicator.

All that said, the use case I have in mind here is just to create a cool demo ;)

### Pull Request checklist

### DIFF
--- a/server/notebooks/templates/notebook.html
+++ b/server/notebooks/templates/notebook.html
@@ -11,6 +11,7 @@
   sandbox="allow-scripts allow-same-origin allow-popups"
   allowfullscreen="true"
   allowvr="yes"
+  allow="microphone"
 ></iframe>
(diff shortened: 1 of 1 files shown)


bcolloran:
this looks fine to me if it works for you in your testing @mdboom :+1:

PR #1688, opened 2019-03-29
File reorg

bcolloran:
@hamilton and @wlach this PR is as far as i got with trying to reorganize the file hierarchy before my eyes glazed over completely. without actually refactoring code we can't do anything like the "redux ducks" pattern or some other more elaborate scheme, so i've just tried to move existing files around into somewhat more logical groupings, mostly breaking up folders that had become trash pits like the various "tools" folders.

there is still more to be done, but i will leave some of this to your discretion -- in particular @hamilton i didn't do much with `/src/server` or `/src/shared`, and they could use a look. both have big `components` subfolders that could do with some organization.

but the `editor` and `eval-frame` folders are in much better shape. and it feels much better to have these things at the top level in `/src` rather than kind of buried inside the editor code.

also: we have come a long way! i have not looked at our whole huge pile of files since the last time we organized everything, which must have been like a year ago by now. we have written a lot of code! :-)

unless you see show stopping problems with this, i think we should just try to merge it in while things are at a relative lull rather than having use this branch as the basis for reorging `/src/server` or `/src/shared` and waiting until that is done before merging. with this big move done, things should be relatively decoupled for a moment, and it should be easier to follow up with smaller additional bits of organization.

### Pull Request checklist


wlach:
looks like some nice cleanup to me! no problem on my end with merging this.

PR #1698, opened 2019-03-31
[MRG] fix internal link targeting

### LINKED ISSUE #1601: Reference links to dom elements broken
using MASTER

### What I Did
I'm trying to make a reference link to a dom element but it doesn't resolve to the location.
https://definitely-staging.iodide.io/notebooks/19/

@wlach suggested a solution to this bug using [this](https://stackoverflow.com/a/19378662).

teonbrooks:
attempts to close #1601 
@wlach, this was my initial attempt to fix the internal link referencing. it seems that the markdown-it-external-links isn't affecting the html tags.

### DIFF
--- a/src/editor/user-tasks/user-task.js
+++ b/src/editor/user-tasks/user-task.js
@@ -1,6 +1,6 @@
 const TASK_ERRORS = {
   noCallback:
-    "you must provide either a callback, a keybindingCalblack, or both (if the callbacks differ)",
+    "you must provide either a callback, a keybindingCallblack, or both (if the callbacks differ)",
   noKeybindingsWithCallback:
     "keybindingCallback supplied, but no keybindings were declared",
   argumentMustBeObject: "must provide an object as an argument",
--- a/src/editor/user-tasks/user-task.js
+++ b/src/editor/user-tasks/user-task.js
@@ -1,6 +1,6 @@
 const TASK_ERRORS = {
   noCallback:
-    "you must provide either a callback, a keybindingCallblack, or both (if the callbacks differ)",
+    "you must provide either a callback, a keybindingCallback, or both (if the callbacks differ)",
   noKeybindingsWithCallback:
     "keybindingCallback supplied, but no keybindings were declared",
   argumentMustBeObject: "must provide an object as an argument",
--- a/src/eval-frame/tools/markdown-it-tags.js
+++ b/src/eval-frame/tools/markdown-it-tags.js
@@ -15,18 +15,12 @@ function tagsPlugin(md, options) {
       if (token.children) {
         token.children.map(applyFilterToTokenHierarchy);
       }
-
-      const href = token.attrGet("href");
-      const tagref = isTagLink(href);
-
-      if (tagref) {
+      if (isTagLink(token.attrGet("href"))) {
         token.attrSet("target", tagTarget);
       }
     }
-
     state.tokens.map(applyFilterToTokenHierarchy);
   }
-
   md.core.ruler.push("tag_links", tagLinks);
 }
 
--- a/src/eval-frame/tools/markdown-it-tags.js
+++ b/src/eval-frame/tools/markdown-it-tags.js
@@ -1,0 +1,49 @@
+// Modified MarkdownItExternalLink
+// Copyright (c) Rotorz Limited. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root.
+
+function tagsPlugin(md, options) {
+  options = options || {};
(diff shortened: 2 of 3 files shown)


wlach:
hey @teonbrooks -- is this the behaviour that we want? I think we do actually want to open most types of internal links in a new tab -- it's just bookmark links that we want to treat specially.

teonbrooks:
@wlach yeah, I just got finished playing around with it and I realized that this would not be the way to go. 

I saw your notebook you shared and that fixed the problem I was having. in essence, we would need to mod the md, one way would be to write a plugin that look for link references and add `target="_self"`. what would be the best way to go about it?

wlach:
@teonbrooks I think markdown-it-external-links might be a good model for what we're doing here, just with a much simplified ruleset that says, basically:

"if the link begins with a hash, set target to self"

We could create such a plugin somewhere inside the iodide source tree, or maybe even just inside the report-pane.js file since it's likely just to be a few lines.

https://github.com/rotorz/markdown-it-external-links/blob/master/lib/index.js

Does that make sense?

src/eval-frame/tools/markdown-it-tags.js
```diff
@@ -0,0 +1,41 @@
+// Modified MarkdownItExternalLink
+// Copyright (c) Rotorz Limited. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root.
+
+function tagsPlugin(md, options) {
+  const tagOptions = options || {};
+  const tagTarget = tagOptions.tagTarget || "_self";
+
+  // helper function
+  const isTagLink = href => {
```
wlach: I think you should be able to simplify this to:

```js
const isTagLink = href => href && href.startsWith("#");
```

package.json
```diff
@@ -58,6 +58,7 @@
     "markdown-it": "8.4.2",
     "markdown-it-anchor": "5.0.2",
     "markdown-it-emoji": "1.4.0",
+    "markdown-it-external-links": "0.0.6",
```
wlach: You probably want to use git rebase to remove the commit that modifies this file and package-lock.

My normal method would be to do something like `git rebase -i HEAD^^^^^^^` and then removing the line in my editor. 

(sorry if you already know this!)

teonbrooks:
> Does it work for you?

it works for me, if you check the link in the report view, it should have the `target="_self"`.

here's a demo i've been using
```

%% md
[Teon's Website](https://teonbrooks.github.io)

%%
<span class="greendot"></span>

%%
<div id='test'>
	<span class="dot"></span>
</div>

%% css
.dot {
  height: 800px;
  width: 800px;
  background-color: #bbb;
  border-radius: 50%;
  display: inline-block;
}

.greendot {
  height: 800px;
  width: 800px;
  background-color: green;
  border-radius: 50%;
  display: inline-block;
}


%% md
[testing](#test)
```

package.json
teonbrooks: good eye, i forgot that i had imported it

package-lock.json
```diff
@@ -1996,7 +1996,7 @@
         },
```
wlach: Can you revert these changes as well?

src/eval-frame/tools/markdown-it-tags.js
```diff
@@ -0,0 +1,33 @@
...
+      if (token.children) {
+        token.children.map(applyFilterToTokenHierarchy);
+      }
+
+      const href = token.attrGet("href");
+      const tagref = isTagLink(href);
+
+      if (tagref) {
+        token.attrSet("target", tagTarget);
+      }
```
wlach: Let's simplify this a bit more, now that we've debugged it. Lines 19 -> 24 could be reduced to:

```js
if (isTagLink(token.attrGet("href"))) {
  token.attrSet("target", tagTarget);
}
```

src/editor/user-tasks/user-task.js
```diff
@@ -1,6 +1,6 @@
 const TASK_ERRORS = {
   noCallback:
-    "you must provide either a callback, a keybindingCalblack, or both (if the callbacks differ)",
+    "you must provide either a callback, a keybindingCallblack, or both (if the callbacks differ)",
```
hamilton: typo – `keybindingCallback`

teonbrooks: hah, my correction only did half the job

teonbrooks:
@wlach anything else you see to fix with this PR?

wlach:
> @wlach anything else you see to fix with this PR?

Nope, this looks great. Let's merge!

PR #1703, opened 2019-04-02
iodide.file implementation

hamilton:
This implements `iodide.file` as described in the API docs i wrote up and had everyone review before. 

[This notebook](https://definitely-staging.iodide.io/notebooks/26/) has a handful of examples. It might be a good idea to deploy this branch to `definitely-staging` to try it out. Otherwise you can copy and paste the jsmd into a notebook on your local copy.

I have a few questions:

- allowing `js` load types: is this a good idea? Seems like there could be some weird security implications. Aside from that, I can also make an infinitely recurring script that calls `iodide.load('script.js', 'js')`, which is a fun hat trick, but I don't want to expose functionality that doesn't actually advance the story.
- allowing `css` load types: this one seems like it could be better than the js one, but still unclear if its the best idea. My main concern is we don't have `iodide.file.unload` to remove the stylesheet if a user doesn't want it. I can see this all becoming a huge point of confusion, but I'll let others decide. It would be easy to have these two cases handled elsewhere eg in the case of style sheets, some special function that allows you to use a custom template stylesheet to save you from having to do the same styles over and over again in reports
- namespacing: I found that `iodide.file.save` and company to be a lot of characters to type. I've come around to just having `io.save`, etc. as our API semantics. No strong feelings, but it'd be nicer to have fewer characters instead of such deeply namespaced ones. Perhaps the solution is in our exmaples we always set `var iosave = iodide.file.save` etc.
- I have tests for most of this code, with a handful of small exceptions. The API, message-sending point from the eval-frame, & the editor actions that handle those messages are pretty well-tested using the patterns elsewhere in the code base. Let me know if anything is amiss here.

### Pull Request checklist


src/editor/port-to-eval-frame.js
```diff
@@ -27,6 +29,61 @@ const approvedKeys = [
   "ctrl+shift+right"
 ];
 
+function validateRequestType(requestType) {
+  if (!["LOAD_FILE", "SAVE_FILE", "DELETE_FILE"].includes(requestType)) {
+    throw Error(`file operation "${requestType}" not defined`);
+  }
+}
+
+function handleFileRequest(message) {
```
hamilton: this might end up going in its own file, so we can better test all the little bits of logic

src/eval-frame/tools/send-file-request-to-editor.js
```diff
@@ -0,0 +1,83 @@
+import generateRandomId from "../../shared/utils/generate-random-id";
+import messagePasserEval from "../../shared/utils/redux-to-port-message-passer";
+
+// NOTE: exporting this for tests.
+export const fileRequestQueue = {};
+
+export function onParentContextFileRequestSuccess(
```
hamilton: this function and the one below it are very close in spirit to `onParentContextFileFetchSuccess` etc., which was used to resolve the local file lines in fetch chunks.

src/eval-frame/tools/send-file-request-to-editor.js
```diff
@@ -0,0 +1,83 @@
...
+    );
+  if (typeof fileRequestID !== "string")
+    throw new TypeError(
+      "onParentContextFileRequestSuccess requires a string fileRequestID"
+    );
+  fileRequestQueue[path].requests[fileRequestID].resolve(responseOrFile);
+  delete fileRequestQueue[path].requests[fileRequestID];
+}
+
+export function onParentContextFileRequestError(reason, path, fileRequestID) {
```
hamilton: leaving as a reminder to add some error handling to this

src/shared/upload-file.js
```diff
@@ -62,3 +64,47 @@ export function selectAndUploadFile(notebookID, successCallback = () => {}) {
...
+    JSON.stringify({
+      filename: file.name,
+      notebook_id: notebookID
+    })
+  );
+  formData.append("file", file);
+  return formData;
+}
+
+export function saveFileToServer(
```
hamilton: This file needs to be refactored and generalized to just be something like `file-operations.js`.

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+uploaded to the Iodide server in your notebook.
+
+
+
+### `iodide.file.save(data, fileName[, saveOptions])`
+
+Returns a
+[Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises),
+which, when resolved, will signal that `data` was uploaded to the server under
+the filename `fileName`.
```
mdboom: If it errors, I assume it calls the `Promise's` `reject`?

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+
+`data` (required) is any object or variable in the eval name space. It can be
+anything, really - an array of data, or a csv, or a binary blob.
+
+`fileName` (required) is a string that represents the file name.
+
+The optional argument `saveOptions` has the following keys:
+
+- `overwrite` (optional, default `false`): if `true`, will overwrite whatever is
+  at `fileName` with `data`. If `false`, will throw an error and halt the
```
mdboom: Does it throw or use `Promise.reject`?

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+
+```javascript
+
+// upload a csv to the server. Assume the object csvData is an array of objects
+// representing a columnar dataset. This overwrites whatever is in `cached-data.csv`
+// because overwrite is set to true.
+
+const csvData = [{x1: 10, x2: 'test1'}, {x1: 20, x2: 'test2'}, ...]
+
+iodide.file.save(csvData, 'cached-data.csv' { overwrite: true })
```
mdboom:
```suggestion
iodide.file.save(csvData, 'cached-data.csv', { overwrite: true })
```

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+iodide.file.save(new ArrayBuffer(...), 'model-output.bin')
+```
+
+
+
+### `iodide.file.load(fileName, fileType[, variableName])`
+
+Returns a
+[Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises)
+that, when resolved, loads into the notebook the save file `fileName`. For each
```
mdboom:
```suggestion
that, when resolved, loads into the notebook the save file `fileName`. Each
```

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+notebook can find a list of uploaded files in the [upload modal](#link-required)
+or on your notebook's [Revisions Page](#needs-link). You can also access any
+uploaded file through the fetch chunk, [following this
+pattern](https://iodide-project.github.io/docs/workflows/#uploading-data-to-an-iodide-notebook).
+For most use cases using [fetch
+chunks](https://iodide-project.github.io/docs/jsmd/#fetch-chunks-fetch) is
+preferable and more straightforward. In more dynamic programmatic cases,
```
mdboom:
```suggestion
preferable and more straightforward. In more dynamic cases,
```

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+however, `iodide.file.load` can provide more nuanced workflows.
+
+`fileName` is the name of the file uploaded to the Iodide server. 
+
+`fileType` is the file type to handle. These are the same as the file types in
+[fetch chunks](https://iodide-project.github.io/docs/jsmd/#fetch-chunks-fetch):
+`js` (load this file as javascript, eg a library), `css` (load this file as a
+stylesheet, applying the styles to the report), `json` (load this file as json
+and parse into a javascript object), `text` (load this file as text), `blob`
+(load this file as binary).
```
mdboom: also `arrayBuffer`?

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+browser `window` namespace.
+
+
+#### `iodide.file.load` patterns
+
+```javascript
+
+// load a csv
+iodide.file.load('cached-data.csv', 'text', 'cachedData').then(() => {
+  cachedData = d3.csvParsed(cachedData)
```
mdboom: It's a little confusing to use the same variable name here for input and output, particularly since the lack of a `var` declaration makes its scope a bit ambiguous.

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+
+### `iodide.file.exists(fileName)`
+
+Returns `true` if the file `fileName` is available to the notebook, and `false`
+otherwise.
+
+The below code chunk, for instance, checks to see if a csv file already exists
+on the server. If it does, then Iodide will load that. Otherwise, the notebook
+will fetch the data, process it using a numb er of steps, and save the output to
+the server.
```
mdboom: Missing the code chunk this paragraph is referring to?

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+      return iodide.file.save(evictionsData, 'dataset.csv', d3.csvFormat)
+    })
+}
+```
+
+
+
+### `iodide.file.lastUpdated(fileName)`
+
+returns a `Date` object that represents when the file associated with `fileName`
```
mdboom:
```suggestion
Returns a `Date` object that represents when the file associated with `fileName`
```

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
 tracker](https://github.com/iodide-project/iodide/issues/new).
 
+
+## `iodide.file`
+
+The `iodide.file` API provides convenience functions around working with files
+uploaded to the Iodide server in your notebook.
+
+
+
```
wlach: Unnecessary newlines here, it looks like.

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
 Please direct clarifications or observations of inaccuracy to [our issue
 tracker](https://github.com/iodide-project/iodide/issues/new).
 
+
+## `iodide.file`
+
+The `iodide.file` API provides convenience functions around working with files
```
wlach: around -> for

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+
+
+
+### `iodide.file.load(fileName, fileType[, variableName])`
+
+Returns a
+[Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises)
+that, when resolved, loads into the notebook the save file `fileName`. For each
+notebook can find a list of uploaded files in the [upload modal](#link-required)
+or on your notebook's [Revisions Page](#needs-link). You can also access any
```
wlach: Don't forget to add these links.

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
 Please direct clarifications or observations of inaccuracy to [our issue
 tracker](https://github.com/iodide-project/iodide/issues/new).
 
+
+## `iodide.file`
+
+The `iodide.file` API provides convenience functions around working with files
+uploaded to the Iodide server in your notebook.
```
wlach: I think we need to make it clear that a user needs to have write access to a notebook (= owning it, currently) in order for these functions to work.

src/editor/reducers/notebook-reducer.js
```diff
@@ -215,6 +215,34 @@ const notebookReducer = (state = newNotebook(), action) => {
...
+      const { files } = state.notebookInfo;
+      if (!files.map(f => f.filename).includes(filename))
+        files.push({
+          filename,
+          lastUpdated,
+          id: fileID
+        });
+      else {
+        const i = files.findIndex(f => f.filename === filename);
+        files[i].lastUpdated = lastUpdated;
```
wlach: I think this could just be:

```js
files.find(f => f.filename === filename).lastUpdated = lastUpdated;
```

docs/api.md
wlach: `the save file`  => `the saved file`

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+chunks](https://iodide-project.github.io/docs/jsmd/#fetch-chunks-fetch) is
+preferable and more straightforward. In more dynamic programmatic cases,
+however, `iodide.file.load` can provide more nuanced workflows.
+
+`fileName` is the name of the file uploaded to the Iodide server. 
+
+`fileType` is the file type to handle. These are the same as the file types in
+[fetch chunks](https://iodide-project.github.io/docs/jsmd/#fetch-chunks-fetch):
+`js` (load this file as javascript, eg a library), `css` (load this file as a
+stylesheet, applying the styles to the report), `json` (load this file as json
```
wlach: Per your question, I am inclined not to support `js` and `css` for now. I don't see much benefit but lots of potential problems.

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+#### `iodide.file.load` patterns
+
+```javascript
+
+// load a csv
+iodide.file.load('cached-data.csv', 'text', 'cachedData').then(() => {
+  cachedData = d3.csvParsed(cachedData)
+})
+
+// load a video
```
teonbrooks: maybe include <video id="gritty" /> so that the user understands that this is where they are targeting the `gritty` obj

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+// load a style sheet and apply the style
+iodide.file.load('corporate-report-style.css', 'css')
+
+// load some javascript helpers
+iodide.file.load('analysis-helpers.js', 'js')
+```
+
+
+
+### `iodide.file.delete(fileName)`
```
teonbrooks: does anything happen if you attempt to delete and there isn't a file present?

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+
+
+### `iodide.file.exists(fileName)`
+
+Returns `true` if the file `fileName` is available to the notebook, and `false`
+otherwise.
+
+The below code chunk, for instance, checks to see if a csv file already exists
+on the server. If it does, then Iodide will load that. Otherwise, the notebook
+will fetch the data, process it using a numb er of steps, and save the output to
```
teonbrooks: numb er -> number

docs/api.md
hamilton: Good catch! It does, and I'll make sure to add this to the docs.

docs/api.md
hamilton: Will be adding that in a subsequent PR – didn't want to give too much surface area to test on this one.

docs/api.md
hamilton: Great point – will clarify this example.

docs/api.md
hamilton: yes, it throws an error, which can be catched in the return value Promise with `.catch`. I'll include an example that shows this.

docs/api.md
hamilton: good idea

docs/api.md
bcolloran: agreed

docs/api.md
```diff
@@ -6,6 +6,206 @@ challenges presented by JavaScript and web browsers.
...
+stylesheet, applying the styles to the report), `json` (load this file as json
+and parse into a javascript object), `text` (load this file as text), `blob`
+(load this file as binary).
+
+`variableName` (required for `json`, `text`, and `blob` file types, otherwise
+not applicable): the variable name in which to load the data, available in the
+browser `window` namespace.
+
+
+#### `iodide.file.load` patterns
```
bcolloran: patterns -> examples

src/shared/utils/file-operations.js
```diff
@@ -62,3 +73,42 @@ export function selectAndUploadFile(notebookID, successCallback = () => {}) {
...
+export function saveFileToServer(
+  notebookID,
+  data,
+  fileName,
+  fileID = undefined
+) {
+  const formData = makeFormData(notebookID, data, fileName);
+  const fetchRequest = fileID // if fileID undefined, upload, otherwise update
+    ? fd => updateFile(fileID, fd)
+    : fd => uploadFile(fd);
```
bcolloran: let's get rid of the `uploadFile` function altogether -- this logic can be handled within `updateFile` based on whether or not an fileId is passed in

docs/api.md
hamilton: I'll remove `js` and `css` but we will for sure need file type – otherwise, one would not be able to specify if it is a blob, text, json, or (before long) array buffer.

docs/api.md
hamilton: hmm, there aren't really links to be added here, it turns out. I thought there was something in the documentation.

docs/api.md
hamilton: ah, leftover. The description is below in the examples section. will remove

src/editor/port-to-eval-frame.js
hamilton: @bcolloran there's no logical place those should go right now. Any suggestions?

hamilton: on second thought maybe just `file-request-actions.js`

src/eval-frame/iodide-api/file.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+function confirmIsStringOrUndefined(key, argument) {
+  if (!isString(argument) && argument !== undefined) {
+    throw new Error(`
+    ${key} name must be a string or undefined, instead received ${typeof argument}`);
+  }
+}
+
+export function connectExists(store) {
```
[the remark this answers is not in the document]
hamilton: hmm, not sure if that would work in this case - that suggests that I can add that argument into, say, `iodide.file.exists('whatever.csv', {})` and I'll get an error because I kept an extra argument, when really I think I want to just ignore that extra argument (which is common with most JS functions). While this approach may not seem as clean as just being able to provide the extra argument, I think it is actually pretty straightforward to reason about, easy to test, and does what we need without exposing any additional surface.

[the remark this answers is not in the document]
bcolloran: oh i see. ya you're totally right -- different scenario in this case

src/shared/utils/file-operations.js
hamilton: hmm, might leave this as-is for now and file a follow-on PR. This is from the PR this is rebased on, not this PR itself, and @openjck is building off of this at the moment. For now I'd prefer to just leave a `FIXME` in that PR since it is not merged yet at any rate. I agree we should clean up this code but I think it'd be a good idea not to shake things up too much while a bunch of stuff depends on this.

hamilton:
@bcolloran this is up for another review when you get a chance.

src/shared/utils/file-operations.js
bcolloran: yep makes sense :+1:

src/eval-frame/iodide-api/file.js
```diff
@@ -0,0 +1,96 @@
...
+import sendFileRequestToEditor from "../tools/send-file-request-to-editor";
+import { FETCH_RETURN_TYPES } from "../../editor/state-schemas/state-schema";
+
+const DEFAULT_SAVE_OPTIONS = { overwrite: false };
+
+function isString(argument) {
+  return typeof argument === "string" || argument instanceof String;
+}
+
+function confirmIsString(key, argument) {
```
bcolloran: i really like that you are being aggressive about checking argument types and throwing when they are bad -- this will help users catch errors early, and i think we should implement it for all of our iodide API functions. it's such a good idea, in fact, that i think you should abstract it out into something re-usable in it's own file `validate-arguments` or something, and that we should make the apia bit more generic, maybe inspiresd by proptypes or something. could be like:

```
export function loadFile(fileName, fetchType, variableName = undefined) {
  argType.validate({
    fileName: [fileName, argType.string],
    fetchType: [fetchType, argType.enum(FETCH_RETURN_TYPES)],
    variableName: [variableName, argType.string.optional]
  })
  return sendFileRequestToEditor(fileName, "LOAD_FILE", {
    fetchType
  }).request.then(handleResourceLoad(fetchType, variableName));
}
```

no need to add any additional validation functions right now, but this is a great idea and i think we should go ahead and make a foundation we can build on. (and it occurs to me that there may be an existing lib that does this better than that strawman API, worth a quick google search)

src/eval-frame/tools/send-file-request-to-editor.js
```diff
@@ -0,0 +1,39 @@
+import generateRandomId from "../../shared/utils/generate-random-id";
+import messagePasserEval from "../../shared/utils/redux-to-port-message-passer";
+
+export const fileRequests = {};
+
+export default function sendFileRequestToEditor(
```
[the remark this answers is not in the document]
bcolloran: actually, looking into this more, i don't believe `fileRequestId` used anywhere except for in testing code, and for the testing case i think you could just mock `generateRandomId` to return a fixed string. let's remove that if it's only used for testing

src/eval-frame/iodide-api/file.js
```diff
@@ -0,0 +1,96 @@
...
+    const files = getFiles(getState());
+    const thisFile = files.filter(f => f.filename === fileName);
+    if (!thisFile.length) throw Error(`fileName ${fileName} does not exist`);
+    return new Date(thisFile[0].lastUpdated);
+  };
+}
+
+export function handleResourceLoad(fetchType, variableName) {
+  return file => {
+    if (FETCH_RETURN_TYPES.includes(fetchType) && variableName) {
```
bcolloran: `fetchType` has already been validated by your argument validation. this whole function seems like overkill if we use async/await, because then we won't need a terse function to use within a `then`

src/eval-frame/iodide-api/file.js
```diff
@@ -0,0 +1,114 @@
...
+    } else if (fetchType === "css") {
+      return addCSS(file);
+    } else if (fetchType === "js") {
+      return loadScriptFromBlob(file);
+    }
+    return file;
+  };
+}
+
+export function loadFile(fileName, fetchType, variableName = undefined) {
```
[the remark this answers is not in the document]
bcolloran: per comments above -- without `handleResourceLoad`, and with async and with validation

```
export async function loadFile(fileName, fetchType, variableName = undefined ) {
  argType.validate({
    fileName: [fileName, argType.string],
    fetchType: [fetchType, argType.enum(FETCH_RETURN_TYPES)],
    variableName: [variableName, argType.string.optional]
  })
  const file = await sendFileRequestToEditor(fileName, "LOAD_FILE", {fetchType})
  if (variableName && variableName.length > 0) window[variableName] = file;
  return file;
}
```

src/eval-frame/iodide-api/file.js
hamilton: True, I forgot to apply the async / await stuff to `file.js` as you noted, so I'll make sure to remove this

src/eval-frame/tools/send-file-request-to-editor.js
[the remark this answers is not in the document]
hamilton: Not true – `fileRequestId` is needed to actually resolve the file request. File name alone is not quite enough, because a user can send off two file requests for the same file at the same time. This minimally allows you to handle these issues gracefully, since the request itself is the unit of interest, not the file name.

This said, 100% right about moving to async / await here. I neglected to do that on the eval frame side - good catch!

src/eval-frame/iodide-api/file.js
hamilton: though this isn't about validation, but about loading the variable into `window`. all the same this will get refactored

src/eval-frame/tools/send-file-request-to-editor.js
```diff
@@ -0,0 +1,39 @@
...
+    request: new Promise((resolve, reject) => {
+      fileRequests[fileRequestID] = { resolve, reject };
+      messagePasserEval.postMessage("FILE_REQUEST", {
+        fileName,
+        requestType,
+        fileRequestID,
+        metadata
+      });
+    }).catch(err => {
+      throw err;
```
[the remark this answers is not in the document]
hamilton: hmm, thought I had fixed this ... might not have pushed my last set of changes, thanks for pointing this out

docs/api.md
```diff
@@ -6,6 +6,272 @@ challenges presented by JavaScript and web browsers.
...
+
+- `overwrite` (optional, default `false`): if `true`, will overwrite whatever is
+  at `fileName` with `data`. If `false`, and the file exists, the Promise will reject.
+
+
+#### `iodide.file.save` examples
+
+```javascript
+
+// upload a csv to the server. Assume the object csvData is an array of objects
```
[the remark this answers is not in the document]
hamilton: good point

src/editor/actions/file-request-actions.js
```diff
@@ -0,0 +1,169 @@
...
+}
+
+export function deleteFileFromNotebook(fileID) {
+  return {
+    type: "DELETE_FILE_FROM_NOTEBOOK",
+    fileID
+  };
+}
+
+function onFileOperationSuccess(fileRequestID, messagePasser, fileOrResponse) {
```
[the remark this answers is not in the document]
hamilton: ah yeah, I think overall I was originally in the position of creating a cyclical reference in this part of the code base, which is why I went this direction (vs the default arg). Will still need to check if that is still true!

src/editor/port-to-eval-frame.js
```diff
@@ -45,20 +45,17 @@ function receiveMessage(event) {
...
-            });
-          })
-          .catch(err => {
-            postMessageToEvalFrame("REQUESTED_FILE_ERROR", {
-              path: message.path,
-              reason: err.message
-            });
-          });
+      case "FILE_REQUEST": {
+        const { requestType, fileName, fileRequestID, metadata } = message;
```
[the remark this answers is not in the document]
hamilton: yeah that's a good point, will change

src/editor/reducers/notebook-reducer.js
```diff
@@ -215,6 +215,33 @@ const notebookReducer = (state = newNotebook(), action) => {
       return Object.assign({}, state, { viewMode });
     }
 
+    case "ADD_FILE_TO_NOTEBOOK": {
+      const { filename, lastUpdated, fileID } = action;
+      const { files } = state.notebookInfo;
+      if (!files.map(f => f.filename).includes(filename))
+        files.push({
```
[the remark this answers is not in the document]
hamilton: oh probably – not sure how that slipped by

src/editor/state-schemas/eval-frame-state-selector.js
```diff
@@ -15,7 +15,8 @@ const propsToCopy = [
   "panePositions.WorkspacePositioner",
   "savedEnvironment",
   "userDefinedVarNames",
-  "viewMode"
+  "viewMode",
+  "notebookInfo"
```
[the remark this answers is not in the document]
hamilton: it is passed down via `notebook_info` on the server, which certainly can change. This part of the code base is a bit sloppy imo

src/editor/state-schemas/state-schema.js
```diff
@@ -2,6 +2,8 @@ import { languageDefinitions } from "./language-definitions";
 
 export const NONCODE_EVAL_TYPES = ["css", "md", "meta", "raw"];
 export const RUNNABLE_CHUNK_TYPES = ["plugin", "fetch"];
+export const FETCH_TYPES = ["css", "js", "text", "blob", "json"];
```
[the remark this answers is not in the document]
hamilton: by default the file api can only load `RETURNABLE_FETCH_TYPES` which is below this, but I'll clarify the terminology here

src/editor/actions/file-request-actions.js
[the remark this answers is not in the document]
bcolloran: it won't create a circular ref if you import `messagePasserEditor` -- the message passer was originally implemented to let us avoid such circularities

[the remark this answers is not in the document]
hamilton: hmm ... ok. This definitely wasn't true during one part of this journey, but it might not have been from the message passer. All the same will go with the default arg

src/editor/actions/__tests__/file-request-actions.test.js
```diff
@@ -0,0 +1,251 @@
...
+  let store;
+
+  beforeEach(() => {
+    messagePasserMock = jest.fn();
+    deleteFileMock = jest.spyOn(FILE_OPS, "deleteFileOnServer");
+    deleteFileMock.mockReset();
+    store = mockStore(initialState());
+  });
+
+  it("fails if invalid arguments are passed in", () => {
```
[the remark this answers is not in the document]
hamilton: yep this is a pretty bad description :)

src/editor/actions/file-request-actions.js
```diff
@@ -0,0 +1,169 @@
...
+      return undefined;
+    }
+  };
+}
+
+export function handleFileRequest(
+  requestType,
+  fileName,
+  fileRequestID,
+  metadata,
```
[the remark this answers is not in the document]
hamilton: great point / agreed that this could be nicer

src/editor/state-schemas/eval-frame-state-selector.js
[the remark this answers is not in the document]
hamilton: also hadn't noticed I can include dots in the selector, so that will make things easier for sure

src/eval-frame/tools/send-file-request-to-editor.js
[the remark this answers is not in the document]
bcolloran: ya i know it's used to resolve the file request, but it looks like in that the value of `fileRequestId` is recieved by `onParentContextFileRequestSuccess` and `onParentContextFileRequestError` in _messages_ that come back from the editor, _not_ as the direct return value of `sendFileRequestToEditor`. i don't think that the `fileRequestID` value that is directly returned from `sendFileRequestToEditor` is used anywhere (and as a conceptual matter, it does not seem like it would need to be used anywhere, since the point of the ID is to set up a resolver that waits for the ID and send the ID over to the editor -- only the resolver needs to be able to see see the ID, nothing else in the eval-frame should need to be aware of it until the message comes back, right?)

src/eval-frame/iodide-api/file.js
```diff
@@ -0,0 +1,114 @@
...
+
+const DEFAULT_SAVE_OPTIONS = { overwrite: false };
+
+function isString(argument) {
+  return typeof argument === "string" || argument instanceof String;
+}
+
+function confirmIsString(key, argument) {
+  if (!isString(argument)) {
+    throw new Error(
```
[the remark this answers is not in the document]
hamilton: @bcolloran many apologies - this was collapsed down so I had missed a number of these comments! will address these on my next rev

src/editor/actions/file-request-actions.js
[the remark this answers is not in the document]
hamilton: oh, I see why the circular ref was there like 20 commits back and wouldn't be now – thanks for catching

src/editor/actions/file-request-actions.js
```diff
@@ -0,0 +1,169 @@
...
+  validateFileExistence,
+  validateFileAbsence,
+  validateFetchType
+} from "../tools/server-tools";
+
+function messagePasserIsFunction(passer) {
+  return typeof passer === "function";
+}
+
+function validateMessagePasser(passer) {
```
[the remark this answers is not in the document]
hamilton: yep

src/eval-frame/iodide-api/file.js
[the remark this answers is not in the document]
hamilton: this is a great idea, by the way

src/eval-frame/tools/send-file-request-to-editor.js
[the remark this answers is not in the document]
hamilton: oh I see your point. I can actually generate the ID as a default arg like we do pretty much everywhere else and that shouldn't affect the testability (seems pretty obvious now :) ). That all said, with the sum of simplifications, we don't really need async here since this function should just return a simple promise.

src/eval-frame/iodide-api/file.js
hamilton: At this point, this is the only bit of feedback I haven't implemented yet. I'd like to think about this a bit. On the one hand, I like the idea of using something for these types of validations, but on the other, I don't want to reinvent the wheel too much. Either way, given that the API surface of this PR is actually fairly small (save, load, delete, exists, lastUpdated, list) this might be more worthwhile to tackle in a follow-on PR.

bcolloran: agreed, this is not critical. filed #1760

src/eval-frame/tools/send-file-request-to-editor.js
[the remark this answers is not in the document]
bcolloran: i'd still like you to address this one if you don't mind. all async functions will return a promise automatically (wrap a return value in a promise) but for code readabilty i'd like us to never explicitly return a `Promise` in the code when we could just as easily write a function that returns a value.

i think the snippet below is clearer about what the intent of this function:
- we post a "FILE_REQUEST" message to the editor
- we await the result of the request
- we return the request result to the caller

```
export default async function sendFileRequestToEditor(
  fileName,
  requestType,
  options,
  fileRequestID = generateRandomId()
) {
  messagePasserEval.postMessage("FILE_REQUEST", {
    fileName,
    requestType,
    fileRequestID,
    options
  });

  const requestResult = await new Promise((resolve, reject) => {
      fileRequests[fileRequestID] = { resolve, reject };
  })

  return requestResult
}
```

As is, this function looks very mysterious to my eyes -- we're returning a promise that does some kind of funky thing that adds an object to the `fileRequests` object, but i need to go look at the call sites to understand the fate of that promise and what it is for.

src/shared/utils/file-operations.js
bcolloran: ...but please file a code quality issue linking back to this discussion before merging this

bcolloran:
oh also: be sure to merge master-- i updated jest the other day, and you appear to have a test failing now.

src/editor/tools/server-tools.js
```diff
@@ -47,3 +57,37 @@ export function getNotebookID(state) {
...
+export function getFileID(state, fileName) {
+  const files = getFiles(state);
+  const file = files.filter(f => f.filename === fileName)[0];
+  if (file) return file.id;
+  return undefined;
+}
+
+export function validateFileExistence(fileName, operation, state) {
+  if (!fileExists(fileName, state)) {
+    throw new Error(fileDoesNotExistMessage(operation, fileName));
```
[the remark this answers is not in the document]
hamilton: I already have a specialized error for this, so I'll make sure to apply it here.

hamilton:
@bcolloran I think we should talk about these things in-person, which would be faster for me to explain why things are the way they are.

src/editor/tools/server-tools.js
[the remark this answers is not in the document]
hamilton: I've replaced these with custom errors, but I don't think it actually buys much.

hamilton:
@bcolloran this is ready for another review.

PR #1705, opened 2019-04-02
add common file operations to src/shared/utils/file-operations.js

hamilton:
@openjck here are some of the functions that I think we'll both be able to leverage. Included here is `saveFileToServer`, `deleteFileOnServer`, and `loadFileFromServer` (which I don't think you'll need).

There's still a bit of file handling workflow to be implemented on your end, but a good example of that is in `notebook-actions-menu.jsx`, which handles the different states. It's not the nicest way to do things, but it works pretty well in practice. Could stand to be cleaned up.

Let me know if these work for you! And if you want to tighten up any of this functionality in your PR, that's also 100% fine – just figured it'd be easier to refactor and rebase off of a common thread.


src/server/components/delete-modal.jsx
```diff
@@ -13,7 +13,9 @@ export default class DeleteModal extends React.Component {
   }
 
   deleteObject() {
-    fetchWithCSRFTokenAndJSONContent(this.props.url, { method: "DELETE" })
+    // add delete here
```
openjck: Was this comment supposed to be removed?

src/server/components/delete-modal.jsx
```diff
@@ -4,7 +4,7 @@ import ModalTitle from "../../shared/components/modal-title";
 import ModalContent from "../../shared/components/modal-content";
 import ModalCall from "../../shared/components/modal-call";
 import { TextButton } from "../../shared/components/buttons";
-import { fetchWithCSRFTokenAndJSONContent } from "../../shared/fetch-with-csrf-token";
+import { deleteFileOnServer } from "../../shared/utils/file-operations";
```
bcolloran: we need to start grouping all of our server API functions in one place rather than dumping them in `utils` with a bunch of other unrelated stuff. this should go in a folder called `server-api` or `server-access-functions` or whatever you wish

bcolloran: also, don't import the API function `deleteFileOnServer` into the component -- the component should dispatch a thunk that is passed into it via mapStateToProps, b/c that will make the component more encapsulated and more easily testable

bcolloran: cc @wlach -- lets make sure that going forward we start gathering all of the code that talks to the Iodide server API in that location rather than sprinkling it throughout the codebase. the architecture we want to aim for is:

- helper functions that wrap details of the API calls and expose the specific functionality available from the server
- redux thunks that call the API wrappers
- components that dispatch redux thunks

(if this does not make sense to any of you, happy to talk about it further)

src/server/components/notebook-actions-menu.jsx
```diff
@@ -9,7 +9,7 @@ import {
   selectFileAndFormatMetadata,
```
bcolloran: this component needs help :-(

i'd originally wanted to suggest that we fix it right now, but i don't want to slow down the progress on the file upload API.

so rather than stopping to fix this now, please add this FIXME in this file:

FIXME this component needs to be refactored to be much dumber than it currently is. (1) it does not need to have internal state (2) it should just dispatch thunks that call `selectFileAndFormatMetadata`, `uploadFile`, and `updateFile` and handle the promises/`.then`s from those those API calls (3) it should not import those functions directly, it should import the action creators that wrap them, and those actions should be passed in via mapDispatchToProps

(or, i guess it's up to you... if you want to tackle this now i don't think it would actually take that long and it would be a fine use of time -- as our codebase grows in size and complexity, we'll need to aim for more architectural scalability, which will mean cleaning up things like this being stricter about our interfaces, so this would be a reasonable thing to spend an hour on towards that end)

src/server/components/delete-modal.jsx
hamilton: This has nothing to do with redux – it's a server React component that doesn't use thunks.  Both the server and the client both need the exact same functions like `deleteFileOnServer`, which was the point of this PR.

src/server/components/notebook-actions-menu.jsx
hamilton: These are server components, not client ones, so this isn't relevant. Since our server components are an afterthought to the rest of the app, I'd vote for not refactoring anything in this PR unless we prioritize actually designing the server components.

src/server/components/delete-modal.jsx
bcolloran: oh my apologies, disregard everything i said in this and the other in-line comment, and sorry for the distraction -- this is all looks good for non-notebook stuff.

the overall architectural direction still applies, though, so it'd still be better to have the API helper functions moved to their own folder in /shared rather than putting them in /shared/utils with other unrelated stuff. we'll want to move a bunch of the existing server API helper functions out of /editor and consolidate them in one place, and the amount of code we'll have is worthy of it's own folder rather than a bunch of files in /utils

PR #1714, opened 2019-04-07
Table rep2

bcolloran:
add a table viewer back to the reps, this time with a "details" box below the table. also sets up storybook to let us look at a bunch of reps stuff quickly-- run `npm run storybook` and visit http://localhost:6006 (storybook with hot reload is amaze for working on components)

still need to add some tests, but this has been super thoroughly QAed. i've tried to be as comprehensive as possible by creating a huge library of test case objects/values to run through the "tinyRep" code.

I'll add some stuff to the docs and changelog too. 

### Pull Request checklist


.eslintrc.js
```diff
@@ -43,5 +41,9 @@ module.exports = {
     "no-console": ["off"],
     // Prefer default exports
     "import/prefer-default-export": ["off"],
-  },
-}
+    "import/no-extraneous-dependencies": [
+      "error",
+      { devDependencies: ["**/__stories__/*", "**/__tests__/*", "/test/*"] }
+    ]
```
bcolloran: added this rule to fix linting problems in the storybook files, the rest of this is just prettier churn

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+const classInfoColor = "#808080";
+const stringColor = "rgb(196, 26, 22)";
+const separatorColor = "black";
+const dateColor = stringColor;
+const symbolColor = stringColor;
+const errorColor = "";
+const functionColor = "rgb(28, 30, 207)";
+const ellipsisColor = "#111 ";
+
+const RepBaseStyledSpan = color => styled("span")`
```
bcolloran: @hamilton i'm not sure if i'm using styled components idiomatically/optimally/correctly throughout this PR. lmk what you think.

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+  RegExp: regexRep,
+  Map: objectRep,
+  Set: objectRep,
+  WeakSet: objectRep,
+  WeakMap: objectRep
+};
+
+const handledTypes = Object.keys(typeToTinyRepMapping);
+
+export default function tinyRep(serializedObj) {
```
bcolloran: note that this tinyRep function (and the sub-components it relies on) are not tested. IMO, since these are basically little tiny components that really just have to look a certain way and that won't have side effects or behaviors/dispatches that affect anything else in the app, the exhaustive set of storybook cases is satisfactory for QA testing that the tinyRep function (1) returns components that look the way we want, and (2) that the tiny rep does not crash when fed any possible input.

test/fileMock.js
```diff
@@ -0,0 +1 @@
+module.exports = "test-file-stub";
```
bcolloran: needed to get tests working with css/file imports

test/styleMock.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
bcolloran: needed to get tests working with css/file imports

src/eval-frame/components/reps/data-table-rep.jsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  render() {
+    const { value } = this.props;
+
+    const columns = Object.keys(value[0]).map(k => ({
+      Header: k,
+      accessor: k,
+      Cell: cell => {
+        return tinyRep(tinyRepSerializer(cell.value));
```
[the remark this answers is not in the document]
bcolloran: Oh but the UX idea here overall is that we just show a "tiny" sized rep in the table itself (to keep the table managable and reasonable) and that you can dig into the details in the expansion thing below the table. @hamilton suggested this approach in  a conversation a few weeks ago. i quite like it so far! (we can refine in the future if we have more ideas after dogfooding for a while)

src/eval-frame/components/reps/__test_helpers__/reps-test-value-cases.js
```diff
@@ -0,0 +1,283 @@
...
+    time: new Date(),
+    id: Math.sin(i)
+  })),
+  rowsTable_compositeObjects: new Array(100).fill(compositeObjects),
+  rowsTable_objectsContainingSimpleTypes: new Array(100).fill(simpleTypes),
+  rowsTable_objectsContainingBaseObjects: new Array(100).fill(baseObjects)
+};
+
+export const rowTableFails = {
+  array_oneEmptyObject: [{}],
```
bcolloran: note that for testing purposes we've created a bunch of tablular-ish things that should *not* be represented as tables. not also that `arrayOfBaseObjects_emptyArray` covers the case of an array of empty objects, which we should not show as a table.

bcolloran:
@hamilton and @mdboom -- this PR is not ready to merge, but it's ready for review. note that many of the file changes are just things that have to do with getting storybook set up, so you can kind of breeze pass them during review. I've also tried to leave a bunch of notes to help you find your way.

changelog, docs, and a couple more tests to come

src/eval-frame/components/reps/rep-utils/rep-type-chooser.js
```diff
@@ -0,0 +1,35 @@
...
+  return true;
+}
+
+export function isRowDf(obj) {
+  if (!_.isArray(obj) || obj.length < 2 || !_.isPlainObject(obj[0])) {
+    return false;
+  }
+  const rowsToCheck = Math.min(ROWS_TO_CHECK, _.size(obj));
+  for (let i = 1; i < rowsToCheck; i++) {
+    if (!_.isPlainObject(obj[i])) {
```
mdboom: Could this be made to also check for an array of small arrays?  That would cover the case where the data is from a CSV without a header row, or a Numpy structured array.

src/eval-frame/components/reps/rep-utils/tiny-rep-serializer.js
```diff
@@ -0,0 +1,80 @@
...
+const typesWithByteLength = ["ArrayBuffer", "DataView"];
+
+const typesWithSize = ["Map", "Set"];
+
+export function objSize(obj) {
+  if (obj === null) return undefined;
+  if (obj === undefined) return undefined;
+  const type = getType(obj);
+  if (type === "RegExp") return obj.source.length;
+  if (typesWithLength.includes(type)) return obj.length;
```
mdboom: Could we just check whether `obj.length !== undefined` (i.e. duck type it)?  Then this would capture as of yet unknown classes.

src/eval-frame/components/reps/rep-utils/tiny-rep-serializer.js
```diff
@@ -0,0 +1,80 @@
...
+    ? [s.slice(0, TRUNCATION_LEN), true]
+    : [s, false];
+
+export function tinyRepStringify(obj) {
+  const type = getType(obj);
+  let stringVal;
+  if (["Number", "Boolean", "Undefined", "Null", "Symbol"].includes(type)) {
+    stringVal = String(obj);
+  } else if (type === "Date") {
+    stringVal = obj.toISOString().slice(0, 19);
```
mdboom: What does this end up stripping off?  Time zone?

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+  <span>
+    <Sep>{lQuote}</Sep>
+    <StringSpan>{obj.stringValue}</StringSpan>
+    {obj.isTruncated ? <Ell>⋯</Ell> : ""}
+    <Sep>{rQuote}</Sep>
+    {obj.isTruncated ? <ClassInfoSpan>({obj.size})</ClassInfoSpan> : ""}
+  </span>
+);
+
+const stringRep = quotedStringRep("“", "”");
```
mdboom: I'm not a fan of using "typographic" quotes in this context, because it means you can't copy and paste the quotes into code.

src/eval-frame/components/reps/rep-utils/tiny-rep-serializer.js
bcolloran: time zone and milliseconds

src/eval-frame/components/reps/tiny-reps.jsx
bcolloran: sure, that makes sense

src/eval-frame/components/reps/rep-utils/tiny-rep-serializer.js
bcolloran: the problem with duck typing in this case is that a user i might very plausibly define e.g.:
```
couch = {height: 3, length:5, depth:2.5, color: "puce", material: "cotton"}
```
in fact, i think i may even have a notebook with objects that have `length` property used this way. i think we're stuck with this non-duck approach, since js `obj.length` has less convenient semantics than python `len(obj)` :-/

(my general sense is that a lot of the stuff around reps will require this kind of careful and fiddly hand-curation and enumeration)

src/eval-frame/components/reps/rep-utils/rep-type-chooser.js
bcolloran: i'd prefer to keep the semantics very clear about what these objects are, b/c i think it will make the code clearer and the testing easier, and will also allow us to keep fine-grained control on what components we're using and make sure we're using them correctly. for example a `rowDf` is always an array of >1 plain objects that have all have identical keys, and it is meant to be shown in a tabular view with column headings, so passing in an array of arrays would break the expectations of the component downstream.

i think it will be easy to expand out from here in future PRs to add things like columnar arrays, arrow objects, tablular things without column names, etc. but we'll want to match those in separate type-chooser functions to make sure that we're handling them correctly in the corresponding component

src/eval-frame/components/reps/value-renderer.jsx
```diff
@@ -4,7 +4,25 @@ import PropTypes from "prop-types";
...
 import HTMLHandler from "./html-handler";
+import TableRenderer from "./data-table-rep";
 import UserReps from "./user-reps-manager";
+import { isRowDf } from "./rep-utils/rep-type-chooser";
+
+function repChooser(value) {
+  if (UserReps.getUserRepIfAvailable(value)) {
+    return "GLOBAL_USER_RENDERER";
+  }
+  if (value && value.iodideRender instanceof Function) {
```
hamilton: what's the diff between this and the `GLOBAL_USER_RENDERER`? Is there a reason we're supporting both with two different bits of functionality? I guess one might be provided by a language, the other not. If that's the case, it'd be great to have flags that are a bit closer to what the use case actually is.

src/eval-frame/components/reps/__stories__/reps.stories.js
```diff
@@ -0,0 +1,84 @@
...
+
+allTinyReps.add("all the tiny reps", () => {
+  return (
+    <table>
+      {Object.entries(allCases).map(caseNameAndVal => {
+        const [name, value] = caseNameAndVal;
+        return (
+          <tr key={name}>
+            <td>{name}</td>
+            <td>{tinyRep(tinyRepSerializer(value))}</td>
```
hamilton: my comments re: the data table & using `<TinyRep value={value} />` apply here, but additionally it would be great to put the non-tiny rep here for comparison as well.

src/eval-frame/components/reps/rep-utils/tiny-rep-serializer.js
```diff
@@ -0,0 +1,80 @@
...
+    stringVal = obj;
+  } else if (["Function", "GeneratorFunction"].includes(type)) {
+    stringVal = obj.name;
+  } else {
+    stringVal = "";
+  }
+  return truncateString(stringVal);
+}
+
+export function tinyRepSerializer(obj) {
```
hamilton: feeling very nit-picky about a lack of verbs for these pure functions, but maybe`serializeForTinyRep` feels better to me

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+`;
+
+const tinyStringValueWithColor = color => {
+  const ValueSpan = RepBaseStyledSpan(color);
+  return x => <ValueSpan>{x.stringValue}</ValueSpan>;
+};
+
+const ClassInfoSpan = RepBaseStyledSpan(classInfoColor);
+
+const objectRep = obj => (
```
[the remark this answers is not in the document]
hamilton: you could also destructure obj so it is clearer to readers what parts of the passed-in object are used

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
+import React from "react";
```
hamilton: this entire fire could use prop types on all of these functional components. In cases where you have a factory, the functional component the factory creates could easily get its proptypes also assigned

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+  RegExp: regexRep,
+  Map: objectRep,
+  Set: objectRep,
+  WeakSet: objectRep,
+  WeakMap: objectRep
+};
+
+const handledTypes = Object.keys(typeToTinyRepMapping);
+
+export default function tinyRep(serializedObj) {
```
hamilton: i would love for this to be `TinyRep`. Anything we can do to suggest that this is a function component (which it technically is) would be great.

src/eval-frame/components/reps/tiny-reps.jsx
hamilton: I might also suggest calling this file `tiny-rep.jsx` since that's the thing we're technically exporting from it

src/eval-frame/components/reps/__test_helpers__/reps-test-value-cases.js
```diff
@@ -0,0 +1,283 @@
...
+  longValueArrays[`array_long_${type}`] = new Array(N_ARRAY).fill(
+    simpleTypes[type]
+  );
+});
+longValueArrays.array_long_mixed = new Array(1000)
+  .fill(Object.values(simpleTypes))
+  .reduce((a, b) => a.concat(b), []);
+
+export const arrayBuffers = {
+  buffer_arrayBuffer: new ArrayBuffer(N_ARRAY),
```
hamilton: what about `Blob`? 

(sidenote - remember the movie ["What About Bob?"](https://en.wikipedia.org/wiki/What_About_Bob%3F)? did that one age well?)

src/eval-frame/components/reps/__test_helpers__/reps-test-value-cases.js
```diff
@@ -0,0 +1,283 @@
...
+
+// ==================== tabular
+
+export const rowTableCases = {
+  rowsTable_plainObjects: new Array(100).fill(0).map((x, i) => ({
+    index: i,
+    time: new Date(),
+    id: Math.sin(i)
+  })),
+  rowsTable_compositeObjects: new Array(100).fill(compositeObjects),
```
hamilton: Could we actually randomly generate some of the values in `compositeObjects` so it's not a big table that never changes when I paginate etc.? I'm not very bright so I had to go check to see if it was doing the thing i wanted it to do.

src/eval-frame/components/reps/react-table-styles.css
```diff
@@ -0,0 +1,46 @@
+.ReactTable {
```
hamilton: I am noticing that the `ReactTable` does not by default stretch to fill its space. Maybe this is a weird interaction w/ the grid?

Screen Shot 2019-04-10 at 1 59 48 PM

src/eval-frame/components/reps/data-table-rep.jsx
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }));
+
+    const pageSize = value.length > 10 ? 10 : value.length;
+    return (
+      <div>
+        <ReactTable
```
hamilton: hmm, noticing I can't resize the columns without everything breaking. Maybe let’s deactivate column resize if we’re going to keep things like this, or fix it.

Apr-10-2019 13-52-12

src/eval-frame/components/reps/react-table-styles.css
```diff
@@ -0,0 +1,46 @@
...
+.ReactTable .-pagination {
+  padding: 2px;
+  box-shadow: none;
+  border: none;
+}
+
+.ReactTable .-pagination .-pageSizeOptions {
+  margin: 0px 10px;
+}
+
```
hamilton: adding this style might help others realize they can click a value and get the detailed rep:
```css
div.rt-td {
    cursor: pointer;
}
```

src/eval-frame/components/reps/data-table-rep.jsx
```diff
@@ -0,0 +1,94 @@
...
+          data={value}
+          columns={columns}
+          showPaginationTop
+          showPaginationBottom={false}
```
hamilton: I would _LOVE_ it if we could, when you click a cell, change its background color so I know which one is the one that gives me the detailed view. Or just _something_ – I'm finding it hard to know what I've selected / am looking at. I love that we provide the index though!

src/eval-frame/components/panes/console/console-message.jsx
```diff
@@ -15,12 +14,13 @@ const MessageContainer = styled(ConsoleContainer)`
 `;
 
 const MessageBody = styled("div")`
-  font-family: ${THEME.client.console.fontFamily};
```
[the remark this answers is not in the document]
hamilton: all for removing it if it makes things easier, but I'm noticing that the individual child reps aren't handling their own styling, so I would suggest going through the other console components and adding them. I noticed for python, for instance, that the font is not set to `THEME.client.console.fontFamily`. Should be an easy fix!

Screen Shot 2019-04-10 at 3 01 50 PM

For reference, here it is on master:

Screen Shot 2019-04-10 at 3 02 47 PM

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+const tinyStringValueWithColor = color => {
+  const ValueSpan = RepBaseStyledSpan(color);
+  return x => <ValueSpan>{x.stringValue}</ValueSpan>;
+};
+
+const ClassInfoSpan = RepBaseStyledSpan(classInfoColor);
+
+const objectRep = obj => (
+  <ClassInfoSpan>
+    {obj.class}({obj.size})
```
hamilton: something to consider playing with:

(1) moving to comma int, so that we're not stuck with `100000` but instead `100,000` will make it easier to read.
(2) moving the parens to `<sub>{obj.size}</sub>` would make the total volume subscripted and less of a big wall of stuff
(3) if you don't like (2) (I'm not even sure how I feel about it) perhaps reduze the text weight for `obj.size` to `300` or something smaller than `normal` so it is a bit easier to parse. The monospace rendering as-is can get tough on the eyes when there are a lot of values in the object.

this might be the wrong approach, but this was something I noticed while looking through the storybook

src/eval-frame/components/panes/console/console-message.jsx
```diff
@@ -15,12 +14,13 @@ const MessageContainer = styled(ConsoleContainer)`
...
   margin: auto;
   margin-left: 0;
   padding-top: 5px;
   padding-bottom: 5px;
   min-height: 20px;
+  max-width: calc(100% - 5px);
+  overflow-x: hidden;
```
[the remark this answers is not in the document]
hamilton: I have a comment elsewhere, but on my machine the table doesn't stretch the full width of the console.

src/eval-frame/components/reps/data-table-rep.jsx
```diff
@@ -0,0 +1,94 @@
...
+    );
+  }
+  return (
+    <TableDetails>
+      <TableDetailsMessage>Click a table cell for details</TableDetailsMessage>
+    </TableDetails>
+  );
+};
+
+export default class TableRenderer extends React.Component {
```
hamilton: The `TableRenderer` has some weird mobile CSS applied where the buttons actually get _taller_ when the width is limited. I'm assuming this is because the table thinks we're on a mobile device. Any chance we can override that?

Screen Shot 2019-04-10 at 3 55 29 PM

src/eval-frame/components/reps/tiny-reps.jsx
hamilton: i think this is fine, but as I mention elsewhere we should add prop types to them

src/eval-frame/components/panes/console/console-message.jsx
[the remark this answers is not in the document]
bcolloran: ah! thanks for catching that!

src/eval-frame/components/reps/data-table-rep.jsx
```diff
@@ -0,0 +1,94 @@
...
+  static propTypes = {
+    value: PropTypes.any.isRequired
+  };
+  constructor(props) {
+    super(props);
+    this.state = { focusedPath: undefined };
+  }
+
+  handleCellClick(rowInfo, column) {
+    this.setState({ focusedPath: `[${rowInfo.index}]["${column.id}"]` });
```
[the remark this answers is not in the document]
bcolloran: oh, definitely. good idea.

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+  Number: tinyStringValueWithColor(numberColor),
+  Boolean: tinyStringValueWithColor(boolColor),
+  Undefined: tinyStringValueWithColor(nullUndefinedColor),
+  Null: tinyStringValueWithColor(nullUndefinedColor),
+  String: stringRep,
+  Symbol: symbolRep,
+  Array: objectRep,
+  ArrayBuffer: objectRep,
+  DataView: objectRep,
+  Int8Array: objectRep,
```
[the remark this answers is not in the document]
bcolloran: yep np :-)

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+const stringColor = "rgb(196, 26, 22)";
+const separatorColor = "black";
+const dateColor = stringColor;
+const symbolColor = stringColor;
+const errorColor = "";
+const functionColor = "rgb(28, 30, 207)";
+const ellipsisColor = "#111 ";
+
+const RepBaseStyledSpan = color => styled("span")`
+  color: ${color};
```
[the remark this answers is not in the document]
bcolloran: perfect, thanks! :+1:

src/eval-frame/components/reps/data-table-rep.jsx
bcolloran: :+1:

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+    ) : (
+      ""
+    )}
+  </span>
+);
+
+const FunctionSpan = styled(RepBaseStyledSpan(functionColor))`
+  font-style: italic;
+`;
+const functionRep = obj => (
```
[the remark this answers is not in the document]
bcolloran: :+1:

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+  <span>
+    <Sep>{lQuote}</Sep>
+    <StringSpan>{obj.stringValue}</StringSpan>
+    {obj.isTruncated ? <Ell>⋯</Ell> : ""}
+    <Sep>{rQuote}</Sep>
+    {obj.isTruncated ? <ClassInfoSpan>({obj.size})</ClassInfoSpan> : ""}
+  </span>
+);
+
+const stringRep = quotedStringRep('"', '"');
```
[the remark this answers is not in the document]
bcolloran: good call

src/eval-frame/components/reps/tiny-reps.jsx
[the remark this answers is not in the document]
bcolloran: :+1:

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+const objectRep = obj => (
+  <ClassInfoSpan>
+    {obj.class}({obj.size})
+  </ClassInfoSpan>
+);
+
+const Sep = RepBaseStyledSpan(separatorColor);
+const Ell = RepBaseStyledSpan(ellipsisColor);
+const StringSpan = RepBaseStyledSpan(stringColor);
+const quotedStringRep = (lQuote, rQuote) => obj => (
```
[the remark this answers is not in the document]
bcolloran: :+1:

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
+import React from "react";
+import styled from "react-emotion";
+
+const numberColor = "rgb(28, 0, 207)";
```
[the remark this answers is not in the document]
bcolloran: would `reps-theme.js` in this folder be ok? this whole section feels like it may get carved into a stand-alone module at some point

src/eval-frame/components/reps/rep-utils/tiny-rep-serializer.js
```diff
@@ -0,0 +1,80 @@
...
+export const getType = obj =>
+  Object()
+    .toString.call(obj)
+    .split(" ")[1]
+    .slice(0, -1);
+
+const typesWithLength = [
+  "Array",
+  "String",
+  "Int8Array",
```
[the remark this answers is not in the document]
bcolloran: :+1:

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+    <Sep>{rQuote}</Sep>
+    {obj.isTruncated ? <ClassInfoSpan>({obj.size})</ClassInfoSpan> : ""}
+  </span>
+);
+
+const stringRep = quotedStringRep('"', '"');
+const regexRep = quotedStringRep("/", "/");
+
+const SymbolSpan = RepBaseStyledSpan(symbolColor);
+const symbolRep = obj => (
```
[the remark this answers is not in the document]
bcolloran: i totally agree, great idea :+1:

src/eval-frame/components/reps/tiny-reps.jsx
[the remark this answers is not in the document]
hamilton: That sounds good to me. Truth be told, I can see down-the-line having one consistent theme for editor + reps, but that's ... pretty far away :)

src/eval-frame/components/reps/tiny-reps.jsx
```diff
@@ -0,0 +1,121 @@
...
+    <FunctionSpan>
+      ƒ{obj.class === "GeneratorFunction" ? "*" : ""} {obj.stringValue}
+    </FunctionSpan>
+    {obj.isTruncated ? <Ell>⋯</Ell> : ""}
+    <FunctionSpan>()</FunctionSpan>
+  </span>
+);
+
+const ErrorSpan = RepBaseStyledSpan(errorColor);
+const errorRep = obj => <ErrorSpan>{obj.class}</ErrorSpan>;
```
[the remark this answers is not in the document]
bcolloran: :+1:

src/eval-frame/components/reps/value-renderer.jsx
bcolloran: the difference between the GLOBAL_USER_RENDERER and OBJECT_USER_RENDER_METHOD is that GLOBAL_USER_RENDERER "globally" checks all values that are returned from an eval to see whether they match the criteria for the user rendered, whereas the other case uses a render method in the `iodideRender` property.

the former is set using `iodide.addOutputRenderer(rendererSpecification)` (see https://iodide-project.github.io/docs/api/ ); the latter is a method manually attached to any object.

these were distinct use cases that @mdboom envisioned, and they seem reasonable; wrt to the terminology, i'm certainly open to suggestions. do either of you two have ideas?

src/eval-frame/components/reps/value-renderer.jsx
```diff
@@ -39,18 +57,20 @@ export default class ValueRenderer extends React.Component {
...
-
-    const htmlString = UserReps.getUserRepIfAvailable(value);
-    if (htmlString) {
-      return <HTMLHandler htmlString={htmlString} />;
-    }
-    if (value && value.iodideRender instanceof Function) {
-      return <HTMLHandler htmlString={value.iodideRender()} />;
+    const repType = repChooser(value);
+    switch (repType) {
+      case "GLOBAL_USER_RENDERER": {
```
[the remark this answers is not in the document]
bcolloran: following up in comment thread above...

src/eval-frame/components/reps/data-table-rep.jsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  render() {
+    const { value } = this.props;
+
+    const columns = Object.keys(value[0]).map(k => ({
+      Header: k,
+      accessor: k,
+      Cell: cell => {
+        return tinyRep(tinyRepSerializer(cell.value));
```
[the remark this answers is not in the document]
bcolloran: sure sure-- to clarify further: i specifically *don't* want the `TinyRep` -- or *any* of the reps -- to directly take a value, because i'm actually looking down the line to the day when the serialization of objects occurs in the eval-frame, but the reps themselves are rendered in a different iframe. Hence, the serialization is intentionally not handle within the component. bonus consequence of separating this out is extra clarity around testing, storybook, etc.

src/eval-frame/components/reps/value-renderer.jsx
hamilton: got it, that makes a lot of sense, and I like that distinction (which is obvious now from the code, just didn't realize we had `.iodideRender` :) ). We want a way for users to dynamically render content that doesn't involve adding an output renderer, which I'm sure is useful for all sorts of things including language plugins. I think I was thrown off by the term `GLOBAL_USER_RENDERER`, since in my head it's more like `USER_REGISTERED_RENDERER` or `CUSTOM_REGISTERED_RENDERER` or something like that, since it suggests that the thing was added to the rep checker

src/eval-frame/components/reps/rep-utils/tiny-rep-serializer.js
bcolloran: ya i think that makes sense, thanks for catching these

src/eval-frame/components/reps/__stories__/reps.stories.js
bcolloran: addressed `<TinyRep value={value} />` in the other thread;

re the "non-tiny" rep -- do you mean the normal react-inspector? in any case, this story is really meant to just show the tiny-reps. also, including the react-inspector reps would probably blow up the page since some of the things here are e.g. giant strings that react-inspector does not handle well at all. (all of the standard reps are available in their own stories to prevent that)

src/eval-frame/components/reps/tiny-reps.jsx
bcolloran: :+1:

src/eval-frame/components/reps/tiny-reps.jsx
bcolloran: can do!

src/eval-frame/components/reps/__test_helpers__/reps-test-value-cases.js
bcolloran: i'm all for adding more cases. i think it's generally considered wise to avoid random inputs in tests (unless you are specifically doing fuzzing), but if you want to send me a few more deterministic recipes, i'll def pop them in. the more cases we have here, the better!

src/eval-frame/components/reps/react-table-styles.css
bcolloran: :+1:

src/eval-frame/components/reps/__stories__/reps.stories.js
hamilton: Yep that's what I mean. I suppose down the line as others work on the reps and use the storybook to get acclimated, they know what they're comparing things to, especially if we end up rolling our own longer reps. I understand they'd definitely blow up on some of the longer ones as is, so probably not worth adding them there, and I imagine the point of the current test set is to show how big objects / primitives boil down. Maybe the solution is to add another storybook page for a subset of sensible data values so comparison is easier. This all said, doesn't have to be done in this PR if you think it's not really worthwhile!

src/eval-frame/components/reps/data-table-rep.jsx
bcolloran: ya that's the thing i was mentioning about how it's dumb that react-table is not actually build on an real <table> element, which would let the browser guarantee that the headers always align correctly :-/

i've seen something like this (tho not exactly what your gif shows) happening in the storybook, so i think it might be a problem with react-table rather than an interaction with containing styles, so we might not be able to fix it, in which case i think you're right that disabling column resizing is a good way forward. (and that should be fine for our use case since the columns just show the tiny reps, which don't get too wide)

src/eval-frame/components/reps/tiny-reps.jsx
bcolloran: interesting. i don't have strong opinion about any of these suggestions, and none of them strike me as "bad", but i guess i'm initially inclined to follow the example from chrome and FF devtools until we've dogfooded enough to see where pain points show up, at which point we can refine? does that sound ok?

src/eval-frame/components/reps/data-table-rep.jsx
bcolloran: ya i can take a look

bcolloran:
thanks @hamilton ad @mike, super useful comments. i've only had time to look through, i may be able to do a revision before i go on vacation, but it might not be until next week.

src/eval-frame/components/reps/tiny-reps.jsx
hamilton: sounds sensible :)

src/eval-frame/components/reps/tiny-reps.jsx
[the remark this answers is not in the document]
bcolloran: upon reflection, i think that it may be most appropriate to keep these in this file for now until it becomes clearer what the future around this is. ATM this is a little snowflake-y b/c this theme was actually chosen to match 'react-inspector', and since i agree that having one set of colors in the editor and reps is the best long term plan, i'm not sure it makes sense to mix this temporary snowflake with the rest of the theme, nor to break it out since nothing else will use it in the immediate future.

src/eval-frame/components/reps/value-renderer.jsx
bcolloran: changed to "USER_REGISTERED_RENDERER"

src/eval-frame/components/reps/__test_helpers__/reps-test-value-cases.js
bcolloran: and of course, the idea is to keep adding more test values as we go, so this list need never be considered "done". b/c this is kind of a permanent work-in-progress, i'm going to mark this thread resolved, but do feel free to send me recipes and i'll pop them in if you get them to me while i'm working on this

src/eval-frame/components/reps/__test_helpers__/reps-test-value-cases.js
bcolloran: thanks, added.

let's just keep adding more things as we think of them in the future. (at some point soon we'll need to add BigInt for example. Chrome already has it, but I don't think we need to add it yet)

src/eval-frame/components/panes/console/console-message.jsx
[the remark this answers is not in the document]
bcolloran: looking at this again, i don't think that "system messages" need to by monospaced, just things that "look like code" as it were. i actually prefer the system messages to not look like code -- i think it sets them apart visually.

anyhoo, i'm not attached to that viewpoint, especially since we have already decided that enhanced console handling will be coming up soon, which will give us another chance to mark things out visually... but since that effort will be coming up soon anyway, and since i think this look might actually be _better_ i'm going to hold off for now and we can revisit in a couple weeks when we dig into the console again.

src/eval-frame/components/reps/react-table-styles.css
bcolloran: i was playing around with this, and i believe that it's actually more readable to not stretch the table if there are just a couple columns. let's try dogfooding it a bit as-is -- i'm not committed to the way it is now, and if we think it will be better if it always stretches, we can make that change

src/eval-frame/components/panes/console/console-message.jsx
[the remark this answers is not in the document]
hamilton: Don't agree with this assessment, which is why I had it explicitly set to monospace before (and started this comment thread). No other console system we or our users use makes this distinction, and I'm not sure why we're doing that here. But I suppose it can get addressed some other time since this has merged.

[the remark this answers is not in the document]
bcolloran: sure sure -- no other console does that, but i think that's probably because in most consoles either (a) they can _only_ display a single font, which has historically been fixed-width from the olden days of green pixel terminal, or (b) they are trying to _emulate_ the appearance of the terminals from the ancient days. But there is nothing sacrosanct about that appearance, and i suspect that it's mostly just a historical accident that most terminals today follow that tradition. it seems that there may be advantages having code (and code like things) be fixed-width, but for messages it really doesn't matter at.

I don't care about this all that much if you have strong feelings, but there is for sure a semantic distinction between our system/app messages and our code inputs and outputs. how we want to indicate that distinction to our users is up to us; maybe we don't need to do anything at all, and it's best to have system messages blend in with the rest of the console. But the fact that other other console systems have not drawn that distinction is beside the point -- most text-based consoles do not have the ability to make that kind of distinction clear with anything other than text color (which is frequently used to indicate different kinds of content). no other console system we use embeds interactive table viewers either -- we can always try something new when what has come before is the result of historical accident and doing something different might be better

so ya: whatever you want to do when we revisit the console will be fine with me :-) but i think it would be nice to recognize this distinction and to go in with an open mind as we think about whether and how to signify that distinction to our users. we can do what we want: we might decide that fixed-width should always be used in the console because it signifies "you are looking at something terminal-like" (in which case we might want to apply that styling in throughout the table as well?), or we might decide that fixed-width specifically means "you are looking at something code-like" -- inputs and outputs.

an empire of signs! let's keep our semiotics on point :-)

PR #1726, opened 2019-04-09
solve high cpu usage of start-and-serve

louisabraham:
Hi,
I experienced high (200%) CPU usage of node even when passively serving static files. This solves the issue.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -107,7 +107,7 @@ module.exports = env => {
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


bcolloran:
awesome fix! thanks much @louisabraham !!

louisabraham:
Little precision: I am really not a user of webpack so don't blame me if I break something 😅

wlach:
OMG thank you!!

PR #1734, opened 2019-04-12
Only show notebooks that have 10 or more edits in the server (partially fixes #1710)

### LINKED ISSUE #1710: Should only show notebooks on the front page that have been changed in some interesting way
Right now the "firehose of notebooks" on alpha.iodide.io is dominated by notebooks which are forks of existing ones, with no changes:

image

I would propose a simple changes to make it much more likely that the information will be useful/interesting:

1. Add a boolean `user_titled` field to the notebook model. Defaults to `False`: https://github.com/iodide-project/iodide/blob/b45ecdc126a2db79cba30f3cd9eb5ceba2e1b6c4/server/notebooks/models.py#L18
2. When the notebook title changes, this field changes to True (if it wasn't already)
3. Forked notebooks do *not* inherit the setting.
4. The main page should filter by this attribute (being True of course) in order for a notebook to appear in the firehose: https://github.com/iodide-project/iodide/blob/b45ecdc126a2db79cba30f3cd9eb5ceba2e1b6c4/server/views.py#L21

Comments/feedback on this approach welcomed. Might be an interesting issue for @robhudson @rafrombrc. I think this can entirely be done inside `server/`. Be sure to add some tests if you do tackle this!

rafrombrc:
index page's notebook list. Fixes #710.

### Pull Request checklist


rafrombrc:
Oops I didn't know to run 'make lint'. Trying to figure out exactly what reformatting it's asking of me now.

server/tests/test_server_pages.py
```diff
@@ -7,7 +7,7 @@
 
 
 @pytest.mark.parametrize("logged_in", [True, False])
-def test_index_view(client, two_test_notebooks, fake_user, logged_in):
+def test_index_view(client, ten_test_notebooks, fake_user, logged_in):
```
wlach: I think we should probably just inline the ten_test_notebooks fixture, since there's a lot of logic about it that's pretty specific to this test (e.g. that only every other notebook has 10 revisions) and the behavior we want to test will almost certainly change, perhaps soon. The fixtures in conftest.py are meant to be broadly reusable across a large range of tests.

server/views.py
```diff
@@ -19,7 +19,10 @@ def get_user_info_dict(user):
 def index(request):
     user_info = get_user_info_dict(request.user)
     notebooks = (
-        Notebook.objects.annotate(latest_revision=Max("revisions__created"))
+        Notebook.objects.annotate(
+            latest_revision=Max("revisions__created"), num_revisions=Count("revisions")
+        )
+        .filter(num_revisions__gte=10)
```
wlach: Could you change this to a constant like `MIN_FIREHOSE_REVISIONS` in `server/settings.py`?

server/tests/conftest.py
```diff
@@ -62,6 +62,31 @@ def two_test_notebooks(fake_user):
...
+    """
+    notebooks = []
+    for i in range(10):
+        notebook = Notebook.objects.create(owner=fake_user, title="Fake notebook %s" % i)
+        NotebookRevision.objects.create(
+            notebook=notebook,
+            title="First revision of notebook %s" % i,
+            content="*fake notebook content %s*" % i,
+        )
+        # every other notebook gets 10 revisions so it will show up in the index page list
```
wlach: Let's key this off of the `MIN_FIREHOSE_REVISIONS` define in settings.py

wlach:
This looks great! Thanks Rob, merging.

PR #1774, opened 2019-04-29
fixes bug where notebooks / revisions cannot be deleted from server

### LINKED ISSUE #1773: "delete notebook" on server appears to be 404ing now
This was likely introduced by a recent change. I think it'll be an easy fix.

hamilton:
fixes #1773, which was likely introduced by the file API PR. This fixes a couple of small aspects of file uploading from the server, but is not meant to refactor the functionality in any substantial way.

### Pull Request checklist

### DIFF
--- a/src/server/components/delete-modal.jsx
+++ b/src/server/components/delete-modal.jsx
@@ -25,19 +25,16 @@ export default class DeleteModal extends React.Component {
     this.deleteObject = this.deleteObject.bind(this);
   }
 
-  deleteObject() {
+  async deleteObject() {
     // add delete here
-
-    this.props
-      .deleteFunction(this.props.elementID)
-      .then(() => {
-        this.props.onClose();
-        this.props.onDelete(this.props.elementID);
-      })
-      .catch(err => {
-        // FIXME: need a better reporting mechanism for delete failures
-        console.error(err);
-      });
+    try {
+      await this.props.deleteFunction(this.props.elementID);
+      this.props.onClose();
+      this.props.onDelete(this.props.elementID);
+    } catch (err) {
+      // FIXME: need a better reporting mechanism for delete failures
+      console.error(err);
+    }
   }
 
   render() {
--- a/src/server/components/delete-modal.jsx
+++ b/src/server/components/delete-modal.jsx
@@ -5,7 +5,6 @@ import ModalTitle from "../../shared/components/modal-title";
 import ModalContent from "../../shared/components/modal-content";
 import ModalCall from "../../shared/components/modal-call";
 import { TextButton } from "../../shared/components/buttons";
-import { deleteFileOnServer } from "../../shared/utils/file-operations";
 
 export default class DeleteModal extends React.Component {
   static propTypes = {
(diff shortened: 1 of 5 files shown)


src/server/components/delete-modal.jsx
```diff
@@ -15,7 +14,8 @@ export default class DeleteModal extends React.Component {
   deleteObject() {
     // add delete here
 
-    deleteFileOnServer(this.props.elementID)
+    this.props
+      .deleteFunction(this.props.elementID)
```
hamilton: by moving to `this.props.deleteFunction` we can test the actual deleting function independently, and we don't have to pass around boilerplate url strings

src/shared/utils/file-operations.js
```diff
@@ -95,23 +91,29 @@ export function saveFileToServer(
...
     if (r.status === 500) throw new Error(r.statusText);
     return r.json();
   });
 }
 
-export function deleteFileOnServer(fileID) {
-  return fetchWithCSRFTokenAndJSONContent(`/api/v1/files/${fileID}/`, {
-    method: "DELETE"
-  }).then(r => {
+export function makeDeleteRequest(url) {
```
hamilton: most delete requests have the exact same structure, with the difference being the URL in which we suggest to delete. This function should not ever be used – rather, I would suggest following the approach used in eg `deleteNotebookOnServer` below.

bcolloran: if it should not be used, should we not export it? or if we need to export it for testing, should we call it something like `makeDeleteRequest_private` and adopt a convention around not using thusly named vars outside of the file that defines them?

src/server/components/notebook-actions-menu.jsx
```diff
@@ -64,21 +64,19 @@ export default class NotebookActionsMenu extends React.Component {
...
-      .then(response => response.json())
-      .then(response => {
-        if (this.props.onUploadFile) this.props.onUploadFile(response);
-        this.hideUploadFileConfirmationModal();
-      });
+  async updateFile() {
+    const response = await uploadFile(
+      this.state.newFile,
+      this.state.oldFile.id
+    ).then(r => r.json());
```
bcolloran: here and above we could kill off another `then` with 

```
 const data = await response.json();
```

src/server/components/file-actions-menu.jsx
```diff
@@ -40,11 +41,11 @@ export default class FileActionsMenu extends React.Component {
           visible={this.state.deleteModalVisible}
           onClose={this.hideDeleteModal}
           title={`delete the file  "${this.props.filename}"?`}
+          deleteFunction={deleteFileOnServer}
```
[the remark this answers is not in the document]
hamilton: yep, I filed this before that merged but will rebase accordingly

src/shared/utils/file-operations.js
```diff
@@ -34,17 +34,13 @@ export function selectFileAndFormatMetadata(notebookID) {
   });
 }
 
-export function uploadFile(formData) {
-  return fetchWithCSRFToken("/api/v1/files/", {
+export function uploadFile(formData, fileID = undefined) {
+  // if fileID not defined, this will upload the file, or return an error if
+  // the filename in formData is already in the files table.
+  // if fileID is provided, will attempt to update the entry in the files table.
+  return fetchWithCSRFToken(`/api/v1/files/${fileID ? `${fileID}/` : ""}`, {
```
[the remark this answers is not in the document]
hamilton: that won't really work without the additional trailing slash so I'll just refactor this bit

hamilton:
@bcolloran thanks for the review. Updated + merging now.

PR #1775, opened 2019-04-30
adds timestamp to revisions page

### LINKED ISSUE #1568: Get unexpected results when clicking run after loading a document for the first time

### LINKED ISSUE #1658: the notebook revisions list on the server needs time stamps
HH:MM:SS

hamilton:
Fixes #1658. This is a stopgap for the time being.

I opted to make the timestamp a separate link that does the same thing as the date link above it, following the GH convention of redundant list item clickability.

### DIFF
--- a/src/server/pages/revisions-page.jsx
+++ b/src/server/pages/revisions-page.jsx
@@ -3,6 +3,7 @@ import PropTypes from "prop-types";
 import styled from "react-emotion";
 import Settings from "@material-ui/icons/Settings";
 import MoreHoriz from "@material-ui/icons/MoreHoriz";
+import format from "date-fns/format";
 
 import PageBody from "../components/page-body";
 import Header from "../components/header";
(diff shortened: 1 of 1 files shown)


hamilton:
Thanks for the review @wlach, edited to 1658 :)

Issue #1725, opened 2019-04-09
[Feature Request] intercept backspace button press outside editor


teonbrooks:
In Jupyter Notebook, if you try to hit backspace outside of the cell, it intercepts this action with a prompt to ask if you want to navigate from the page. The default behavior of backspace within a website is to navigate to previous page. This might be accidental in the context of a notebook so we can launch a modal to ask the user if they are sure they want to navigate away.

below is an example of a modal in the jupyter notebook:
<img width="1792" alt="Screen Shot 2019-01-14 at 11 24 16 AM" src="https://user-images.githubusercontent.com/1578674/55808765-b9d60400-5ab2-11e9-9f48-974dfb8daf31.png">

PR #1779, opened 2019-04-30
Intercept backspace when not focused to prevent backward page navigation

hamilton:
Partial solution to #1725. If the main scope is in-focus, this PR triggers the `beforeunload` modal on backspace (but not on page refresh, etc.). If the iframe is in-focus but no inputs are selected, we have to call `e.preventDefault` here since it seems there is more that must happen to pull up the same browser confirmation modal.

If we want to just trigger this for _all_ unloads (page refresh, for instance) then this woudl simplify the code a bit, but if we're interested in _just_ intercepting backspace, this would be one approach.

### DIFF
--- a/src/editor/initialization/keybindings.js
+++ b/src/editor/initialization/keybindings.js
@@ -1,8 +1,17 @@
 import Mousetrap from "mousetrap";
 import tasks from "../user-tasks/task-definitions";
-// for now, let's just keep the keybindings here.
 
 Mousetrap.prototype.stopCallback = () => false;
+
+let warnUser = false;
+
+const preventBacknav = e => {
+  if (e.target === document.body) {
+    warnUser = true;
+  } else {
+    warnUser = false;
+  }
+};
 
 export function initializeDefaultKeybindings() {
   Object.keys(tasks).forEach(t => {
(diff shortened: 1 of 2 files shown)


src/editor/initialization/keybindings.js
```diff
@@ -1,18 +1,36 @@
...
 }
 
+window.onbeforeunload = () => {
+  if (warnUser) {
+    warnUser = false;
+    return "Are you sure you want to leave?";
+  }
+  return undefined;
+};
+
```
hamilton: this approach allows us to only pop up the browser `beforeunload` modal if a user hits backspace. I could have this modal trigger _any_ time we try to unload from the client (that includes browser refresh) - thoughts here woudl be appreciated. Always triggering this on unload would simplify things quite a bit I think.

bcolloran: the simplification would be just not having the mousetrap.bind and the `preventBacknav`, or is there something else?

bcolloran: i think that simplifcation seems like the right thing actually -- let's just always warn before unload. let's make it hard for users to accidentally lose state.

bcolloran:
@hamilton plz circle back to this, let's wrap it up

bcolloran:
@Algentile -- this is somewhat related to the stuff you tackled in https://github.com/iodide-project/iodide/pull/2154/ -- any interest in taking a look and getting it over the line? @hamilton can help

Algentile:
Hey @bcolloran,

Yeah sure, I can check this out!

bcolloran:
closing in favor of #2252

Issue #1421, opened 2019-01-23
add a keybinding to move to next chunk without evaluating


hamilton:
@teon asked if there's a way to progress to another chunk without evaluating. We currently don't have a keybinding for this, but it seems like we probably should.

hamilton:
I had the idea of `alt+up` + `alt+down` this morning, which is unused, and is close to where the fingers are naturally.

hamilton:
Additionally, maybe this keybinding could also be used in the console to get to the next / previous evaluated selection without having to hold the arrow keys up or down. I've found this to be a rough edge, trying to get past a big code chunk in my history to get to a smaller one.

teonbrooks:
i like it. these are similar keybinding with jupyter (assuming alt and command are the same).

PR #1780, opened 2019-04-30
proposes alt+up and alt+down as a keybinding for moving between chunks

hamilton:
This PR implements the next / previous chunk keybinding suggested in #1421. For this PR, I implemented `alt+up` / `alt+down` since it doesn't seem to clash with any other keybindings to my knowledge, but I'm pretty indifferent to what we end up using, as long as it is easy on the hands.

### DIFF
--- a/src/editor/actions/__tests__/cursor-actions.test.js
+++ b/src/editor/actions/__tests__/cursor-actions.test.js
@@ -1,6 +1,10 @@
 import configureMockStore from "redux-mock-store";
 import thunk from "redux-thunk";
-import { moveCursorToNextChunk, updateEditorCursor } from "../actions";
+import {
+  moveCursorToNextChunk,
+  moveCursorToPreviousChunk,
+  updateEditorCursor
+} from "../actions";
 
 const middlewares = [thunk];
 const mockStore = configureMockStore(middlewares);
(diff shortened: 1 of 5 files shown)


src/editor/actions/actions.js
```diff
@@ -204,6 +204,14 @@ export function updateEditorCursor(line, col, forceUpdate = false) {
   return { type: "UPDATE_CURSOR", line, col, forceUpdate };
 }
 
+export function moveToNextChunk() {
```
bcolloran: there is already an action `moveCursorToNextChunk`

src/editor/reducers/notebook-reducer.js
```diff
@@ -329,6 +329,38 @@ const notebookReducer = (state = newNotebook(), action) => {
       });
     }
 
+    case "MOVE_TO_NEXT_CHUNK": {
```
bcolloran: you shouldn't need to touch the reducer to get this done -- take a look at the actions
`updateEditorCursor`and `moveCursorToNextChunk`

src/editor/actions/actions.js
hamilton: oh, thanks for the heads up. Of course there is! :|

src/editor/actions/__tests__/cursor-actions.test.js
```diff
@@ -176,3 +180,151 @@ describe("moveCursorToNextChunk dispatches correct actions", () => {
     expect(store.getActions()).toEqual(expectedActions);
   });
 });
+
+describe("moveCursorToPreviousChunk dispatches correct actions", () => {
```
hamilton: this test mirrors the `moveCursorToNextChunk` tests, updated appropriately.

hamilton:
@bcolloran thanks for the review. I've updated to use the action already there, and added another action - `moveCursorToPreviousChunk` + added tests. I've QA'd this locally against a handful of different notebooks too, just in case the tests aren't thorough enough.

If this all checks out, I'm happy to wait until we have more clarity on the right keybinding.

src/editor/actions/actions.js
```diff
@@ -218,16 +218,45 @@ export function moveCursorToNextChunk() {
       jsmdChunks,
       editorCursor
     } = getState();
+    const lastChunk = jsmdChunks[jsmdChunks.length - 1];
+    if (editorCursor.line === lastChunk.endLine) {
+      dispatch(updateEditorCursor(lastChunk.endLine, Infinity, true));
+      return;
+    }
```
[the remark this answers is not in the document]
hamilton: sure, will add a test case. I think it depends on what should happen if you're already on the last chunk. Should the cursor just go to the end of the entire doc? If so, CodeMirror reads Infinity and places at the end.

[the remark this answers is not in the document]
hamilton: I'll verify this bit though if this seems confusing. Before this PR the function was throwing an error at the edges of the document so the additional check was needed.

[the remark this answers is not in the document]
bcolloran: i think it should just go down to the front of the last line rather than the end of the line

bcolloran:
this is now getting kind of stale, and the tests will conflict with the monaco stuff (which uses 1-based indexing for buffer lines). i propose we close, since it will be quick to do again and it hasn't been touched in two months.

hamilton:
Yep feel free to close. It was meant as a proof of concept, not as a
mergeable feature. The work isn't really lost if its closed.

bcolloran:
(fwiw: i would have been happy to merge this! iirc, you consulted with a bunch of people and found this kbd shortcut to be the best option, right? we can pick it back up with monaco)

PR #1782, opened 2019-04-30
adds arrayBuffer info to fetch chunk documentation

### LINKED ISSUE #1686: arrayBuffer fetch type not documented in the fetch chunk section of the docs
~Array Buffers are pretty common response formats for `fetch` calls in JS. At the moment the closest we have is `blob` which requires some gymnastics to get into the right format. Critically, I think this would prevent us from exploiting some of the more interesting binary-based data representations we care about, including apache arrow tables.~

~This would make a really great first issue, since it just requires that we use `.arrayBuffer()`  on the `Response` object generated in the fetch actions and add the logic and syntax to handle that.~

(NB: this issue was closed with a recent PR. We need to add `arrayBuffer` to the docs) and then we can close this one. Updating the title to reflect this)

hamilton:
fixes #1686 – this was missing when added via the file API PR. 

UPDATE: This also adds syntax highlighting to `arrayBuffer` (thanks to @vadimkantorov for the catch there).

### DIFF
--- a/src/editor/components/codemirror-fetch-mode.js
+++ b/src/editor/components/codemirror-fetch-mode.js
@@ -6,10 +6,13 @@ CodeMirror.defineSimpleMode("fetch", {
     { regex: /\/\/.*/, token: "comment", sol: true },
     { regex: /[\s]+\/\/.*/, token: "comment" },
     {
-      regex: /(blob: |text: |json: )(\w+)(\s*=\s*)(\S+)/,
+      regex: /(blob: |text: |json: |arrayBuffer: )(\w+)(\s*=\s*)(\S+)/,
       token: ["fetch-type", "fetch-variable-declaration", null, "fetch-path"]
     },
     { regex: /(js: |css: )(\S+)/, token: ["fetch-type", "fetch-path"] },
-    { regex: /text: |blob: |json: |js: |css: /, token: "fetch-type" }
+    {
+      regex: /text: |blob: |json: |js: |css |arrayBuffer: /,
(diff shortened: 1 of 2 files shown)


hamilton:
Please hold off on any review here. I'm going to add syntax highlighting to the fetch chunk for `arrayBuffer` first.

src/editor/components/codemirror-fetch-mode.js
```diff
@@ -6,10 +6,13 @@ CodeMirror.defineSimpleMode("fetch", {
...
     { regex: /[\s]+\/\/.*/, token: "comment" },
     {
-      regex: /(blob: |text: |json: )(\w+)(\s*=\s*)(\S+)/,
+      regex: /(blob: |text: |json: |arrayBuffer: )(\w+)(\s*=\s*)(\S+)/,
       token: ["fetch-type", "fetch-variable-declaration", null, "fetch-path"]
     },
     { regex: /(js: |css: )(\S+)/, token: ["fetch-type", "fetch-path"] },
-    { regex: /text: |blob: |json: |js: |css: /, token: "fetch-type" }
+    {
+      regex: /text: |blob: |json: |js: |css |arrayBuffer: /,
```
bcolloran: missing colon after css?

hamilton: thanks for the catch :) went ahead and verified the rest of them work as well.

hamilton:
thanks for the review @bcolloran. Merging now.

PR #1783, opened 2019-04-30
moved EXPORT_NOTEBOOK logic into exportNotebook action creator

### LINKED ISSUE #995: make sure all reducers are pure functions, move side effects and impure data access to action creators
we're long overdue to go back and clean up some of the reducer code we wrote in our early days writing Iodide, and in particular before we had redux-thunk to help us manage asyncronous dispatch. One thing we can do is take a sweep through all of our reducers to make sure that they are pure functions, which will make them deterministic and more easily testable.

(and of course, we should also make sure we're not adding more such code before we get to this issue)

hamilton:
This fixes #995 by removing the `EXPORT_NOTEBOOK` case in the reducer and putting it into the action creator. This case was probably one of our very first ones, before we used thunks. At this point, the reducer itself should be pure-ish. We still have to look at `initialVariables` which are all the various variables from 3rd parties (mousetrap, etc.) in `window` that happens upon the page load.

We will probably refactor the export stuff down the line but this should keep the ball moving forward for now.

### DIFF
--- a/src/editor/actions/actions.js
+++ b/src/editor/actions/actions.js
@@ -26,6 +26,8 @@ import { fetchWithCSRFTokenAndJSONContent } from "../../shared/fetch-with-csrf-t
 import { jsmdParser } from "../jsmd-tools/jsmd-parser";
 import { getChunkContainingLine } from "../jsmd-tools/jsmd-selection";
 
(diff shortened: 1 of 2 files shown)


bcolloran:
code looks good, have not QAed. merge at you discretion. thanks hamilton! :-)

PR #1789, opened 2019-05-01
Consolidate code to access iodide server api in a module

wlach:
This is a little easier to reason about, and allows us to avoid
describing implementation details (e.g. the current use of cookies
to sign privileged requests) all over our code.


src/server/components/notebook-actions-menu.jsx
```diff
@@ -8,9 +8,9 @@ import DeleteModal from "./delete-modal";
 import UploadModal from "./upload-modal";
 import {
   selectFileAndFormatMetadata,
-  uploadFile,
-  deleteNotebookOnServer
```
wlach: I don't think a notebook deletion is, strictly speaking, a "file" operation (same with notebook revision deletion, in the file below). I wound up replacing it with a straightforward request here, since the semantics are more or less the same. If you think it's valuable, we could create a wrapper "notebook operations" module inside `utils/`, but am honestly not sure it's needed.

src/server/components/notebook-actions-menu.jsx
```diff
@@ -8,9 +8,9 @@ import DeleteModal from "./delete-modal";
...
+  uploadFile
 } from "../../shared/utils/file-operations";
+import { deleteNotebookRequest } from "../../shared/server-api/notebook";
 
 export default class NotebookActionsMenu extends React.Component {
```
hamilton: the `uploadFile` method already returns json, so the `NotebookActionMenu` throws an error on file upload

src/shared/server-api/api-request.js
```diff
@@ -0,0 +1,62 @@
+function getCookie(name) {
```
hamilton: @wlach I'm assuming we want devs to only use this to craft new API functions, not use these directly, correct?

@bcolloran you had mentioned before that we might want to put a note or something to make it very clear that using functions like these directly is a no-go. Any thoughts on how we might handle that here?

src/server/pages/revisions-page.jsx
```diff
@@ -136,12 +136,9 @@ export default class RevisionsPage extends React.Component {
 
   onDeleteRevision(revisionID) {
     if (this.state.revisions.length === 1) {
-      fetchWithCSRFTokenAndJSONContent(
-        `/api/v1/notebooks/${this.props.ownerInfo.notebookId}/`,
-        {
-          method: "DELETE"
-        }
-      ).then(this.onDeleteNotebook);
+      deleteNotebookRequest(this.props.ownerInfo.notebookId).then(
```
hamilton: we are making a point to make functions handling promises be async by default so they're a bit easier to read. Could you make this `async onDeleteRevision` & then deal with the promise w/ `await`?

src/shared/server-api/api-request.js
bcolloran: hmm well in this case the function is not exported, so it's "private" within the module. iirc the case that came up the other day was about a function that you were exporting, but which you said should not be used in other code?

since this is not exported, it's a bit different, but the risk remains that someone will forget about it and export it down the road. i think it would be good to mark things like this that we want to be treated as "private" _somehow_.

unlike in python, the js community seems to frown on the underscore prefix to indicate that a variable should be treated as private, but i'm not convinced by the arguments i've seen. i would be ok with adopting that convention (and i'm also open to other ideas). what do you guys think?

hamilton: I was referring to the other functions in this file which are exported, not this one. Seems to me the model we want is for devs to use this to create new APIs (eg the file or notebook apis will made in this PR) not directly

hamilton: I don't have a strong feeling one way or another on this. We coudl at the very least put a big comment at the top saying this file should not be used directly.

package.json
```diff
@@ -96,6 +96,10 @@
     "@babel/plugin-transform-modules-commonjs": "7.4.4",
     "@babel/preset-env": "7.4.4",
     "@babel/preset-react": "7.0.0",
+    "@storybook/addon-actions": "5.0.11",
```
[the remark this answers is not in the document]
bcolloran: looks like it was just resorted alphabetically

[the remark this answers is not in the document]
wlach: Yeah, I'll land a seperate commit changing just that in a few moments.

src/shared/server-api/api-request.js
wlach: Yeah, I don't think javascript really has any methods to enforce private methods. 

I think a simple comment like this at the top of the file should do the trick:

```js
// Please do not use any of these methods directly in your code! Instead, create
// methods over them, as is done elsewhere in this directory. This will make it easier to 
// switch to different authentication methods in the future.
```

src/server/components/notebook-actions-menu.jsx
wlach: Oops, right.

wlach:
Fixed review nits, ready for another look.

src/editor/actions/actions.js
```diff
@@ -256,80 +259,29 @@ export function getNotebookSaveRequestOptions(state, options = undefined) {
...
-      forkedFrom: options.forkedFrom
-    });
-    return saveNotebookRequest(
-      "/api/v1/notebooks/",
-      postRequestOptions,
-      dispatch
-    )
-      .then(json => {
+    return createNotebookRequest(state.title, state.jsmd, options)
+      .then(notebook => {
```
bcolloran: plz refactor this as async/await

src/editor/actions/actions.js
```diff
@@ -341,12 +293,7 @@ export function saveNotebookToServer(appMsg = true) {
...
     if (notebookInServer) {
       // Update Exisiting Notebook
-      return saveNotebookRequest(
-        `/api/v1/notebooks/${notebookId}/revisions/`,
-        getNotebookSaveRequestOptions(state),
-        dispatch,
-        appMsg
-      )
+      return updateNotebookRequest(notebookId, state.title, state.jsmd)
         .then(newRevision => {
```
bcolloran: plz refactor this as async/await

src/shared/utils/file-operations.js
```diff
@@ -63,12 +62,7 @@ export function selectAndUploadFile(notebookID, successCallback = () => {}) {
       })
     );
     formData.append("file", file);
-    fetchWithCSRFToken("/api/v1/files/", {
-      body: formData,
-      method: "POST"
-    })
-      .then(output => output.json())
-      .then(output => successCallback(output));
+    createFileRequest(formData).then(output => successCallback(output));
```
bcolloran: plz refactor this as async/await

src/editor/actions/actions.js
wlach: We shouldn't do this, as the function calling this is expecting a promise as a return value.

src/editor/actions/actions.js
wlach: Same

src/shared/utils/file-operations.js
wlach: In this case we can do the refactor.

src/editor/actions/actions.js
hamilton: async functions always return a Promise, and if there is no return value, I believe it will return `undefined`

wlach: to pass eslint I needed to make these functions return a dummy value of undefined. Not so happy with that, but it seems to work

src/editor/actions/actions.js
```diff
@@ -245,58 +245,69 @@ export function moveCursorToNextChunk() {
...
+        dispatch(
+          updateAppMessages({
+            message,
+            messageType: "NOTEBOOK_SAVED"
+          })
+        );
+      }
+      dispatch({ type: "NOTEBOOK_SAVED", newRevisionId: newRevision.id });
+    } catch (err) {
+      throw Error(err);
```
bcolloran: this pattern of
```
catch (err) {
      throw Error(err);
```
is redundant. if we throw anyway, there is no point in catching. this function can be cleaned up further

wlach: Hmm, looking at this again -- we should probably actually pop up an app message here, unless I'm missing something? I can open up a new pr for that.

src/editor/actions/actions.js
```diff
@@ -245,58 +245,69 @@ export function moveCursorToNextChunk() {
...
+      dispatch(
+        updateAppMessages({
+          message,
+          messageType: `NOTEBOOK_SAVED`
+        })
+      );
+      dispatch({ type: "ADD_NOTEBOOK_ID", id: notebook.id });
+      window.history.replaceState({}, "", `/notebooks/${notebook.id}`);
+      dispatch({ type: "NOTEBOOK_SAVED" });
+    } catch (err) {
```
bcolloran: this try/catch really only applies to `createNotebookRequest` i think -- if that fails, we should catch, update the message, and return from the function. the rest of those dispatches and stuff can happen outside of the try/catch

src/editor/actions/actions.js
bcolloran: perhaps. handling the error seems right, but if we don't handle it, we shouldn't rethrow.

also, i think that the try/catch only needs to wrap the actual server request

src/server/components/notebook-actions-menu.jsx
```diff
@@ -57,7 +57,7 @@ export default class NotebookActionsMenu extends React.Component {
       let filename;
       try {
         const metadata = JSON.parse(formData.get("metadata"));
-          filename = metadata.filename // eslint-disable-line
+        filename = metadata.filename; // eslint-disable-line
       } catch (err) {
         throw err;
       }
```
bcolloran: this ```
catch (err) {
        throw err;
      }
```
pattern again...

src/editor/actions/actions.js
wlach: We need the try/catch block around the whole thing, because the try block creates a block scope where the notebook is defined. I suppose we could get around this by defining the notebook object in a let statement, but I don't think it's the right approach-- the whole idea of the try block is to define a bunch of behaviour that should happen in sequence if an operation succeeds.

src/editor/actions/actions.js
wlach: See my comment above on this.

src/editor/actions/actions.js
bcolloran: ah, i see. i've always tried to only wrap the thing that i was worried might fail in a `try`, but this makes total sense too. thanks @wlach !

PR #1791, opened 2019-05-02
console/history perf tuneup

bcolloran:
various tweaks for around console/history performance
- moved where objects are passed into the value renderer tree -- was in a mapStateToProps, which always caused rerenders. Now a string id is passed in as props to ValueRendered, which looks up the value from IODIDE_EVALUATION_RESULTS (or `window` in the case of DeclaredVariables). 
- tweaks and clean up around history handling in general
    - fetch chunks return a string in "contents" of their history item rather than using a history value, which put an object in the global `IODIDE_EVALUATION_RESULTS`
    - deprecated `historyItem.value` in general, now just use `IODIDE_EVALUATION_RESULTS` with the `historyItem.historyId` as a key
    - broke history stuff into it's own schema -- might be a good first place to play with immer.js and to get it's own reducer
    - cleaned up the tree of functions used to create historyItems in the eval-frame (might do a little more on this)


src/editor/actions/console-message-actions.js
```diff
@@ -1,45 +1,37 @@
-import createHistoryItem from "../../shared/utils/create-history-item";
```
bcolloran: got rid of `createHistoryItem` in favor of doing it's job of making sure random historyIds are always added here and at the similar point in the eval-frame

src/eval-frame/actions/actions.js
```diff
@@ -87,19 +89,20 @@ export function runCodeWithLanguage(language, code) {
 export async function evaluateCode(code, language, chunkId, evalId) {
   try {
     MOST_RECENT_CHUNK_ID.set(chunkId);
-    const output = await runCodeWithLanguage(language, code);
-    addConsoleEntryInEditor({
-      historyType: "CONSOLE_OUTPUT",
-      value: output
+    const value = await runCodeWithLanguage(language, code);
+
+    const historyId = addConsoleEntryInEditor({
```
bcolloran: addConsoleEntryInEditor now returns a historyId

src/eval-frame/actions/actions.js
```diff
@@ -87,19 +89,20 @@ export function runCodeWithLanguage(language, code) {
...
+      historyType: "CONSOLE_OUTPUT"
     });
+    IODIDE_EVALUATION_RESULTS[historyId] = value;
```
bcolloran: adding to IODIDE_EVALUATION_RESULTS now only happens in the case of true evaluation results, so we can do it here rather than when we add to things to the history

src/eval-frame/actions/fetch-cell-eval-actions.js
```diff
@@ -115,54 +110,49 @@ export async function handleFetch(fetchInfo) {
...
-        historyId: outputHistoryId,
-        level: "ERROR"
-      })
-    );
+    addConsoleEntryInEditor({
+      historyType: "FETCH_CELL_INFO",
+      content: syntaxErrors
+        .map(fetchProgressInitialStrings)
+        .map(t => t.text)
+        .join(""),
```
bcolloran: just use `historyItem.content` strings rather than array of objects. strings behave better with shallow compare and serialize easier everywhere, and this was just displayed in a <pre> anyway

src/eval-frame/components/panes/declared-variable.jsx
```diff
@@ -5,15 +5,14 @@ import ValueRenderer from "../reps/value-renderer";
...
-    value: PropTypes.any, // eslint-disable-line react/forbid-prop-types
     varName: PropTypes.string
   };
   render() {
     return (
       <div className="declared-variable">
         <div className="declared-variable-name">{this.props.varName} = </div>
         <div className="declared-variable-value">
-          <ValueRenderer valueToRender={this.props.value} />
+          <ValueRenderer windowValue valueKey={this.props.varName} />
```
bcolloran: `windowValue` flag tells the ValueRendered to look on the window object for the value it wants

src/eval-frame/components/panes/history-item.jsx
```diff
@@ -45,16 +50,14 @@ export class HistoryItemUnconnected extends React.Component {
...
           <ConsoleMessage level={this.props.level || "OUTPUT"}>
-            <ValueRenderer valueToRender={this.props.valueToRender} />
+            <ValueRenderer valueKey={this.props.historyId} />
           </ConsoleMessage>
         );
       }
       case "FETCH_CELL_INFO": {
         return (
           <ConsoleMessage level={this.props.level || "OUTPUT"}>
-            <PreformattedTextItemsHandler
```
bcolloran: PreformattedTextItemsHandler was overkill. we can just render the fetch cell info as a single <pre> string, which has advantages in many places

src/eval-frame/components/reps/__tests__/value-renderer.test.js
```diff
@@ -9,6 +9,22 @@ import ErrorRenderer from "../error-handler";
...
+  { type: "string", val: "foo" },
+  { type: "null", val: null },
+  { type: "undefined", val: undefined },
+  { type: "function", val: () => 123 }
+];
+
+simpleTestCases.forEach((testCase, i) => {
+  IODIDE_EVALUATION_RESULTS[`result_${i}`] = testCase.val;
+});
+
```
bcolloran: updating tests now that ValueRenderer takes a valueKey prop that it uses to look up values in IODIDE_EVALUATION_RESULTS

src/eval-frame/components/reps/data-table-rep.jsx
```diff
@@ -64,7 +64,7 @@ class CellRenderer extends React.Component {
             : "none"
         }}
       >
-        <TinyRep serializedObj={serializeForTinyRep(this.props.value)} />
+        <TinyRep {...serializeForTinyRep(this.props.value)} />
```
bcolloran: spread props out individually rather than passing down a big object with children. facilitates shallow comparison of props

src/eval-frame/components/reps/tiny-rep.jsx
```diff
@@ -164,19 +164,19 @@ const typeToTinyRepMapping = {
...
-      size: PropTypes.number,
-      stringValue: PropTypes.string.isRequired,
-      isTruncated: PropTypes.bool.isRequired
-    })
+    objType: PropTypes.string.isRequired,
+    /* eslint-disable react/no-unused-prop-types */
+    isTruncated: PropTypes.bool.isRequired,
+    objClass: PropTypes.string.isRequired,
+    size: PropTypes.number,
+    stringValue: PropTypes.string.isRequired
```
bcolloran: as above: spread props out individually rather than passing down a big object with children. facilitates shallow comparison of props.

(lesson: in general, we'll probably want to avoid PropTypes.shape in favor of flat props with simple types)

src/eval-frame/global-state-extras.js
```diff
@@ -0,0 +1 @@
+export const IODIDE_EVALUATION_RESULTS = {};
```
bcolloran: IODIDE_EVALUATION_RESULTS is basically a global variable that is used to manage a part of state that we cannot put in the redux store. i thought we should put that (and anything else like it) into it's own file rather than having a stateful thing defined in some random place

src/shared/state-selectors/history-selectors.js
```diff
@@ -0,0 +1,25 @@
+import { createSelectorCreator, defaultMemoize } from "reselect";
```
hamilton: out of curiosity, is there a reason you went with `reselect` and not the native [`areStatesEqual`](https://react-redux.js.org/api/connect#arestatesequal-next-object-prev-object-boolean) in redux? I can see the value in reselect more broadly, but we're only using it in one place in the entire app here, and `areStatesEqual` feels like it can do the same thing without the added dependency. I don't think Redux performance is _across the board_ a problem (and indeed this PR greatly improves eval performance!), and places like history (and chunks) are the two heaviest part of the stores.

src/eval-frame/components/reps/tiny-rep.jsx
```diff
@@ -164,19 +164,19 @@ const typeToTinyRepMapping = {
...
+    /* eslint-enable react/no-unused-prop-types */
   };
   render() {
-    const { objType } = this.props.serializedObj;
+    const { objType } = this.props;
     const repType = handledTypes.includes(objType) ? objType : "Object";
-    return typeToTinyRepMapping[repType](this.props.serializedObj);
+    return typeToTinyRepMapping[repType](this.props);
```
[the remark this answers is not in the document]
hamilton: Generally, not returning JSX when the output value is a React COmponent makes it hard to reason about what is happening.

src/editor/actions/console-message-actions.js
```diff
@@ -1,45 +1,37 @@
...
+  });
+
+export const addInputToConsole = (evalText, evalType) =>
+  addToConsoleHistory({
+    historyType: "CONSOLE_INPUT",
+    content: evalText,
+    language: evalType
+  });
+
+export const pluginParseError = errorMessage =>
```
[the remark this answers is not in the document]
bcolloran: yep np

src/eval-frame/components/reps/data-table-rep.jsx
```diff
@@ -47,7 +47,7 @@ CellDetails.propTypes = {
   focusedCol: PropTypes.string
 };
 
-class CellRenderer extends React.Component {
+class CellRenderer extends React.PureComponent {
```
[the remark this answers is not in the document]
bcolloran: i think it does help because IIUC purecomponent does a shallow equality check, and as long as the object passed in is identical by reference, the component will not rerender. but in any case: this PR is just phase 1 of rep cleanup; more to come that will optimize this further, so I'm going to hold off on that refactor as part of this PR

src/eval-frame/components/reps/tiny-rep.jsx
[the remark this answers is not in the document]
bcolloran: refactoring in follow up PR

src/shared/state-selectors/history-selectors.js
bcolloran: i looked into this, and not including reselect at this point is fine with me. nice tip :-)

src/eval-frame/components/reps/data-table-rep.jsx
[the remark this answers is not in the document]
hamilton: sounds good

PR #1796, opened 2019-05-02
Do not block compilation for prettier warnings

openjck:
It can be a little frustrating when prettier blocks compilation during
development. This often happens to me while tinkering with props in JSX.
Adding or removing a prop can put the line over or under the 80
character limit, making prettier require whitespace changes. Until those
stylistic changes are made, the new code cannot be compiled or tested in
the browser.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,9 @@ module.exports = {
   plugins: ["jest"],
   extends: ["airbnb", "plugin:prettier/recommended", "prettier/react"],
   rules: {
+    // Make prettier emit warnings rather than errors
+    "prettier/prettier": ["warn"],
+
     // Restricting for..of seems pretty controversial, let's disable that.
     // See https://github.com/airbnb/javascript/issues/1271
     "no-restricted-syntax": ["off"],
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -47,7 +47,9 @@ const plugins = [];
 module.exports = env => {
   env = env || ""; // eslint-disable-line no-param-reassign
 
-  if (!env.startsWith("dev")) {
+  const isDev = env.startsWith("dev");
+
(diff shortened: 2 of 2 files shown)


openjck:
One issue here is that eslint warnings are printed above these warnings, which I have made a habit of ignoring:



If we don't see the eslint warnings, we're unlikely to fix them. Can the above warnings be easily addressed or moved above the eslint warnings?

bcolloran:
hey @openjck, i think the way i'd do this to have the webpack config check whether you're running in dev or prod mod, and to set those options to warn for dev but error for prod. we want to make sure that CI fails on lint problems.

bcolloran:
(oh also-- i think @wlach is going to address those warnings as part of the work he is doing this week)

openjck:
That makes sense. Updated!

wlach:
@openjck you may want to consider having prettier reformat your code on save, I've been doing this with atom and it works great.

bcolloran:
+1 on having your editor do this on save. i'm using that in vscode and it's pretty seamless

openjck:
Great tip. I've set up Vim to do this. :smiley:

wlach:
I'm kind of on the fence about this PR. On the one hand, I get this is frustrating -- on the other, I feel like this might be good incentive for people to configure their editors to work the right way (and not accidentally commit code that doesn't pass lint).

@openjck what do you think? Maybe we should just update the docs.

openjck:
That works for me. I'll open a PR to update the documentation. I might also add a `format` NPM script for those who aren't using editor plugins, unless anyone has any objections to that. :smiley:

openjck:
Actually, there's already a note about this in the docs. So I may just open a PR for a format script.

`It is recommended to configure your editor to use the above linting tools [including Prettier].`

PR #1802, opened 2019-05-03
Clean up some issues around error handling in some server-api code

wlach:
* Show a proper error message if we fail to update a notebook, rather than
  uselessly rethrowing the exception
* Some other misc. cleanup


wlach:
I'm actually going to close this, as I'm realizing some parts of this are not quite right. Will fix as part of a larger rework of autosave error handling in another PR.

PR #1839, opened 2019-05-14
fix argument order in save example

bcolloran:
@hamilton pl double check me on this and merge if correct. thanks!


hamilton:
lgtm - thanks for the catch ~~~~

Issue #826, opened 2018-08-30
favicon?


mdboom:
Should we add a favicon to the app and website?

bcolloran:
yes we definitely should, but ATM we don't have a real logo, or even a lame one suitable for use as a favicon see #267 , #167

PR #1856, opened 2019-05-22
Add favicon to iodide

teonbrooks:


teonbrooks:
I didn't know where to place the assets so currently they are in the editor folder.

this closes #826

teonbrooks:
@wlach, i'm not following the relationship of the template parts in django and the part that is coming from the editor. I just tried to play around with it but I wasn't successful at getting the favicon this way. mind walking me through what's going on here?

wlach:
> @wlach, i'm not following the relationship of the template parts in django and the part that is coming from the editor. I just tried to play around with it but I wasn't successful at getting the favicon this way. mind walking me through what's going on here?

I can try. :)

Basically when you access the iodide from the production server (or a local dev instance), it's going through a web server run by django. Django has a concept of something called "views", which are basically python functions which return a bunch of content (usually either html or json). There is an internal process that maps url requests to these functions. For example, the view for the front page is here:

https://github.com/iodide-project/iodide/blob/b1832fa22af266241c0e8b4b2ce312bd6cabd5a8/server/views.py#L19

Templates are only intended to be used by these views-- they can't be served on their own.

In addition to these views, Django is also configured to allowing serving static files via preset directories which is what we use to serve the webpack js that is compiled on the client side. You can see the configuration for that here:

https://github.com/iodide-project/iodide/blob/b1832fa22af266241c0e8b4b2ce312bd6cabd5a8/server/settings.py#L201

Actually on investigation I think we're already set up for that, so just create a directory called
`server/static/images/` in iodide, dump the files there, and then reference these static assets using the methodology here inside the existing templates:

http://whitenoise.evans.io/en/stable/django.html#make-sure-staticfiles-is-configured-correctly

lmk if you have any more questions!

teonbrooks:
thanks @wlach for the explanation. I followed those steps but I am getting a 404 error in the console log saying the favicon.ico can't be found. it seems like everything is point the right locations. 

just looked at the build error and it seems that the files are missing from the manifest. any suggestions?

wlach:
@teonbrooks this was actually super close, we just needed to add a STATICFILES_STORAGE definition to reference the files you added, as described here:

https://docs.djangoproject.com/en/2.2/howto/static-files/#configuring-static-files

The fix is a oneliner like this in settings.py:

```py
STATICFILES_DIRS = (os.path.join(BASE_DIR, "server", "static"),)
```

server/templates/base.html
```diff
@@ -2,7 +2,11 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    {% block head %}{% endblock %}
```
wlach: I don't think we want to remove these block directives, they are necessary to be able to vary the inheritance between the editor and other server-side pages (the editor needs different css directives e.g.)

src/editor/html-template.js
```diff
@@ -1,6 +1,7 @@
 const template = `<!DOCTYPE html>
 <html>
 <head>
+<meta name="theme-color" content="#ffffff">
```
wlach: I'm guessing this was accidental?

wlach:
Looks like the tests are failing, I think because we're failing to collect static assets correctly:

https://stackoverflow.com/a/51580328

We do actually call collectstatic, but it's possible the `-c` option we're passing (which clears the static files dir) is problematic:

```
Step 14/16 : RUN DEBUG=False SECRET_KEY=foo ./manage.py collectstatic --noinput -c
 ---> Running in 5a849678a868
Deleting 'assets/favicon.ico'
Deleting 'assets/favicon-32x32.png'
Deleting 'assets/safari-pinned-tab.svg'
Deleting 'assets/favicon-16x16.png'
Deleting 'assets/apple-touch-icon.png'
```
https://circleci.com/gh/iodide-project/iodide/9110?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

@teon Can you try removing the `-c` option from this line and seeing if it helps?

https://github.com/iodide-project/iodide/blob/80c1a741c79fb611d50d9e56a47c869a49251875/Dockerfile#L37

teonbrooks:
@wlach is there an easy way to run the pytest locally without running docker? 

also, remove the flag didn't change anything

wlach:
> @wlach is there an easy way to run the pytest locally without running docker?

Not really, the tests assume the presence of a database and other things. For local testing I usually bring the environment up (`make up`) and then get a shell (`make shell`) and run `py.test server/tests` from there

> also, removing the flag didn't change anything

Grump. :( Could you try re-running `./manage.py collectstatic` with the instructions above before the unit tests?

teonbrooks:
> Grump. :( Could you try re-running `./manage.py collectstatic` with the instructions above before the unit tests?

running this above requires being in the docker environment, correct? 
I tried to run it locally and I tried to set up docker locally and both failed spectacularly, and couldn't find any docs on how our docker environment is setup to run.

wlach:
> > Grump. :( Could you try re-running `./manage.py collectstatic` with the instructions above before the unit tests?
> 
> running this above requires being in the docker environment, correct?
> I tried to run it locally and I tried to set up docker locally and both failed spectacularly, and couldn't find any docs on how our docker environment is setup to run.

The instructions under "server mode" here should work: https://iodide-project.github.io/docs/local-dev/#runningbuilding

teonbrooks:
thanks @wlach, I somehow missed it that step after getting docker installed and was frustrated with myself with this PR trying to fix a seemingly simple thing.  I got the favicon appearing locally. let's see what happens with the tests.

server/settings.py
```diff
@@ -197,9 +197,10 @@
 
 
 # Files in this directory will be served by WhiteNoise at the site root.
-WHITENOISE_ROOT = os.path.join(ROOT, "build")
-STATIC_ROOT = os.path.join(ROOT, "static")
+WHITENOISE_ROOT = os.path.join(BASE_DIR, "build")
+STATIC_ROOT = os.path.join(BASE_DIR, "server")
```
wlach: I think this line should be reverted, we want the static assets to be in `static/`, not in `server/` (which is where the application files live)

server/settings.py
```diff
@@ -197,9 +197,10 @@
...
 STATIC_URL = EVAL_FRAME_ORIGIN
+STATICFILES_DIRS = (os.path.join(STATIC_ROOT, "static"),)
```
wlach: Then change this to STATICFILES_DIRS = (os.path.join(BASE_DIR, "server", "static"),)

teonbrooks: when I do this, I get this error: 
```
server_1  | ?: (staticfiles.E002) The STATICFILES_DIRS setting should not contain the STATIC_ROOT setting.
```

server/settings.py
teonbrooks: ooh, so this is a folder at the root level, not in the server folder. i think i had been confusing these the whole time

wlach:
@teonbrooks everything in your PR looks good, I think the problem is due to the fact that we're trying to test in non-debug mode. One of the solutions here will probably fix things:

https://stackoverflow.com/questions/44160666/valueerror-missing-staticfiles-manifest-entry-for-favicon-ico

This is all a bit of a mess, sorry you ran into it. If you like I can try to fix things tomorrow.

teonbrooks:
so I commented out the `STATICFILES_STORAGE` line, and now the test passes. i guess this would mean the static files would no longer be compressed. is this only being used for the favicon right now? is compression something we would want it in the future?

teonbrooks:
> [](https://stackoverflow.com/questions/44160666/valueerror-missing-staticfiles-manifest-entry-for-favicon-ico)
> 
> This is all a bit of a mess, sorry you ran into it. If you like I can try to fix things tomorrow.

feel free to take over this if you would like, i'm not exactly sure what would be the best solution

wlach:
> so I commented out the `STATICFILES_STORAGE` line, and now the test passes. i guess this would mean the static files would no longer be compressed. is this only being used for the favicon right now? is compression something we would want it in the future?

I am pretty sure it is used for compressing the entire frontend (the .js/.css files in build/), so I think it's something we need.

Just adding `IODIDE_SERVER_DEBUG=1` to the end of this line may fix it: I think that will change whitenoise's behaviour and the tests should work again.

https://github.com/iodide-project/iodide/blob/5b910301f181de7613102e233b87d9521d86e3a7/Makefile#L27

wlach:
Investigating... this problem is frustratingly hard to reproduce locally

wlach:
I think #1910 should fix this, rebasing the PR on top of that to confirm

teonbrooks:
> I think #1910 should fix this, rebasing the PR on top of that to confirm

sweet, that seems to have done it, @wlach :)

PR #1877, opened 2019-05-29
Support and use JWT-based authentication to access the API via the client

wlach:
### Pull Request checklist

### DIFF
--- a/src/shared/server-api/__test__/api-request.test.js
+++ b/src/shared/server-api/__test__/api-request.test.js
@@ -1,0 +1,137 @@
+import { signedAPIRequestWithJSONContent } from "../api-request";
+
+const apiRequest = require("../api-request");
+
+describe("signedAPIRequest", () => {
+  beforeEach(() => {
+    fetch.resetMocks();
+
+    // jest's cookie mock is (currently) incomplete so we need to override it
+    apiRequest.setCookie = jest.fn((a, b) => {
+      document.cookie = `${a}=${b}`;
+    });
+
+    // reset cookie state between tests
+    document.cookie = "jwt-access-token=;expires=1 Jan 1970 00:00:00 GMT;";
+    document.cookie = "jwt-refresh-token=;expires=1 Jan 1970 00:00:00 GMT;";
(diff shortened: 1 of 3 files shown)


wlach:
This seems to be basically working. I have tests for the server side, need to write something up for the client (and fix whatever broke).

jezdez:
@wlach Anything I can help here?

server/jwt/views.py
```diff
@@ -0,0 +1,21 @@
...
+    rest_frame_simplejwt by default requires a username/password post blob for
+    its token-obtaining endpoint (which we usually aren't in a position to provide),
+    so we provide this simple method which does the equivalent, using Django's
+    existing session-based authentication to identify and validate the user.
+    """
+    if not request.user.is_authenticated:
+        raise PermissionDenied
+
+    refresh = RefreshToken.for_user(request.user)
+    return JsonResponse(data={"refresh": str(refresh), "access": str(refresh.access_token)})
```
jezdez: I understand why we'd not want to use the built-in view, but I think it would be better to implement an own `GenericAPIView` following the general API that `rest_framework_simplejwt` follows. The advantage being that we could use DRF's session auth backend just locally for that view.

wlach: This sounds like a good idea, I will give it a go

bcolloran:
i did the quickest of passes, and the client side code looked ok to me -- flagging @hamilton for review tho since it is client code that talks to the server (not my specialty)

wlach:
@bcolloran / @hamilton it may be better to hold off on further review of the client side until I've finished writing the jest tests as some things may change.

server/jwt/views.py
wlach: Implemented, we'll continue discussion of the server pieces in #1906

wlach:
@hamilton this should be ready for review now, just ignore the python parts (that's being reviewed seperately by @jezdez in #1906)

wlach:
#1906 landed so rebasing on master, for easier reviewing! Should be just the js/client parts now.

Issue #1748, opened 2019-04-17
A value renderer encountered an error.


renedlog:
### What I Did
code sample:  https://alpha.iodide.io/notebooks/300/

Steps:
Run Full Notebook -> Workspace -> Expand pyodide object -> "A value renderer encountered an error."

### What I Expected

Do not allow expansion or do render an info why it cannot expand (catch error).

### What Happened 

Error msg

### (if applicable) notebook that reproduces bug 
https://alpha.iodide.io/notebooks/300/

bcolloran:
superseded by #2254

Issue #1787, opened 2019-05-01
Displaying large typed arrays to the console hangs


mdboom:
Unlike regular Arrays, displaying large typed arrays to the console seems to hang Iodide definitely.

### What I Did

`new UInt8Array(40000000)`

`new Array(40000000)` works and is correctly truncated.

bcolloran:
fixed with #1884

PR #1884, opened 2019-05-31
Rep refresh

bcolloran:
this is the most WIP PR ever, but feel free to play around with it a bit. the code is not ready for review, there is huge cleanup and more testing needed, but we have a pretty good story pretty much all the JS types, including folding of long arrays and strings... and even folding of property lists for objects that have a zillion (oh, and `window` doesn't crash now, so you can see what i mean there). I think this puts us on better footing for more robustness and comprehensiveness overall.

among other things, there is still some work to be done to get this to integrate with the tabular viewer so that is known to be broken.

i recommend looking around in the storybook (`npm run storybook`) at the "expandable rep" bullet. that will quickly let you see the behavior in a ton of cases.

please let me know if you notice anything that looks wrong. also, if you encounter any cases that cause weird behavior that are not in the storybook cases, please paste code here so i can add a storybook case for testing.

(@mdboom fyi -- i can't recall what kinds of issues were coming up with pyodide. if you spot any please flag them)

### Pull Request checklist

- [ ] **Documentation**: not documented, b/c this needs to Just Work without explanation and documentation
- [ ] **Tests**: some parts not tested fully at this point, b/c per discussion the thorough Storybook QA is satisfactory given our resourcing


mdboom:
I played with the storybook examples, and they all look like a major improvement.  Don't really have much to add there.

For Python stuff, displaying a Python object that doesn't have an obvious Javascript conversion (a proxy object) displays the Python `__repr__`, which is perfect, as far as I'm concerned.

It doesn't seem to break the display of Pandas data frames (where Pandas provides its own HTML through a custom output handler).  matplotlib likewise works fine.

One weirdness, which isn't actually a regression in this PR, is that Numpy arrays are weird.

image

Numpy arrays get converted to `Arrays` of `TypedArrays` when converted from Python to Javascript, so that the entire set of data doesn't need to be copied.  (Way earlier version of Pyodide just built Arrays of Arrays which was expensive, but the reps did kinda the right thing with them).  The long term plan is to have a proper n-dimensional array data type in Javascript and use that (and then the reps would probably have to have a special case to do the right thing with them).  I don't know if that's worth fixing now, but just wanted to point out the issue.

bcolloran:
thanks for looking into the pyodide cases @mdboom!

I think that for the case of arrays of arrays, we could look at once again offering a matrix view, which could reuse some of the primitives from this code.

bcolloran:
@hamilton per our plan last week, and even though i feel very guilty inflicting this on you without being able to take the time i'd like to write up thorough notes for review, i think i'd like to go ahead and hand this to you for QA in hopes of getting it landed by Whistler.

hamilton:
No prob! Looking forward to reading through.

bcolloran:
fixes #1748 #1787

bcolloran:
@hamilton going to merge this this week i think. it's now been a while, so i'm thinking that based on our past discussions you decided not to review further?

PR #1906, opened 2019-06-05
Server-side implementation of JWT-based authentication (fixes #1755)

### LINKED ISSUE #1755: Support accessing /api/ endpoints using jwt authentication
To support the internal server use case, @jezdez had the idea of using https://github.com/davesque/django-rest-framework-simplejwt to allow users to get a token which allows accessing the iodide api's with a known expiry (unlike the django session authentication, which randomly times out due to the vagaries of openidc).

The server-side implementation of this is likely to be very similar to the authtoken authentication I added in a4a8f58ec77455ca9cd8ce9bdfa3ab8173cf6b67. Obviously the client side will also need to be reworked to use it, but I think we can get the server parts working standalone and test it, and then work on the client parts as a seperate step.

@rafrombrc @robhudson (or others) -- any interest in giving this a shot? I will be away this coming week but can review when I return.

wlach:
### Pull Request checklist


wlach:
This is just the python/server parts of #1877 with @jezdez's suggestion in https://github.com/iodide-project/iodide/pull/1877#discussion_r290361719 and a few other small fixes

PR #1925, opened 2019-06-11
Fix isort errors in two server/ files

openjck:


openjck:
Sounds like I have an outdated version of isort.

PR #1926, opened 2019-06-11
Add ability to restrict API to authenticated only

### LINKED ISSUE #1879: Allow completely restricting /api/ endpoints to those with authorization
For the internal mozilla use case, we would like to be able to restrict *any* access to /api/ to those clients with either a token or a valid session. Currently we rely on openidc auth for these endpoints but this approach doesn't work well with the ajax approach we use inside the notebook.

For this to be useful, we need #1755 first.

/cc @robotblake

jezdez:
Fix #1879.

### Pull Request checklist


jezdez:
Tests upcoming... 😬

server/tests/test_jwt_authentication.py
```diff
@@ -1,6 +1,3 @@
-from datetime import timedelta
-from time import sleep
```
wlach: We should be running flake8 to detect these errors

jezdez:
> There seems to be a black linter error. Aside from that, looks good to me!

Yeah, there is https://github.com/timothycrosley/isort/issues/694 that documents a incompatibility of isort and black in their default configuration. Luckily there is a [workaround](https://black.readthedocs.io/en/stable/the_black_code_style.html?highlight=isort#how-black-wraps-lines).

server/tests/test_jwt_authentication.py
jezdez: Done in #1931.

PR #1933, opened 2019-06-12
pull in a smaller gravatar icon

### LINKED ISSUE #1711: Should use smaller version of user icons in main view
On the front page, we currently get github and gravatar images using their native sizes, which are rather large (80x80 in the case of gravatar, up to 500x500 in the case of github)

We get the URL here:

https://github.com/iodide-project/iodide/blob/b45ecdc126a2db79cba30f3cd9eb5ceba2e1b6c4/server/views.py#L21

(the part of the query where we fetch the avatar URL is `owner__avatar`)

And render it in react here:

https://github.com/iodide-project/iodide/blob/b45ecdc126a2db79cba30f3cd9eb5ceba2e1b6c4/src/server/components/user-name.jsx#L24

This results in a large amount of unnecessary bandwidth being used and the pages loading more slowly.

For gravatar, we should be able to add `s=30` to the URL to get them at the right size: https://en.gravatar.com/site/implement/images/

Github seems to support the same, e.g. https://avatars3.githubusercontent.com/u/20569?v=4&s=30

To do this properly, I'd recommend a small helper function on the client that takes a URL and adds either a `?s=30` or a `&s=30` to it. I'm a bit torn on whether to do this on the client or server side, but am leaning towards doing it on the client. @robhudson @rafrombrc this would be another good bug to tackle.

teonbrooks:
this reduces the bandwidth of pulling in a gravatar.
closes #1711

### Pull Request checklist

### DIFF
--- a/src/server/components/user-name.jsx
+++ b/src/server/components/user-name.jsx
@@ -1,6 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import styled from "react-emotion";
+import THEME from "../../shared/theme";
 
 const UserNameContainer = styled("a")`
   display: flex;
@@ -21,7 +22,9 @@ const UserNameContainer = styled("a")`
   }
 `;
 
-const Avatar = ({ src }) => <img src={`${src}&s=30`} alt={src} />;
+const Avatar = ({ src }) => (
+  <img src={`${src}&s=${THEME.defaultUserIconSize}`} alt={src} />
+);
 
 Avatar.propTypes = {
   src: PropTypes.string.isRequired
--- a/src/server/components/user-name.jsx
+++ b/src/server/components/user-name.jsx
@@ -21,7 +21,7 @@ const UserNameContainer = styled("a")`
   }
 `;
 
-const Avatar = ({ src }) => <img src={src} alt={src} />;
+const Avatar = ({ src }) => <img src={`${src}&s=30`} alt={src} />;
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -64,6 +64,7 @@ typings/
 
 # Python cache directory
 .cache
+*.pyc
```
wlach: This should go in a separate PR, though I am not sure if it is needed? Not seeing any .pyc files in my iodide checkout

src/server/components/user-name.jsx
```diff
@@ -21,7 +21,7 @@ const UserNameContainer = styled("a")`
   }
 `;
 
-const Avatar = ({ src }) => <img src={src} alt={src} />;
+const Avatar = ({ src }) => <img src={`${src}&s=30`} alt={src} />;
```
wlach: It might make sense to define the `30` in `src/shared/theme` as something like `THEME.defaultUserIconSize`. @hamilton might have opinions/suggestions on this.

teonbrooks:
no test was included since this was a styling modification. let me know if you would like me to add a test

teonbrooks:
@wlach, is this in line with what you were thinking?

wlach:
Looks good! Testing locally, it seems like using a size of 40 gives better results (30 looks a little blurred), so let's go with that.

PR #1938, opened 2019-06-14
Fix failing file uploads

openjck:


openjck:
There are some changes in *file-request-actions.js*, but all they do is revert the file to where it was before my initial file upload PR.

hamilton:
Thanks @openjck – filed another branch that has the full fixes.

PR #1944, opened 2019-06-14
Remove requirement for email header in openidc middleware

wlach:
On our internal deployment, this is no longer required for some endpoints.
While we're here, remove the "openidc auth whitelist" settings, in iodide
this "whitelist" is really configured at the web server level.

This whole authentication scheme should actually should be documented soon (it's on my todo list), but I don't think it needs to be part of this commit.

### Pull Request checklist

- [N/A] **Changelog**: This PR updates the [changelog](../CHANGELOG.md) with any user-visible changes.

### DIFF
--- a/server/openidc/middleware.py
+++ b/server/openidc/middleware.py
@@ -11,8 +11,9 @@ class OpenIDCAuthMiddleware(object):
     request. This header will be populated by nginx configured to authenticate
     with OpenIDC.
 
-    We will automatically create a user object and attach it to the
-    experimenters group.
+    If no header is set, we will silently bypass this authentication layer (relying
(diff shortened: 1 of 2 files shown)


robotblake:
Wouldn't this still allow someone to create a user by sending a request with the correct header?

wlach:
> Wouldn't this still allow someone to create a user by sending a request with the correct header?

Hmm, I think you might be right -- unless we go ahead and sanitize the headers at the nginx level (or maybe they already are?). The header we're looking for is `HTTP_X_FORWARDED_USER`

wlach:
Going to try another approach after discussing with @robotblake ...

PR #1970, opened 2019-06-25
include pycache from server

teonbrooks:
I am seeing  `__pycache__` files from the server when developing. adding the directory to the gitignore list


wlach:
Hadn't noticed this but it can't hurt.

PR #1984, opened 2019-07-01
Fix authentication holes in files api

wlach:
We allowed modifying or adding files to any notebook for a logged in user,
which obviously isn't what we want. We should only allow the user that
owns a notebook to do this.

### Pull Request checklist

### DIFF
--- a/server/files/api_views.py
+++ b/server/files/api_views.py
@@ -21,7 +21,8 @@ class FileViewSet(viewsets.ModelViewSet):
         super().perform_destroy(instance)
 
     def create(self, request):
-        (metadata, file) = (json.loads(self.request.data["metadata"]), self.request.data["file"])
+        metadata = json.loads(self.request.data["metadata"])
+        file =  self.request.data["file"]
 
         notebook = Notebook.objects.get(id=metadata["notebook_id"])
         if notebook.owner != self.request.user:
--- a/server/files/api_views.py
+++ b/server/files/api_views.py
@@ -4,6 +4,7 @@ from django.core.exceptions import PermissionDenied
 from rest_framework import viewsets
 from rest_framework.response import Response
 
+from ..notebooks.models import Notebook
 from .models import File
 from .serializers import FilesSerializer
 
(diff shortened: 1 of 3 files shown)


server/files/api_views.py
```diff
@@ -21,14 +22,24 @@ def perform_destroy(self, instance):
 
     def create(self, request):
         (metadata, file) = (json.loads(self.request.data["metadata"]), self.request.data["file"])
```
jezdez: No need for the extra value unpacking here, easier to spot diffs later on.
```suggestion
        metadata = json.loads(self.request.data["metadata"])
        file =  self.request.data["file"]
```

server/files/api_views.py
```diff
@@ -21,14 +22,24 @@ def perform_destroy(self, instance):
...
+
+        notebook = Notebook.objects.get(id=metadata["notebook_id"])
```
jezdez: The `get()` queryset method raises `DoesNotExist` or `MultipleObjectsReturned` in case there are no items or more than one (the latter being harder here since it queries for a primary key here). I would suggest to use `get_object_or_404` here to explicitly handling `DoesNotExist` here.

server/files/api_views.py
```diff
@@ -21,14 +22,24 @@ def perform_destroy(self, instance):
...
+        if notebook.owner != self.request.user:
+            raise PermissionDenied
+
         f = File.objects.create(
```
jezdez: Let's use more-than-one-character variable names please. 🌞

server/files/api_views.py
```diff
@@ -21,14 +22,24 @@ def perform_destroy(self, instance):
...
-            notebook_id=metadata["notebook_id"], filename=metadata["filename"], content=file.read()
+            notebook_id=notebook.id, filename=metadata["filename"], content=file.read()
         )
         serializer = FilesSerializer(f)
         return Response(serializer.data, status=201)
 
     def update(self, request, pk):
         (metadata, file) = (json.loads(self.request.data["metadata"]), self.request.data["file"])
```
jezdez: Same as above.

server/files/api_views.py
```diff
@@ -21,14 +22,24 @@ def perform_destroy(self, instance):
```
jezdez: Since you're checking on the HTTP level for permissions, it makes sense to override this in the `destroy` method instead of the `perform_destroy` method:

```pytho
    def destroy(self, request, *args, **kwargs):
        instance = self.get_object()
        if instance.notebook.owner != request.user:
             raise PermissionDenied
        return super().destroy(request, *args, **kwargs)
```

server/files/api_views.py
wlach: I think I still have a hangover from python2, where `file` was a reserved keyword...

jezdez: *totally* get it!

wlach:
Thanks for the review! I implemented all your suggestions.

PR #1989, opened 2019-07-03
Use mapDispatchToProps shorthand

wlach:
Where possible / easy, let's use redux's convenience APIs (https://react-redux.js.org/using-react-redux/connect-mapdispatch) for connecting dispatchers to avoid boilerplate.

This is just some incremental cleanup


bcolloran:
looks great, thanks for cleaning this up @wlach

Issue #1943, opened 2019-06-14
Iodide should not hardcode server addresses


wlach:
Right now we hardcode server addresses in the iodide .js during the webpack phase of the build:

https://github.com/iodide-project/iodide/blob/b026e53cfb48d334254175562045d5f980295b43/webpack.config.js#L27

To deploy internally at Mozilla (where we use a docker container), we set the circle environment for these two variables at build time. This has a number of disadvantages:

1. If the origins ever change (as just happened with GCP), we need to remember to update the CircleCI configuration from their interface (very easy to forget)
2. We can't use the same docker container for a production and stage version of the iodide server.

We should somehow fix the webpack builds to not require hardcoding this information. Not sure offhand what the right set of solutions here is, hopefully it's not too complicated.

/cc @bcolloran @robotblake

bcolloran:
The domains are needed to verify the that the communications between the eval-frame and the editor come from the correct hosts. since we're pretty much all-in on the server at this point, i guess we could have django inject the correct domains when it serves the pages (as we do with other NB info) -- they'd still need to be configured in .env for each deployment, but you wouln't have to re-build. so questions then:

1) would that solve this use case?
2) would serving the host info via django rather than hardcoding it at build time be less secure? (i assume not since we serve all kinds of secure token things this way using django?)
3) how big of a deal is this? is it worth a day of someone's time to fix it?

one other thing: if we don't set the host domains at build time, we'd need to fully deprecate start-and-serve mode. i don't use that anymore ever, but it's worth noting.

also cc @hamilton and @mdboom re the start-and-serve bit

wlach:
On some level, yeah, we need this information to come from Django. No, I don't think it's any less secure: If you can't trust what Django is sending you, well, you basically can't trust anything. There's almost certainly a way to keep start-and-serve working with this approach, I suspect it's something we can just add to the webpack config.

I would say this is pretty much something we have to fix before we can put this into production internally. Our internal configuration is already quite complex and has many moving pieces and assumptions, for sure this will come back to bite us again (perhaps at an inopportune time) if we don't fix it sooner.

wlach:
Can probably use something like this technique: https://facebook.github.io/create-react-app/docs/title-and-meta-tags#injecting-data-from-the-server-into-the-page

PR #2024, opened 2019-07-15
Serve iframe from django + resurrect iodide local-only dev environment

wlach:
* The iframe is now served via django from the special `/notebooks/eval-frame/` endpoint. This is architecturally a little cleaner and paves the way towards #1943
* For testing changes to the editor only, you can now use the (renamed)
"simple-serve" command.

### Pull Request checklist


wlach:
@bcolloran please review js parts (python portions optional)
@jezdez could you review the python changes? (and their interactions with the js parts, if you feel up to it). likely a prereq to reviewing is understand the "iodide content in an iframe model" that we have -- either me or @bcolloran could explain this. maybe we could go over it in the iodide meeting tomorrow if you'll be around
@robotblake this is more an FYI than anything else, but grepping through the PR for "middleware" might highlight areas of concern from a deployment POV

docs/local-dev.md
```diff
@@ -8,26 +8,26 @@ Run `npm install` after cloning this repository.
...
+The command runs in watch mode, so changes to files will be detected and bundled automatically, but you will need to refresh the page in your browser manually to see the changes -- we have disabled "hot reloading" because automatically refreshing the browser would cause any active notebooks to lose their evaluation state.
 
-If you require verbose Redux logging, you can use the command `npm run start-and-serve -- reduxVerbose`
-
-#### Exporting from client-only dev mode
-
-In this mode, resource paths are set to be relative to the `dev/` directory. Thus, if you export a bundled notebook from a dev notebook, you need to be sure save the exported HTML file in the `dev/` folder for the relative paths to correctly resolve the required js, css, and font files (and if you want to share a notebook that you created in a dev environment, you'll need to update the paths to point to the web-accessible resources at `iodide.io` and `iodide.app`).
+If you require verbose Redux logging, you can use the command `npm run simple-serve -- reduxVerbose`
```
bcolloran: oops, `--reduxVerbose` doesn't work anymore, one needs to set the environment var `REDUX_LOGGING` to `VERBOSE` or `VERY_VERBOSE`-- e.g.:

`REDUX_LOGGING=VERBOSE npm run start`
or
`REDUX_LOGGING=VERBOSE npm run simple-serve`

i think i need to update the docs for this... :-(

robotblake:
I don't see anything I'd be concerned about off the top of my head.

bcolloran:
(oh also: there aren't really an JS changes yet in this -- i expect you have other stuff coming in follow-up)

wlach:
@bcolloran the port issue comes from the fact that currently webpack is picking up some stuff from .env, where we had specified these values with our dockerflow-based environment in mind. Oops. For now I'll address this by just serving from port 8000 again, as we do for the internal server -- we can change this later after I move to the next phase of this work, which is to stop specifying server names inside our webpack-generated js altogether.

docs/local-dev.md
wlach: It seems like this is the only place where `--reduxVerbose` is mentioned, so I think we can just change it here and be done with it.

QA-CHECKLIST.md
```diff
@@ -11,7 +11,7 @@
 When going through the following checklist, keep the browser console open. There should not be any errors in the console!
 
 a rigourous check will verify that the items listed below work as expected in:
-1. localhost in client only mode (`npm start-and-serve`)
+1. localhost in client only mode (`npm simple-serve`)
```
rafrombrc: This is `npm run simple-serve` everywhere else... should it be here?

wlach: Oops, yes! Good check, will fix before landing

PR #2038, opened 2019-07-16
deprecate the "environment" api and related code

bcolloran:
@hamilton this touches quite a few files actually, but should be quick to check. QA is easy -- just run any code and verify that the declared vars pane works as expected

### Pull Request checklist

- [ ] **Documentation**: we were already not documenting this feature (it's been soft-deprecated for a while)
- [ ] **Tests**: cleared out and fixed tests


src/eval-frame/components/panes/declared-variables-pane.jsx
```diff
@@ -23,55 +19,21 @@ export class DeclaredVariablesPaneUnconnected extends React.Component {
...
-      </div>
-    ) : (
-      undefined
-    );
     return (
       <div className="pane-content">
-        {noVariables}
-        {edvElem}
-        {declaredVariables}
+        <div className="declared-variables-list">
```
bcolloran: turns out we don't need any of the conditionals removed from above, b/c the declared variables pane will now _always_ show at least `iodide` in the user variable explorer.

src/eval-frame/style/side-panes.css
```diff
@@ -1,5 +1,5 @@
 .side-pane {
-  width:auto !important;
+  width: auto !important;
```
bcolloran: i just cut a few things from this file, but prettier made a ton of changes :-/

PR #2039, opened 2019-07-17
Extend shape of property types for user

mehmetoguzderin:
This pull request extends shape of property types for users, effectively making https://github.com/iodide-project/iodide/pull/1962 compliant.


wlach:
thank you!

PR #2066, opened 2019-07-19
Remove origin hardcoding

wlach:
### Pull Request checklist


wlach:
@bcolloran this looks somewhat hacky, but I think it's the right approach. or at least, I can't think of a better way of doing it.

src/editor/initialization/initialize-dom.js
```diff
@@ -1,48 +1,7 @@
...
-}
-if (iframeElt !== null && panesContainerElt !== null) {
-  // in case of html template v2, nothing is required,
-  // but note that this will leave a legacy div#panes-container
-  // wrapping iframe#eval-frame
-}
-if (iframeElt !== null && panesContainerElt === null) {
-  // in case of html template v3, nothing is required
-}
-
```
bcolloran: if we're cutting all the stuff above in this file to support legacy docs (which is totally the right thing to do at this point), we should also now put the two divs specified below directly into the body of the html templates we serve, so that we can eliminate this file entirely

wlach:
@bcolloran cleaned things up as you suggested, this should be ready for another review

bcolloran:
looks good to me @wlach. fyi i haven't done a local QA to check that all the modes work etc because this looks pretty clear. i can do local QA if you want me to, but if you're happy that it's working how you want, please feel free to merge :-)

wlach:
I'm pretty sure this will work! I've already tested it both locally (both simple serve mode and docker compose) and on heroku. I currently have no way of testing it on iodide stage but am 99% sure it should be fine there too. I'll go ahead and merge.

Issue #2046, opened 2019-07-17
Revision browser shouldn't show unsaved changes when there are no changes


wlach:
Currently the iodide revision browser shows an unsaved changes when the notebook is already fully up to date:

![image](https://user-images.githubusercontent.com/20569/61396223-3ee06e80-a895-11e9-8ed3-7622a830210a.png)

We should only show this item when there are changes not yet synchronized with the server (i.e. we actually have a difference to show, not just "no difference").

The logic to display this item in the list is here:

https://github.com/iodide-project/iodide/blob/e361ed03c6ae04c015a1ba19e4f51dd012870f4e/src/editor/components/modals/revision-list.jsx#L52

The logic to calculate the difference between two revisions is here, it might provide some hints on how to implement this feature:

https://github.com/iodide-project/iodide/blob/e361ed03c6ae04c015a1ba19e4f51dd012870f4e/src/editor/components/modals/revision-diff.jsx

PR #2074, opened 2019-07-22
Hide empty revision

mehmetoguzderin:
This pull request hides "Unsaved Changes" button and resorts to last available revision when there is no change, potentially solving https://github.com/iodide-project/iodide/issues/2046 issue.

### DIFF
--- a/src/editor/actions/actions.js
+++ b/src/editor/actions/actions.js
@@ -166,51 +166,43 @@ export function getNotebookRevisionList() {
     });
     getRevisionList(getNotebookID(getState()), isLoggedIn(getState()))
       .then(revisionList => {
-        if (getState().notebookHistory.selectedRevisionId === undefined) {
-          getRevisions(
-            getNotebookID(getState()),
-            [revisionList[0].id],
-            isLoggedIn(getState())
-          )
-            .then(revisions => {
-              if (revisions.length > 0) {
-                if (getState().iomd.trim() === revisions[0].content.trim()) {
-                  dispatch({
-                    type:
-                      "UPDATE_NOTEBOOK_HISTORY_BROWSER_SELECTED_REVISION_ID",
-                    selectedRevisionId: revisions[0].id
-                  });
-                  dispatch({
-                    type:
-                      "UPDATE_NOTEBOOK_HISTORY_BROWSER_HAS_LOCAL_ONLY_CHANGES",
-                    hasLocalOnlyChanges: false
-                  });
-                  dispatch({
-                    type: "GOT_NOTEBOOK_REVISION_LIST",
-                    selectedRevisionId: revisions[0].id,
-                    revisionList
-                  });
-                  getRequiredRevisionContent(getState(), dispatch);
-                } else {
-                  dispatch({
-                    type: "GOT_NOTEBOOK_REVISION_LIST",
-                    revisionList
-                  });
-                }
+        getRevisions(
+          getNotebookID(getState()),
+          [revisionList[0].id],
+          isLoggedIn(getState())
+        )
+          .then(revisions => {
+            if (revisions.length > 0) {
+              if (getState().iomd.trim() === revisions[0].content.trim()) {
+                dispatch({
+                  type: "UPDATE_NOTEBOOK_HISTORY_BROWSER_SELECTED_REVISION_ID",
+                  selectedRevisionId: revisions[0].id
+                });
+                dispatch({
+                  type:
+                    "UPDATE_NOTEBOOK_HISTORY_BROWSER_HAS_LOCAL_ONLY_CHANGES",
+                  hasLocalOnlyChanges: false
+                });
+                dispatch({
+                  type: "GOT_NOTEBOOK_REVISION_LIST",
+                  selectedRevisionId: revisions[0].id,
+                  revisionList
+                });
+                getRequiredRevisionContent(getState(), dispatch);
               } else {
-                dispatch({ type: "ERROR_GETTING_NOTEBOOK_REVISION_LIST" });
+                dispatch({
+                  type: "GOT_NOTEBOOK_REVISION_LIST",
+                  revisionList
+                });
+                getRequiredRevisionContent(getState(), dispatch);
               }
-            })
-            .catch(() => {
+            } else {
               dispatch({ type: "ERROR_GETTING_NOTEBOOK_REVISION_LIST" });
-            });
-        } else {
-          dispatch({
-            type: "GOT_NOTEBOOK_REVISION_LIST",
-            revisionList
+            }
+          })
+          .catch(() => {
+            dispatch({ type: "ERROR_GETTING_NOTEBOOK_REVISION_LIST" });
           });
-          getRequiredRevisionContent(getState(), dispatch);
-        }
       })
       .catch(() => {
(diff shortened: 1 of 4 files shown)


src/editor/components/modals/revision-list.jsx
```diff
@@ -75,11 +89,26 @@ class RevisionListUnconnected extends React.Component {
...
+  } = notebookHistory;
+
+  let hideLocalChanges = true;
+
+  if (revisionContentFetchStatus === "IDLE") {
+    const previousRevisionContent =
+      revisionList.length > 0 ? revisionContent[revisionList[0].id] : "";
+
+    hideLocalChanges = state.iomd === previousRevisionContent;
+  }
```
wlach: I think we could rewrite this as something like this:

```js
const haveLocalChanges = revisionList && revisionList.length && state.iomd !== previousRevisionContent;
```

At which point you could just include it below as a parameter to the object you're returning: 

```js
return {
  revisionList,
  ...
  haveLocalChanges: revisionList && revisionList.length && state.iomd !== previousRevisionContent
}
```

src/editor/components/modals/revision-list.jsx
```diff
@@ -23,6 +23,7 @@ class RevisionListUnconnected extends React.Component {
       })
     ),
     selectedRevisionId: PropTypes.number,
+    hideLocalChanges: PropTypes.bool,
```
wlach: I think this would be more intuitive to express as `haveLocalChanges`, rather than its negation.

src/editor/components/modals/revision-list.jsx
```diff
@@ -40,17 +41,30 @@ class RevisionListUnconnected extends React.Component {
...
+    if (this.props.revisionList !== undefined) {
+      if (
+        this.props.hideLocalChanges &&
+        this.props.selectedRevisionId === undefined
+      ) {
+        if (this.props.revisionList.length > 0) {
+          this.revisionClicked(this.props.revisionList[0].id);
+        }
+      }
+    }
```
wlach: I don't think this should be necessary? We seemed able to show the right set of changes without firing the event on render before, not sure why that should be different now?

mehmetoguzderin: In case this does not take place, it just shows "No difference". With this, modal defaults to last revision, showing its difference with previous version.

src/editor/components/modals/revision-list.jsx
mehmetoguzderin: Makes sense

src/editor/components/modals/revision-list.jsx
wlach: Ah, I see. I think we might need to put some of this calculation at the reducer level, which will let the sub-components (like this one and the actual revision viewer) act accordingly. You probably want to put the logic that tries to figure out whether we have local changes here:

https://github.com/iodide-project/iodide/blob/52170a96a5bce10828cf113e395cebdf4365bc8b/src/editor/reducers/notebook-reducer.js#L81

You may also want to perform some more intelligent logic to pick a sensible default value for selectedRevisionId (should be `undefined` if we have local-only changes, some integer value if we don't).

In general, we try to avoid business logic inside a render() function, so if you find yourself needing to do that it's a good sign that things are a bit off-track.

Hope that wasn't too confusing. Please feel free to ask more questions if anything else is unclear!

mehmetoguzderin: Yes, that will make the code more consistent. I will implement that.

src/editor/reducers/notebook-reducer.js
```diff
@@ -79,13 +79,26 @@ const notebookReducer = (state = newNotebook(), action) => {
...
+        passedRevisionId = revisionList[0].id;
+      }
+
       return Object.assign({}, state, {
         notebookHistory: {
           ...(state.notebookHistory || {}),
           revisionList,
           revisionListFetchStatus: "IDLE",
-          selectedRevisionId
+          passedRevisionId
```
wlach: I think you may have forgotten to commit some other changes? I don't see passedRevisionId (or any changed logic to use it) in the other files.

Also, I think the right thing here is to calculate `haveLocalChanges` in this reducer. I'm actually not sure what passedRevisionId is supposed to represent.

src/editor/actions/actions.js
```diff
@@ -160,13 +160,49 @@ export function updateSelectedRevisionId(selectedRevisionId) {
 export function getNotebookRevisionList() {
   return (dispatch, getState) => {
     dispatch({ type: "GETTING_NOTEBOOK_REVISION_LIST" });
+    dispatch({
+      type: "UPDATE_NOTEBOOK_HISTORY_BROWSER_HAS_LOCAL_ONLY_CHANGES",
+      hasLocalOnlyChanges: true
+    });
```
wlach: Hmm, I'm not sure if this clause is necessary? We should figure this out after we've finished downloading the revision list. Also see below for more comments on this reducer. :)

src/editor/actions/actions.js
```diff
@@ -160,13 +160,49 @@ export function updateSelectedRevisionId(selectedRevisionId) {
...
+          [revisionList[0].id],
+          isLoggedIn(getState())
+        )
+          .then(revisions => {
+            if (revisions.length > 0) {
+              if (getState().iomd.trim() === revisions[0].content.trim()) {
+                dispatch({
+                  type: "UPDATE_NOTEBOOK_HISTORY_BROWSER_SELECTED_REVISION_ID",
+                  selectedRevisionId: revisions[0].id
+                });
```
wlach: This dispatch should not be necessary, it should be covered by the dispatch of GOT_NOTEBOOK_REVISION_LIST below.

src/editor/actions/actions.js
```diff
@@ -160,13 +160,49 @@ export function updateSelectedRevisionId(selectedRevisionId) {
...
+                dispatch({
+                  type:
+                    "UPDATE_NOTEBOOK_HISTORY_BROWSER_HAS_LOCAL_ONLY_CHANGES",
+                  hasLocalOnlyChanges: false
+                });
```
wlach: I don't think this is necessary -- we can calculate it on-demand inside GOT_NOTEBOOK_REVISION_LIST (see my comment inside notebook-reducer.js) as follows:

```js
{
  ...
  revisionList,
  haveLocalOnlyChanges: state.iomd === revisionList[0].content
}
```

In that case, I think we might actually be able to avoid making changes to this file entirely, unless there's something I'm missing.

src/editor/reducers/notebook-reducer.js
```diff
@@ -110,6 +110,16 @@ const notebookReducer = (state = newNotebook(), action) => {
       });
     }
 
+    case "UPDATE_NOTEBOOK_HISTORY_BROWSER_HAS_LOCAL_ONLY_CHANGES": {
```
wlach: You should be able to remove this if you follow my suggestions above.

src/editor/state-schemas/state-schema.js
```diff
@@ -213,6 +213,10 @@ export const stateProperties = {
   notebookHistory: {
     type: "object",
     properties: {
+      hasLocalOnlyChanges: {
+        type: "boolean",
+        default: true
```
wlach: I feel like the default for this variable should be "false".

PR #2085, opened 2019-07-24
clean out eval frame (move workspace and console panes to editor, remote reps, etc)

bcolloran:
### Pull Request checklist


bcolloran:
@hamilton @mdboom -- this PR guts the eval frame...

!!!!

highlights:
- console and workspace panes now in editor scope
- rep code moved to top level and clarified substantially (tho there is more to do...)
- added jupyter rendered_html styles to editor scope to allow jupyter HTML strings to render as desired in the reps (since these are part of jupyter and not just python, i figure it makes sense to inclde them to support html string fallback reps from all languages that have them) @mdboom -- when we land that, i think we can remove the need for the pyodide loader to load its own stylesheet?
- added html santization to user rep html strings -- only allows a whitelist of tag types, following chrome devtools: https://docs.google.com/document/d/1FTascZXT9cxfetuPRT2eXPQKXui4nWFivUnS_335T3U/preview#  (i also added in a couple that i saw the pandas table using)

i have looked at this a bunch, using the following notebooks (locally of course)
https://alpha.iodide.io/notebooks/2785/
https://alpha.iodide.io/notebooks/2784/
https://alpha.iodide.io/notebooks/2718/
https://alpha.iodide.io/notebooks/2756/
and I've fiddled around in the storybook a lot to make sure the reps all work as expected now that i've plumbed them to operate on remote serializations

but i haven't tested everything -- i don't know if there are pyodide reps other than the pandas rep that should be tested, i haven't played around in the console extensively, and i haven't kicked the tires on every last button and widget.

i will do some more of that tomorrow, but i'd love eyes on QA in particular-- just testing every little nook and cranny to make sure things behave as expected/desired.

@hamilton once we land this PR, there's a bit more clean up i can do in the eval-frame before handing over to you to tackle the styling stuff, but this lays a solid foundation, and it seems like a good place to stop for now, since just rewiring everything has already touched a ton of stuff. but we're in great shape!!!

src/eval-frame/iodide-evaluation-results.js
```diff
@@ -1,2 +1,5 @@
 export const IODIDE_EVALUATION_RESULTS = {};
+// FIXME: this should not be atttached to window --
+// in repInfoRequestResponse, we can set the environment to use the correct
+// root object rather than passing down rootObjName
```
hamilton: couldn't this in fact now be part of the redux store? the `IODIDE_EVALUATION_RESULTS` are serialized objs in some format, correct?

hamilton: oh nvm

hamilton:
maybe this also counts as a follow-on improvement, but with this change does the iframe need to be any bigger than the area it takes up in golden layout?

Screen Shot 2019-07-26 at 1 05 05 PM

mdboom:
Commenting just on the Python / Jupyter CSS stuff.  I can confirm that Pandas rendering looks correct with this change.  The actual source of that CSS is written in Less [here](https://github.com/jupyter/notebook/blob/b8b66332e2023e83d2ee04f83d8814f567e01a4e/notebook/static/notebook/less/renderedhtml.less).  Pyodide includes the `less` in its source and runs it through `lessc` at build time to make it easier to update it in the future if Jupyter changes it upstream.  Is it hard to build `less` into the pipeline here and convert it to `css` at build time?  If so, maybe add a comment to the top of the `.css` file with a link to the original source and the command necessary to convert it to `.css`.

Also for completeness, we should remove [these lines](https://github.com/iodide-project/pyodide/blob/master/src/pyodide.js#L373) from Pyodide if we go this route.

bcolloran:
@hamilton good catch on the scrolling, css fixed. i also went through a search-all-files for the other styles that might have been affected -- there was just one that was missing (responsible for the scrolling problem), and i was able to eliminate a ton of styles that are not used anywhere. I think the iframe size should be a follow-up, because i'll have to touch several other files in the editor that are somewhat orthogonal. i created #2099 to track that.

@mdboom added #2098 to track that issue

hamilton:
@bcolloran this looks good to me. Feel free to merge at your discretion.

bcolloran:
ok, i did another QA and didn't find any. i think i clicked through every menu -- i noticed a new issue that is already present on master (filed #2101, but beyond the scope of this particular PR). something may have slipped through, but i don't know what else to look for, so i'm going to merge. thanks for review guys!

PR #2086, opened 2019-07-24
Clear local autosave on load if no different + fix local autosave detection

wlach:
It is occasionally possible for us to have a copy of a local autosave
which has identical contents to what's already on the server in certain
rare cases (e.g. if we send a no-op update to the server).

We actually need a better model of what's synced/not synced with the
server to handle this properly, but I believe this fix will handle 99% of
the likely cases.

### Pull Request checklist

- [N/A] **Changelog**: This PR updates the [changelog](../CHANGELOG.md) with any user-visible changes.

### DIFF
--- a/src/editor/actions/actions.js
+++ b/src/editor/actions/actions.js
@@ -4,7 +4,7 @@ import {
   getRevisionList,
   getRevisions
 } from "../../shared/server-api/revisions";
-import { havePendingAutosavedChanges } from "../tools/local-autosave";
+import { haveLocalAutosave } from "../tools/local-autosave";
 import {
   getNotebookID,
   getUserDataFromDocument,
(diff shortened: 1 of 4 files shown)


wlach:
Wow, what a rabbit hole #2046 turned out to be -- what I thought was a simple bug turned out to reveal some big holes in our autosave model. 

My original assumption was that if we had an "autosaved" revision, it was a clear sign that there were pending unsynced changes to the server, and we should display those as "Unsaved changes" in the revision list. We indeed have code that does this:

https://github.com/iodide-project/iodide/blob/6490ad15dd3b4e16737776af3da816c61d6898da/src/editor/actions/server-save-actions.js#L63

What I forgot is that it's relatively easy when editing to attempt to save a no-op revision (i.e. identical to previous to the server), which generates an error which we currently swallow:

https://github.com/iodide-project/iodide/blob/6490ad15dd3b4e16737776af3da816c61d6898da/src/editor/actions/server-save-actions.js#L33

In this case, an autosaved entry will stick around indefinitely, until you make further edits to the document. My proposed (hacky) solution is to just to delete a corrupt/no-op autosave when attempting to restore it, which I think should solve most instances of this problem.

A future fix would be to keep track of what the current set of changes are compared to our last known-revision and only send a new revision to the server in the case that it is actually different.

wlach:
I'm just going to go ahead and merge this for now, as I think I've solved the issue and am pretty sure it's not worth discussing further.

PR #2094, opened 2019-07-27
Fix client only mode port in local-dev.md

ZiyaoWei:
Could be a configuration quirk on my local machine but `npm run simple-serve` seems to serve the notebook at 8000 and not 8080. Don't know iodide's stack well enough to figure out if it's a config issue so please consider this part bug report part pull request.

EDIT: After a couple minutes of digging it seems since https://github.com/iodide-project/iodide/blob/master/webpack.config.js#L20 is set to 8000 the notebook environment is always served at 8000. LMK if I missed something!

### Pull Request checklist


bcolloran:
ah yes, you are correct! some changes around this stuff were reverted recently, but we must have forgotten to revert the docs. thank you for the PR @ZiyaoWei  :-)

(@wlach just fyi in case you want to look at serving on a different port in the future)

Issue #2047, opened 2019-07-17
Revision browser should allow reverting notebook to previous state


wlach:
Currently the revision browser lets you view a summary of what changed, but doesn't let you actually revert the notebook content to that of a previous revision.

![image](https://user-images.githubusercontent.com/20569/61397697-76044f00-a898-11e9-9ffc-bfe51dd716e1.png)

We should, at minimum, add a button to do exactly like this. Before actually replacing the content, it should pop up a warning dialog confirming the user wants to do this. Google docs provides a good UI example on how this is done:

![image](https://user-images.githubusercontent.com/20569/61397808-b1068280-a898-11e9-80c5-02ee4ca0fdff.png)

Depending on where we go with iodide, we probably want to enable other things in the future  (e.g. forking a "branch" from a previous revision) but for now this will move the ball forward a bit.

ZiyaoWei:
Interested in contributing to Iodide - is this issue a good first issue to pickup? If not, I'll browse around and try to find something else (open to suggestions!).

wlach:
> Interested in contributing to Iodide - is this issue a good first issue to pickup? If not, I'll browse around and try to find something else (open to suggestions!).

I think @mehmetoguzderin wanted to pick this one up, #2114 which you filed would be a good first issue to try. I can probably suggest other things on gitter when I get back in on Tuesday.

bcolloran:
this functionality was added somewhere along the way...

PR #2146, opened 2019-08-12
Add revert functionality for history modal

mehmetoguzderin:
This pull request adds revert functionality for history modal with a button that gets displayed when the selected revision identifier is set. Developed according to https://github.com/iodide-project/iodide/issues/2047 issue.


src/editor/reducers/notebook-reducer.js
```diff
@@ -103,6 +103,15 @@ const notebookReducer = (state = newNotebook(), action) => {
       });
     }
 
+    case "REVERT_TO_NOTEBOOK_HISTORY_BROWSER_SELECTED_REVISION_ID": {
```
wlach: There should be no need to define a specific action for this. You should be able to use the initial `UPDATE_IOMD_CONTENT` action (exposed via the `updateIomdContent` function)

src/editor/components/modals/history-modal.jsx
```diff
@@ -54,15 +91,21 @@ class HistoryModalUnconnected extends React.Component {
...
+    revisionListFetchStatus,
+    revisionContent,
+    selectedRevisionId
+  } = notebookHistory;
 
   return {
     errorGettingRevisionList,
-    revisionListFetchStatus
+    revisionListFetchStatus,
+    canRevert: selectedRevisionId !== undefined && revisionContent
```
wlach: We should also disable / hide this button if the user is on the latest revision.

wlach:
Another alternative, now that I think about it, would be to put a top-level header above the diff with the title of the old revision, something like what github uses for its diff view:

image

You could then align the "revert" button to the right of the title, maybe something like this:

image

We could add lots of interesting options to this header in the future: fork a notebook at a specific revision, toggle between diff view vs. viewing the raw iomd, etc

src/editor/components/modals/history-modal.jsx
mehmetoguzderin: With migration of logic to revision-diff, behavior is corrected.

src/editor/reducers/notebook-reducer.js
mehmetoguzderin: Done

src/editor/components/modals/revision-diff.jsx
```diff
@@ -34,6 +42,22 @@ class RevisionDiffUnconnected extends React.Component {
...
+            style={{
+              right: 32,
+              bottom: 16,
+              position: "fixed",
+              color: "#fff",
+              background: THEME.clientModal.background,
+              borderRadius: 8
+            }}
+          >
+            Revert
```
wlach: I think "Revert to this revision" is a little clearer.

src/editor/components/modals/reversion-modal.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+  render() {
+    return (
+      <Modal
+        visible={this.props.visible}
+        onCloseOrCancel={this.props.onCloseOrCancel}
+        aboveOtherModals={this.props.aboveOtherModals}
+      >
+        <ModalTitle>Revert to this revision</ModalTitle>
+        <ModalContent>This action cannot be undone.</ModalContent>
```
wlach: This actually isn't technically true. :) You can always revert back to what was there before. I'd recommend using something similar to what Google docs uses as outlined in the orginal issue:

https://github.com/iodide-project/iodide/issues/2047#issue-469355732

Something like:

```jsx
<ModalTitle>Restore this revision?</ModalTitle>
<ModalContent>This will restore your notebook to version from 16 July at 14:53</ModalContent>
```

src/editor/components/modals/reversion-modal.jsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Modal from "../../../shared/components/modal";
+import ModalTitle from "../../../shared/components/modal-title";
+import ModalContent from "../../../shared/components/modal-content";
+import ModalCall from "../../../shared/components/modal-call";
+import { TextButton } from "../../../shared/components/buttons";
+
+export default class DeleteModal extends React.Component {
```
wlach: DeleteModal is clearly wrong -- maybe use something like `RevertConfirmationModal`? (best to give the file a similar name)

src/editor/components/modals/revision-diff.jsx
```diff
@@ -69,8 +114,15 @@ export function mapStateToProps(state) {
   return {
     currentRevisionContent,
     revisionContentFetchStatus,
-    previousRevisionContent
+    previousRevisionContent,
+    canRevert:
+      selectedRevisionId !== undefined &&
+      state.iomd !== currentRevisionContent &&
```
wlach: technically we might also want to check the title for differences here, but in practice this dialog obscures the title so we're probably best off leaving this as-is for now.

src/editor/components/modals/reversion-modal.jsx
mehmetoguzderin: 😸

Date is now displayed.

src/editor/actions/history-modal-actions.js
```diff
@@ -48,6 +49,22 @@ export function updateSelectedRevisionId(selectedRevisionId) {
...
 }
 
+export function revertToSelectedRevisionId() {
+  return (dispatch, getState) => {
+    dispatch({
+      type: "UPDATE_IOMD_CONTENT",
+      iomd: getState().notebookHistory.revisionContent[
+        getState().notebookHistory.selectedRevisionId
+      ]
+    });
```
wlach: It occurs to me also that this won't restore the title of the previous revision, but again, it might be better to wait for that until we show the title in this dialog.

src/editor/components/modals/revision-diff.jsx
```diff
@@ -50,7 +98,13 @@ export function mapStateToProps(state) {
 
   let currentRevisionContent;
   let previousRevisionContent;
+  let revisionDate;
   if (selectedRevisionId !== undefined && revisionContent) {
+    for (const revision of revisionList) {
+      if (revision.id === selectedRevisionId) {
+        revisionDate = revision.created;
+      }
+    }
```
wlach: I think we'd be best off just keeping a reference to the selected revision here, that way we can get the needed properties (like `id` or `created`) where needed. In this case, you can just use find to get it:

```js
const selectedRevision = revisionList.find(r => r.id === selectedRevisionId);
```

src/editor/components/modals/revision-diff.jsx
```diff
@@ -69,8 +123,18 @@ export function mapStateToProps(state) {
   return {
     currentRevisionContent,
     revisionContentFetchStatus,
-    previousRevisionContent
+    previousRevisionContent,
+    canRevert:
+      selectedRevisionId !== undefined &&
+      state.iomd !== currentRevisionContent &&
+      revisionContentFetchStatus === "IDLE",
+    date: revisionDate
```
wlach: Let's just pass `selectedRevision` here, and do the formatting where it is actually required based on its `created` property.

src/editor/components/modals/revision-diff.jsx
```diff
@@ -29,11 +48,40 @@ class RevisionDiffUnconnected extends React.Component {
...
       <DiffContainer>
+        <RevertModal
+          visible={this.state.pendingRevert !== false}
+          onCloseOrCancel={() => {
+            this.setState({ pendingRevert: false });
+          }}
+          onRevert={() => {
+            this.props.revertToSelectedRevisionId();
+          }}
+          date={this.props.date}
```
wlach: If you follow my suggestion, this would then be: `this.props.selectedRevision.created`

src/editor/components/modals/revert-modal.jsx
```diff
@@ -0,0 +1,36 @@
...
+  render() {
+    return (
+      <Modal
+        visible={this.props.visible}
+        onCloseOrCancel={this.props.onCloseOrCancel}
+        aboveOtherModals={this.props.aboveOtherModals}
+      >
+        <ModalTitle>Restore this revision?</ModalTitle>
+        <ModalContent>
+          {`This will restore your notebook to version from ${this.props.date}`}
```
wlach: If this is a raw date from the previously selected revision (see below), you would do the actual formatting here.

src/editor/components/modals/revision-diff.jsx
mehmetoguzderin: Much better for further cases

src/editor/components/modals/revision-diff.jsx
mehmetoguzderin: Just needed a tertiary operation to prevent undefined access

src/editor/components/modals/revert-modal.jsx
```diff
@@ -0,0 +1,42 @@
...
+import ModalCall from "../../../shared/components/modal-call";
+import { TextButton } from "../../../shared/components/buttons";
+
+export default class RevertModal extends React.Component {
+  static propTypes = {
+    onCloseOrCancel: PropTypes.func.isRequired,
+    onRevert: PropTypes.func.isRequired,
+    visible: PropTypes.bool,
+    aboveOtherModals: PropTypes.bool,
+    date: PropTypes.object // eslint-disable-line react/forbid-prop-types
```
wlach: Try this instead:

https://stackoverflow.com/a/39627576

src/editor/components/modals/revert-modal.jsx
```diff
@@ -0,0 +1,42 @@
...
+import ModalContent from "../../../shared/components/modal-content";
+import ModalCall from "../../../shared/components/modal-call";
+import { TextButton } from "../../../shared/components/buttons";
+
+export default class RevertModal extends React.Component {
+  static propTypes = {
+    onCloseOrCancel: PropTypes.func.isRequired,
+    onRevert: PropTypes.func.isRequired,
+    visible: PropTypes.bool,
+    aboveOtherModals: PropTypes.bool,
```
wlach: I don't think there is any need to define this (or pass it in)? I think it's always true.

wlach:
@mehmetoguzderin can you add an entry in CHANGELOG.md for this? Aside from that this looks ready.

src/editor/components/modals/revert-modal.jsx
```diff
@@ -0,0 +1,41 @@
...
+        <ModalContent>
+          {`This will restore your notebook to version from ${
+            this.props.date
+              ? format(new Date(this.props.date), "MMM dd, uuuu HH:mm:ss")
+              : ""
+          }`}
+        </ModalContent>
+        <ModalCall>
+          <TextButton onClick={this.props.onCloseOrCancel}>Cancel</TextButton>
+          <TextButton onClick={this.props.onRevert}>Revert</TextButton>
```
hamilton: Would you mind making the "revert" button a `ContainedButton`? This follows the pattern of `ContainedButton` for major action, `TextButton` for minor / cancel actions.

Following the suggestion elsewhere, it should also probably be "restore" instead of "revert"

src/editor/components/modals/revert-modal.jsx
```diff
@@ -0,0 +1,41 @@
...
+  render() {
+    return (
+      <Modal
+        visible={this.props.visible}
+        onCloseOrCancel={this.props.onCloseOrCancel}
+        aboveOtherModals
+      >
+        <ModalTitle>Restore this revision?</ModalTitle>
+        <ModalContent>
+          {`This will restore your notebook to version from ${
```
hamilton: We should add more copy to this so the user has a better understanding of what this does. This functionality actually creates a new revision at the top of the revision stack, making it the latest revision. As a user, this is not clear from the text.

My suggestion:

```
This will restore your notebook to the version from ${dateLogicHere} by saving a copy of it as the most recent version. Your current changes will be preserved as a revision.
```

src/editor/components/modals/revision-diff.jsx
```diff
@@ -29,11 +50,43 @@ class RevisionDiffUnconnected extends React.Component {
...
+            style={{
+              right: 32,
+              bottom: 16,
+              position: "fixed",
+              color: "#fff",
+              background: THEME.clientModal.background,
+              borderRadius: 8
+            }}
+          >
+            Revert
```
hamilton: I would be in favor of using _restore_ as the word to use here, not revert. "Revert" implies undoing all the history that came after a revision, whereas "restore" implies what this functionality is doing – restoring the revision as the most recent one, preserving the rest of the revisions as history.

As such, I would suggest the text here "restore this version", a la Google Docs.

More broadly, I understand that would require going through these files and renaming all the `revert` parts to "restore`. In the long run I think it'll be clearer terminology.

src/editor/components/modals/revert-modal.jsx
mehmetoguzderin: I also believe this wording is more intuitive.

src/editor/components/modals/revision-diff.jsx
mehmetoguzderin: Made the files match the new terminology

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Disable "History" menu item for the tryit notebook
 - Add documentation for how the Iodide server's authentication system works
 - Migrated to Monaco editor, providing several text editing improvements (find/replace text; autocomplete, including hits for Iodide custom chunk types; etc)
+- Add revert functionality for revision browser (#2047)
```
hamilton:
```suggestion
- Add "restore" functionality for revision browser (#2047)
```

hamilton:
@mehmetoguzderin the changes look great. Thanks for the PR!

PR #2147, opened 2019-08-12
More actions cleanup

wlach:
Decided to finish this off, since it was still nagging me. The generic "actions" file is gone for the editor. Added some tests at the same time for app messages and a few other things.

### Pull Request checklist

- [N/A] **Changelog**: This PR updates the [changelog](https://github.com/iodide-project/iodide/blob/master/CHANGELOG.md) with any user-visible changes.
- [*] **Tests**: This PR includes thorough tests or an explanation of why it does not


wlach:
(note: when merging this, please use "rebase and merge"-- the commits are meant to be logical and self-contained)

src/editor/user-tasks/task-definitions.js
```diff
@@ -1,27 +1,17 @@
...
+  toggleWrapInEditors
+} from "../actions/editor-actions";
 import { evaluateText, evaluateNotebook } from "../actions/eval-actions";
 import {
   toggleFileModal,
   toggleHelpModal,
   toggleHistoryModal
 } from "../actions/modal-actions";
-
-// FIXME: remove requirement to import store in this file by attaching
```
bcolloran: please retain this FIXME

src/editor/user-tasks/task-definitions.js
```diff
@@ -120,16 +110,4 @@ tasks.toggleHelpModal = new UserTask({
   }
 });
 
-tasks.fileAnIssue = new ExternalLinkTask({
```
bcolloran: i'm pretty sure these two tasks still need to be defined?

wlach: They aren't referenced anywhere in the UI, I think we're best off removing them.

src/editor/user-tasks/task-definitions.js
wlach: This FIXME looks wrong though. There is no doubt a way to avoid importing store in this file (if that really does matter that much), but I don't think it has anything to do with keypress handling.

bcolloran: it does, the keypress handler needs access to the store to be able to dispatch action.just trust me on this -- tasks are a bit of a tangle right now, and i just added that FIXME like last week to give myself a breadcrumb :-)

src/editor/user-tasks/task-definitions.js
bcolloran: https://github.com/iodide-project/iodide/blob/b846ff15c5b5fe166e5adc3377bf8308357838f2/src/editor/components/menu/notebook-task-button.jsx#L6

https://github.com/iodide-project/iodide/blob/b846ff15c5b5fe166e5adc3377bf8308357838f2/src/editor/components/menu/notebook-menu-item.jsx#L6

wlach: Those aren't examples of usage of those tasks? That's just importing the class from which they are derived.

src/editor/user-tasks/task-definitions.js
wlach: Well, ok, I can add it back although I'm still confused about what is meant here and I think I understand this code about as well as anyone except for you at this point. In general I would prefer if we aimed to have the codebase be readable by anyone, not just by one person with specific context or an idea of what they want to do next.

src/editor/user-tasks/task-definitions.js
bcolloran: you're totally right, my mistake

src/editor/user-tasks/task-definitions.js
bcolloran: i understand that concern, but as always, there are many tradeoffs. the truth of the matter is that unless/until we get more help, i'm doing the majority of the client-side engineering, and context switching to write out everything as an issue or switching branches to fix something that i notice while fixing an unrelated issue slows me down a lot and disrupts my thinking process. we've been using FIXMEs this way for a long time to capture these thoughts in a less disruptive way, and for now i'm comfortable with the way this tradeoff balances -- i think FIXME comments like this are valuable to me and hamilton for leaving notes in context, and that they can be easily ignored by folks for whom they are not relevant. but your concern is noted, hopefully we'll be able to improve on this over time :-)

wlach: Yeah, I definitely have concerns about whether the big bang PRs that have been common up to now will scale as we add more contributors. I'd rather get people used to making smaller and more focused PRs, and filing issues as appropriate. But we can save that conversation for a later time, for now I've added this back.

PR #2152, opened 2019-08-13
WIP: drag and drop file uploader on top-level iodide.io

hamilton:
This is a prototype that demonstrates how drag-and-drop _might_ function, meant to be passed off to @openjck who is working through the implementation.

### DIFF
--- a/src/server/pages/home-page.jsx
+++ b/src/server/pages/home-page.jsx
@@ -14,6 +14,10 @@ import NewNotebookButton from "../components/new-notebook-button";
 import FeaturedNotebooks from "../../shared/components/featured-notebooks";
 import { sharedProperties } from "../../server/style/base";
 
+import FileDrop from "../../shared/components/file-drop";
+import { createNotebookRequest } from "../../shared/server-api/notebook";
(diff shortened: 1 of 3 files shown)


src/shared/utils/file-operations.js
```diff
@@ -98,8 +98,18 @@ export function selectAndUploadFile(notebookID, successCallback = () => {}) {
   });
 }
 
-export function makeFormData(notebookID, data, fileName) {
-  const file = valueToFile(data, fileName);
+export function makeFormData(
```
hamilton: this change to `makeFormData` allows you to pass in a file as the data arg. Subject to change.

src/server/pages/home-page.jsx
```diff
@@ -44,6 +48,26 @@ export default class HomePage extends React.Component {
     const { notebookList } = this.props;
     return (
       <div>
+        <FileDrop
```
hamilton: my `FileDrop` component (for now) targets _an entire page_, not a particular component, so I've added it at the top level here. As designed it can actually go _anywhere_ in a component. Feel free to change that!

PR #2163, opened 2019-08-18
Introduce multi stage docker builds to project (fixes #2156)

### LINKED ISSUE #2156: Our docker images should be built using a multi-stage builds
@armenzg just taught me about [multistage docker builds](https://docs.docker.com/develop/develop-images/multistage-build/), I think we use these for generating the iodide docker images. Advantages:

* Much smaller images we can use deployment (should speed up deploys internally, and on Heroku whenever we start using docker images there)
* Can much more easily create a "production" image without dev and test dependencies (should probably do that as part of this work)

The version of docker used to generate the images must be relatively recent (17.05 or higher), but I believe the resulting images can be used just about anywhere.

Here's an example file, much of what's in here would apply to iodide as well: https://github.com/mozilla/bedrock/blob/master/Dockerfile

This isn't a "good first bug" in the sense that some of the exact details of the work required are undefined but should be relatively approachable to someone newish to iodide. I would recommend reading the section on https://iodide-project.github.io/docs/local-dev/ before diving in.

Algentile:
Introduces multi stage docker builds to project which allows the ability to the project while allows the ability to skip specific configuration depending on the target stage provided. #2156
 
### Pull Request checklist


wlach:
@Algentile please ping me when you're ready for review!

Algentile:
Hey @wlach,

Will do, I am going to go back through today and change some things as I was having a few issues with installing the permissions for the python libraries.

Dockerfile
```diff
@@ -22,23 +20,26 @@ RUN apk --no-cache add \
...
+FROM ${APP_ENV}_preinstall AS release
+RUN chown app:app -R .
+USER app
 # Using /bin/bash as the entrypoint works around some volume mount issues on Windows
 # where volume-mounted files do not have execute bits set.
 # https://github.com/docker/compose/issues/2301#issuecomment-154450785 has additional background.
-ENTRYPOINT ["/bin/bash", "/app/bin/run"]
-
-CMD ["dev"]
+ENTRYPOINT ["/bin/bash", "/app/bin/run"]
```
wlach: Should have newline at end of file.

Dockerfile
```diff
@@ -22,23 +20,26 @@ RUN apk --no-cache add \
...
+RUN pip install --require-hashes --no-cache-dir -r requirements/build.txt
+
+FROM base AS dev_preinstall
+RUN echo "Installing dev dependencies"
+
+COPY requirements ./requirements/
+RUN pip install --require-hashes --no-cache-dir -r requirements/all.txt
 RUN DEBUG=False SECRET_KEY=foo ./manage.py collectstatic --noinput -c
 
+FROM ${APP_ENV}_preinstall AS release
```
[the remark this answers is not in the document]
Algentile: Hey @ wlach,

I added changed the stages to be more explicit, now to build successfully we just need to change the CI to build with the command `docker build -t app:build --target <devapp | release>`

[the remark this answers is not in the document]
Algentile: Also, as a followup, I tried to use setup a virtual env while setting up the container, but even after activating I was still having issues after removing the required dependencies needed for installing the python binaries. Since we are using an alpine image what I decided to do instead was to set the dependencies to virtual dependencies and then uninstall them after we built and installed the required binaries. As of now I am only deleting postgresql-dev be we can alias any of the build dependencies and delete them in one step .

Dockerfile
```diff
@@ -1,8 +1,7 @@
-FROM python:3.7.3-alpine
+FROM python:3.7.3-alpine AS base
```
jezdez: I would encourage you to use the non-Alpine Python docker image for this, since it has a bunch of downsides and the recent buster based default Python image is not a lot bigger than Alpine. More details: https://pythonspeed.com/articles/base-image-python-docker-images/

jezdez: Awesome, let me know if you need any help with the review!

Dockerfile
```diff
@@ -7,26 +7,21 @@ EXPOSE 8000
...
-    py-cffi \
-    postgresql \
-    postgresql-dev \
-    postgresql-client
+RUN apt-get update && \
+    apt-get -y upgrade && \
+    apt-get -y install libpq-dev && \
+    apt-get -y install libffi-dev && \
+    apt-get -y install python-dev && \
+    apt-get -y install build-essential
```
jezdez: Any reason why those packages are installed in individual apt-get install calls?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7.3-alpine AS base
+FROM python:3.8.0b3-slim-buster AS base
```
jezdez: I think it's fair to keep it at the non-beta release, so `python:3.7-slim` suffices, looking at https://hub.docker.com/_/python.

Dockerfile
```diff
@@ -1,44 +1,63 @@
...
 
 # Using /bin/bash as the entrypoint works around some volume mount issues on Windows
 # where volume-mounted files do not have execute bits set.
 # https://github.com/docker/compose/issues/2301#issuecomment-154450785 has additional background.
 ENTRYPOINT ["/bin/bash", "/app/bin/run"]
 
-CMD ["dev"]
+FROM base AS release
+
+COPY --from=base --chown=app:app /app/static /app/static
```
wlach: You shouldn't need to copy files from an image you're already inheriting from.

Dockerfile
```diff
@@ -1,44 +1,63 @@
...
+
+# Set user and user permissions
+COPY --chown=app:app . .
```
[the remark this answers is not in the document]
Algentile: Hey @wlach,

I believe I can change the application of the base permissions from release to the base build fine. The only thing with static files is once you change all the permissions you need to copy the assets over applying the specified permissions otherwise the CI will fail because it will with a permissions error when you try to copy the files.

Dockerfile
```diff
@@ -1,44 +1,63 @@
...
+COPY --from=base --chown=app:app /app/static /app/static
+
+# Set user and user permissions
+COPY --chown=app:app . . 
+USER app
+
+FROM base AS devapp
+
+COPY --from=base --chown=app:app /venv /venv
+COPY --from=base --chown=app:app /app/static /app/static
```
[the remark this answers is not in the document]
Algentile: Hey @wlach,

I believe you are right about the static files, although one issue that I was running into with the virtual env is if apply permissions before you pip install the dependencies will install to app/local instead which results in things like CI not being able to find certain commands. This is also why I apply permissions at the release stage. I think a similar issue with dependencies like isort is found with mozilla's Kitsune [https://github.com/mozilla/kitsune/blob/master/Dockerfile](https://github.com/mozilla/kitsune/blob/master/Dockerfile). They also do multi stage builds using virtualenv although it is specific to python2.

Dockerfile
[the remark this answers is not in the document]
Algentile: I commented below, I might be able to move this permission change to the base image, it is just the dev dependencies will be installed to the user_site directory instead which is the main blocker .

Dockerfile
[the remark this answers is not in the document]
wlach: I haven't tested this but I believe since the virtualenv is in your path, any new dependencies should be installed into the virtualenv. You will want to do this after the `USER app` declaration, so they are installed with the correct permissions.

Dockerfile
```diff
@@ -1,44 +1,63 @@
...
+COPY --chown=app:app . . 
+USER app
+
+FROM base AS devapp
+
+COPY --from=base --chown=app:app /venv /venv
+COPY --from=base --chown=app:app /app/static /app/static 
+
+# Install dev python dependencies
+COPY requirements/tests.txt ./requirements/
```
wlach: Didn't notice this before, but I believe this is a no-op, since you already copied the contents of the current directory into `app`.

Dockerfile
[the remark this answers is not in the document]
Algentile: Hey @wlach,

Thank you for the suggestions, i fixed most of the inheritance. 

I think the issue I was running into was really setting the User before any of the installs. I could not install the dependencies to the user level because some of the packages were being checked at the system level. 

In the case that you do not provide a --target I just made sure that if the user is prematurely set at the release stage then the devapp stage will first set the user to root to make sure the dev dependencies are installed at the system level before setting the user back to app. This way you can run either stage individually via docker build kit, individually with only target set, or all without providing a target.

Dockerfile
Algentile: Hey @wlach,

Good catch I removed that copy.

Dockerfile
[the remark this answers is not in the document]
Algentile: It is now added to the base image.

Dockerfile
```diff
@@ -1,44 +1,57 @@
...
+# Set user permissions
+COPY --chown=app:app . . 
+
 # Using /bin/bash as the entrypoint works around some volume mount issues on Windows
 # where volume-mounted files do not have execute bits set.
 # https://github.com/docker/compose/issues/2301#issuecomment-154450785 has additional background.
 ENTRYPOINT ["/bin/bash", "/app/bin/run"]
 
-CMD ["dev"]
+FROM base AS release
```
wlach: I suppose at this point we might as well just remove the `base` app and just call it `release`. :)

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 - Adds a new feature, file sources, which allows notebook authors to    
   schedule the fetching of files from URLs. This feature is available in the notebook menu (click _Menu > Manage Files_)
-
+- Add multi stage docker builds for smaller docker images (#2156)
```
```suggestion
- Add multi stage docker builds for smaller docker images (#2156)

```

Dockerfile
```diff
@@ -1,44 +1,57 @@
...
+USER app
+
+FROM base AS devapp
+
+USER root
```
wlach: I am curious why this is required -- are you sure it's not possible to install these extra dependencies inside the virtualenv as the current user?

jezdez:
@wlach Yeah, while working on the remote kernel feature I thought that I would suggest to stop using `docker build` for building the image but `docker-compose build` since it simplifies the development workflow somewhat. Right now you have to do multiple steps of first building the Docker image and then advising docker-compose to use it in a loosely coupled fashion. Instead I would suggest to just use `docker-compose build --pull`, `docker-compose push` and `docker-compose up` etc.

For specifying which multi-stage target to use when building the compose stack, there is the [target](https://docs.docker.com/compose/compose-file/#target) parameter.

Algentile:
Hey @jezdez ,
I made a small dev workflow change to the docker-compose and tested the changes for docker-compose with the target attribute, but an error at the test-server stage suggests that we will have to update circle ci to 2.1 to support any docker-compose versions past 3.4. Currently max supported docker-compose image is 3.2 with version 2.0 of circle ci.

Also, updating to version 2.1 has a few structural changes with how we deploy because it introduces orbs, and command attributes. This should not be a lot of work to change, but for now I think we should put this work in a separate PR specific to circle CI right after this PR is merged.

Dockerfile
Algentile: This will cause an environment error if you set the app User prior to install.

.circleci/config.yml
```diff
@@ -52,17 +52,18 @@ jobs:
...
-          name: Build Docker image
-          command: docker build -t app:build .
+          name: Build Docker images
+          command: docker build -t app:prod --target=base . && docker build -t app:build . 
       # save the built docker container into CircleCI's cache. This is
       # required since Workflows do not have the same remote docker instance.
       - run:
-          name: docker save app:build
-          command: mkdir -p /tmp; docker save -o /tmp/docker.tar "app:build"
+          name: docker save app:build and app:prod
```
wlach:
```suggestion
          name: Save built images
```

Dockerfile
```diff
@@ -1,44 +1,54 @@
...
+RUN pip install --require-hashes --no-cache-dir -r requirements/build.txt
+
+FROM python:3.7-slim AS base
+
+ENV PIP_DISABLE_PIP_VERSION_CHECK=1
+ENV PATH="/venv/bin:$PATH"
+
+RUN groupadd --gid 10001 app && useradd -g app --uid 10001 --shell /usr/sbin/nologin app
+
+COPY --from=python-builder /venv /venv
```
wlach:
```suggestion
COPY --chown=app:app --from=python-builder /venv /venv
```

Dockerfile
```diff
@@ -1,44 +1,54 @@
...
+
+EXPOSE 8000
```
wlach: I would leave this until the very end of the image definition.

Dockerfile
```diff
@@ -1,44 +1,54 @@
...
 
 WORKDIR /app
 COPY . /app
```
wlach: We can set user to app here as long as we're careful about chowning.

```suggestion
COPY --chown=app . .
RUN chown app /app
USER APP
```

Dockerfile
```diff
@@ -1,44 +1,54 @@
...
-RUN chown app:app -R .
-USER app
 
+# Collect static files
 RUN DEBUG=False SECRET_KEY=foo ./manage.py collectstatic --noinput -c
 
+# Set user permissions
+COPY --chown=app:app . . 
+USER app
```
wlach: Preferable to do this before we run collectstatic (see above). This way the collectstatic command above will validate that the permissions are set correctly.

Dockerfile
wlach: If you make the adjustments above you should be able to remove this. I've tested it locally.

Dockerfile
```diff
@@ -1,44 +1,54 @@
...
 # where volume-mounted files do not have execute bits set.
 # https://github.com/docker/compose/issues/2301#issuecomment-154450785 has additional background.
 ENTRYPOINT ["/bin/bash", "/app/bin/run"]
 
-CMD ["dev"]
+FROM base AS devapp
+
+USER root
+# Install dev python dependencies
+RUN pip install --require-hashes --no-cache-dir -r requirements/tests.txt
```
wlach: We should install everything now: some of the test dependencies now depend on some main dependencies, which means their hashes need to be specified. No harm in this, as pip is smart enough not to reinstall anything already installed.

```suggestion
RUN pip install --require-hashes --no-cache-dir -r requirements/all.txt
```

package-lock.json
```diff
@@ -4504,7 +4504,7 @@
     },
     "callsites": {
       "version": "2.0.0",
-      "resolved": "http://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
```
wlach: Please revert any changes to this file.

Dockerfile
Algentile: Ah, I have tried that, I think if we install all the dependencies like you specified below we could do it this way. Otherwise, it will not work when running through the CI steps because if we install after words then the subsequent installs will be made to the app/local folder and some of the dependencies will not be found when running on the test-server phase.

wlach: As long as the virtualenv is activated, all libraries should be installed into `/venv/`

Dockerfile
Algentile: Hey @wlach, 

Thanks that worked after I added the chown when copying the virtual env.

docker-compose.yml
```diff
@@ -2,6 +2,8 @@ version: '3'
 services:
   server:
     image: app:build
+    build: 
+      context: .
```
jezdez: Just noticed, that this will need to be applied to the other services here, too.

docker-compose.yml
```diff
@@ -2,6 +2,8 @@ version: '3'
 services:
   server:
     image: app:build
```
jezdez: Not need to specify the image here, docker-compose will build this on its own when needed.

docker-compose.yml
jezdez: This will also need the [target](https://docs.docker.com/compose/compose-file/#target) parameter in the docker-compose.yml to be able to use the right stage.

docker-compose.yml
Algentile: Good call I will remove that image field

docker-compose.yml
Algentile: Hey @jezdez,

I can add the target field, but I think I want to hold off on that for now with docker-compose because I will need to up the version to 3.4 and the max supported version for docker-compose with circle CI 2.0 is 3.2.

The reason why I want to hold of is mainly because migrating to circle CI 2.1 adds a few new things that I think could cause some breaking changes and we might want to separate that work into a separate small pr.

package-lock.json
```diff
@@ -6392,9 +6392,9 @@
       }
     },
     "eslint": {
-      "version": "6.3.0",
```
wlach: This still needs to be reverted before landing. You can get the last version of package-lock before these changes by doing this:

```bash
git log package-lock.json
```

Then take that revision, and put it in this command:

```bash
git checkout <revision hash> -- package-lock.json
```

Commit the result and the changes should be reverted.

.circleci/config.yml
```diff
@@ -147,19 +148,19 @@ jobs:
       # build to use only the required Python files.
       - run:
           name: Build Docker image
-          command: docker build -t app:build --build-arg PIP_FILE=build.txt .
+          command: docker build -t app:prod --target=base .
```
wlach: It occurs to me that we could actually restore the production image here, now that we're building it. Not a big deal though, this can wait until a subsequent PR.

jezdez:
@wlach Nope, other than following-up with a docker-compose upgrade to get target support in the docker-compose config.

.circleci/config.yml
Algentile: I can modify this to load the image instead of rebuilding, it should just be a small change.

package-lock.json
Algentile: Thanks @wlach, it appears that the changes have been reverted.

Algentile:
Thanks @wlach and @jezdez for your patience on this PR.  Your guidance was very helpful! :)

PR #2172, opened 2019-08-20
Add server overview and architecture diagram to docs

wlach:


wlach:
@mdboom mostly asking for review to make sure you're ok with changing the mkdocs setup. Installing the `mkdocs-mermaid-plugin` from github is unfortunate but seems unavoidable due to https://github.com/pugong/mkdocs-mermaid-plugin/issues/5

If you want to see what the embedded mermaid diagram looks like, you can preview it here:

https://definitely-staging.iodide.io/notebooks/60/?viewMode=report

bcolloran:
@wlach i don't know exactly what the technical term "iodide webhead django" means, but i'm pretty sure there should be _no_ arrow from the iframe to the webhead in the chart at:
https://definitely-staging.iodide.io/notebooks/60/?viewMode=report

One of the main things we've tried to do to protect security is prevent user code in the iframe from directly accessing the server -- all communications need to be relayed through the editor scope

wlach:
> @wlach i don't know exactly what the technical term "iodide webhead django" means, but i'm pretty sure there should be _no_ arrow from the iframe to the webhead in the chart at:
> https://definitely-staging.iodide.io/notebooks/60/?viewMode=report
> 
> One of the main things we've tried to do to protect security is prevent user code in the iframe from directly accessing the server -- all communications need to be relayed through the editor scope

Makes sense, I'll try to update the diagram to make that more clear. Does the text convey that intention?

bcolloran:
The text is fine. I don't know what level of detail you want here; if you just want to allude to the iframe set up that is fine, but I suppose you could also choose to be more explicit about routing all communications through the editor context to prevent user code from having direct access to server APIs and data. I don't think that is clear from the text, but it may not need to be in this in context.

wlach:
> The text is fine. I don't know what level of detail you want here; if you just want to allude to the iframe set up that is fine, but I suppose you could also choose to be more explicit about routing all communications through the editor context to prevent user code from having direct access to server APIs and data. I don't think that is clear from the text, but it may not need to be in this in context.

A few sentences on the topic can't hurt, especially since this is supposed to be supporting documentation for an internal security audit. :) Updated the docs and the diagram (preview of diagram @ same URL as above)

docs/server-admin-overview.md
```diff
@@ -1,15 +1,42 @@
...
+Redis(fa:fa-database Redis Cache)
+end
+
+Main --> DB
+Main --> Redis
+</div>
+
+The main unconventional piece is that (when properly configured), Iodide
+will serve content from two domains: the bulk of the site and JavaScript
+is served from a "primary" domain (e.g. https://alpha.iodide.io/), but the actual evaluation context for the notebook content is served from a *seperate* domain (e.g. https://alpha.iodide.app/), loaded transparently from the primary one. This is designed to provide some measure of security against a malicious notebook being used (for example) to steal a user's credentials or private information: the evaluation context (eval-frame) has no credentials or cookies associated with it, so is not able to directly access server APIs which are privileged.
```
mdboom:
```suggestion
is served from a "primary" domain (e.g. https://alpha.iodide.io/), but the actual evaluation context for the notebook content is served from a *separate* domain (e.g. https://alpha.iodide.app/), loaded transparently from the primary one. This is designed to provide some measure of security against a malicious notebook being used (for example) to steal a user's credentials or private information: the evaluation context (eval-frame) has no credentials or cookies associated with it, so is not able to directly access server APIs which are privileged.
```

wlach:
good idea re: installing from a specific git revision. i'll try that

Issue #2113, opened 2019-07-31
restructure file heirarchy, state-schemas, actions, reducers, etc to organize by feature rather than function


bcolloran:
just wanted to put together an issue to track the idea @hamilton proposed last week:
right now, our files are mostly split up by _function_ -- components, reducers, actions, etc -- but this makes it hard to know what is in scope when you are working with only one slice of the app.

some references:

- organizing by feature: https://marmelab.com/blog/2015/12/17/react-directory-structure.html -- seems fine. i'd be open to putting tests at the top level rather than in a `__test__` subfolder if folks think that is useful. i see the author's point that it makes it easier to see what stuff needs testing, but it is also kind of clutter, so i'm ambivalent. i'd default to no change unless someone is specifically taken by the idea.

- the "ducks"/"reducks" approach seems popular: https://github.com/erikras/ducks-modular-redux and https://github.com/alexnm/re-ducks . There are other variants too. i really like the idea of kind of name-spacing redux actions per feature/concept as e.g "my-app/widgets/LOAD". I think that would be quite clarifying. we could adapt one of those patterns by also moving the relevant bits of state schema. my hunch is that something based on the folder-oriented "reducks" pattern will scale better.

maybe i'll do a draft PR organizing some feature to see what an option could look like, just for discussion. maybe the console would be good since i was just working on that.

@hamilton @wlach @mdboom @openjck more ideas and reading materials very welcome! :-)

wlach:
I think it would indeed be nice if the actions/reducers/component logic was a little closer together for each feature. I find myself using project search to navigate between files more often than I'd like to admit -- it would be much more convenient if things were closer together in the filesystem so I could use the directory sidebar in atom more easily (since like would be close to like spatially).

The feature example you linked to makes the most initial sense to me -- the reducks approach seems weird, though I'm sure I would get used to it. Doing up a prototype for discussion sounds good.

PR #2180, opened 2019-08-22
[WIP/proposal] Organize code by feature - test using console

bcolloran:
@mdboom @wlach @hamilton following from #2113, as we've discussed a lot recently, here is a test case/proposal for moving to feature-based organization of our code. The console turned out to be an interesting case, because as I began digging in it quickly became clear the the console is architecturally _two distinct_ features that are actually not strongly coupled: the history viewer and the console input. This ended up falling out as a _nested_ feature-based hierarchy, which is probably something we'll see elsewhere, so perhaps its a good case to examine.

i think the basic file structure i'd propose to try out (very much inspired by the links in that issue) for now would be:
* a "components" folder (maybe not needed if there are only a couple components?)
* an "actions" file with *only action creators*, but not thunks, sagas, etc. I think we should _definitely_ adopt the practice of "namespacing" actions like `console/history/UPDATE"; optionally, we could move to the common Redux practice of defining the "types" as a separate object in the actions file and importing them in reducers. I tried that in one of the subfolders, but I'm not sure whether this is actually better. One drawback is that it makes it harder to use search-all-files since the strings are less distinct. i guess it might improve IDE support? I'd really like people's opinions -- TBH i lean towards continuing to hard code the strings b/c i think i'm not understanding the value in exporting them enum-style in this case. but i'm very open to being shown the light ;-)
* a "reducer" file. one thing we'll probably have to start doing in this version of the world is having more than one reducer in different parts of the app respond to the same action. this is not something we've needed to do so far (we currently have a 1-1 mapping between actions and reducer cases) but it is an accepted practice in redux world to have an action fired by e.g. some UI interaction trigger updates to different parts of the state tree that are managed by different reducers. having gone through this exercise, i can already see how that will be needed.
* a "state-schema" file that specifies the state elements specifically relevant to and added by the feature
* "thunks" and "sagas" files as needed. rather than keeping these in "actions" i think it makes sense to keep plain action creators as distinct entities, and to move more involved thunks and sagas that contain business logic to separate files

in the case of this particular nested feature, there are a couple components at the top level under `src/console`, but the subfolders `src/console/history` and `src/console/input` follow the ideas above.

one other thing that we could consider from those examples would be to add an "index" at the root of each feature hierarchy that could export the public API of the feature that would be needed by other features (I think this would almost exclusively be action creators / thunks, since those are the triggers through which parts of the app should communicate by updating the store). This might buy us something in terms of encapsulation, but i'm not sure we need it.

anyway, take a look through this and add your thoughts and ideas. i'm sure there are refinements to be made, but i think doing something like this feels like the right next step


bcolloran:
(oh also: i didn't really make any substantive code changes here. looking at e.g. Immer, combineReducers, etc can be next steps)

bcolloran:
@wlach @hamilton @mdboom did any of you have a chance to look at this? any thoughts?

hamilton:
Read through – looks great! The main benefit / thing I'm hoping for with this change is _sensible colocation_. of files needed for feature development. One confusing aspect of the codebase is that feature work requires me to be in about 5-6 distinct / distinctly-organized directories. It's a real drag. So on the whole, this feels like a welcome direction, probably with some wiggle room as we go. 

A few initial thoughts about some of the details:

1. happy to follow your lead on the action type namespacing, but agree that we should stick to the hard-coding for the time being if we can help it. I haven't really run into any cognitive problems with hard-coding at this point, but making them less distinct would definitely add some workflow friction as I try to search for the string I need. Though this may be fixed because the reducers are colocated next to the actions so we're not hunting for the right thing, but could be an issue for sure if multiple reducers in different parts of the code base get triggered (see the next point)
2. a bit conflicted about the idea of multiple reducers handling the same action. On the one hand, it may be that there are actions that _always_ need to be caught in another reducer. But it seems like that's a lot less common than simply composing required dispatches as thunks, leaving the atomic state updates the provenance of the reducers. Where possible, it would be good to try to follow a model where reducers / action creators are the building blocks for more complex state changes.  Also, keeping those multiple dispatches in thunks does make it much easier to comprehend what's happening in the code since they're quite readable. Looking through our current reducers, most of the cases are fairly atomic as-is, but perhaps that might change as a result of this direction in some way. At any rate, I could be underthinking this – can you point toward where we might need that?
3. keeping `actions` separate from `thunks` etc. is a great idea and does a great job of demarcating what is async/composite and what is atomic.
4. regarding the idea of having the `index.js`, we could also just have `server-api.js` for those particular features.
5. are there actions / reducer cases that aren't features? Does this break the model / how should we address those?

I would be interested in @wlach's thoughts here as well.

bcolloran:
cool thanks for looking at this @hamilton. I'm glad this feels like a mostly right direction, and yes there is definitely wiggle and refinement room as we learn!

I'm also very interested in hearing @wlach's thoughts. I'll leave this as-is for him to review so that he can see both options, but I'm with you in terms of leaning towards hard-coding action names.

my mind is not really deep in this since i was working on it, so i can't recall exactly where it was that i thought i saw a benefit to having multiple reducers to respond to an action, but in any case, that's not anything we'd need to decide on in this PR. sequencing actions in thunks can probably cover basically all cases (it's worked so far!) but i'm open to the idea that we might be able to streamline the interaction of different features that need to affect each other. basically, thinking about how features might interact, the alternatives i can imagine ATM are:
1. have simple reducers that handle disjoint parts of the state tree; and have 1-1 mapping of reducer case to action; and have a thunk that fires two simple actions in sequence 
2. have simple reducers that handle disjoint parts of the state tree; and have these reducers both respond to the same action; and only need to dispatch a single simple action
3. have a larger reducer that handles multiple parts of the state tree; and have 1-1 mapping of reducer case to action; and only need to dispatch a single simple action

i don't think any of these is clearly better or definitively "correct" (though option 3 seems like the worst of them (and is probably closest to where we are today, but just for legacy reasons)), it's just a matter of where we want to put the code that crosses the boundary between features (or where we want to write the code that will be aware of the other feature). In the one case, it would be writing thunks that are aware of the action creators from another feature and dispatch them; in the other, it would be writing reducers that are aware of the actions from another feature. Anyway, I want us to be aware of and open to this other pattern in case it ends up fitting some problem better as we shift how things are organized, but there are certainly no pressing cases of that ATM.

what do you mean by "regarding the idea of having the index.js, we could also just have server-api.js for those particular features."? which particular feature did you have in mind-- just server features?

regarding your question 5 "are there actions / reducer cases that aren't features? Does this break the model / how should we address those?" -- can you restate? i'm not sure what you mean.

wlach:
This all makes sense to me. I would advise keeping the server api interaction bits in their own directory for now -- they kind of stand independently from the machinery that displays/updates the UI and should fairly rarely need to be updated.

Namespacing action names also makes sense to me. I'm with you in not being sure if using enums adds a ton of value. I'd be in favour of just doing the directory reorganization along with a minimal set of code changes (e.g. breaking out seperate reducers), then seeing if further modifications are necessary. 

I've mentioned this before, but I'd like to see us move to making these kinds of changes a little more quickly in smaller PRs (and being open to further adjustments in the future), instead of trying to change/fix everything at once. In that spirit, maybe we should look into breaking out the changing of the names of actions into a follow-up PR.

PR #2183, opened 2019-08-24
Issue 2133 issue 2134 drag and drop

hamilton:
### Pull Request checklist

### DIFF
--- a/src/shared/__stories__/drop-target/components/ImageGallery.jsx
+++ b/src/shared/__stories__/drop-target/components/ImageGallery.jsx
@@ -1,0 +1,185 @@
+import React from "react";
+import ArrowUpward from "@material-ui/icons/ArrowUpward";
+import CircularProgress from "@material-ui/core/CircularProgress";
+import styled from "@emotion/styled";
+
+import DropTarget from "../../../components/drop-target";
+import THEME from "../../../theme";
+import { fadeIn, bounce } from "../../../keyframes";
+
+const Overlay = styled.div`
+  animation: ${fadeIn} 0.25s;
+  background: ${THEME.header.backgroundLeft};
+  font-family: sans-serif;
+  z-index: 10;
+
+  /* Expand to fill the Storybook iframe */
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+
+  &.inactive {
+    display: none;
+  }
+`;
+
+const OverlayContent = styled.div`
+  align-items: center;
+  color: #fff;
+  display: grid;
+  font-weight: bold;
+  height: 100%;
+  justify-items: center;
+  width: 100%;
+
+  &.uploading {
+    align-content: center;
+    grid-gap: 50px;
+  }
+
+  span {
+    font-size: 3rem;
+  }
+
+  #upload-arrow {
+    animation: ${bounce} 1s ease-in-out infinite alternate;
+    font-size: 10rem;
+  }
+`;
+
+const Gallery = styled.div`
+  display: grid;
+  font-family: sans-serif;
+  grid-template-rows: min-content;
+
+  /* Expand to fill the Storybook iframe */
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+`;
+
+const Header = styled.header`
+  background: ${THEME.header.background};
+  color: #fff;
+  display: grid;
+  grid-gap: 10px;
+  padding: 20px;
+
+  h1 {
+    margin: 0;
+  }
+`;
+
+const Title = styled.h1`
+  color: #fff;
+  font-size: 1.3em;
+`;
+
+const Instructions = styled.span`
+  font-size: 0.9em;
+`;
+
+const Body = styled.main`
+  align-items: center;
+  display: grid;
+  justify-items: center;
+  overflow: auto;
+  padding: 20px;
+
+  &.has-images {
+    align-items: start;
+    grid-gap: 20px;
+    grid-template-columns: repeat(auto-fill, 200px);
+    justify-items: start;
+  }
+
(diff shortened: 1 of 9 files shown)


src/editor/components/modals/file-modal/body.jsx
```diff
@@ -2,6 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import styled from "react-emotion";
 
+import DropTarget from "../../../../shared/components/drop-target";
```
hamilton: I would hesitate to have this implemented in the PR too soon. The file modal area is being pretty heavily revamped / moved around a bit in the code base, and once this PR graduates out of "draft" it'll already require a rebase and possibly a refactor. Could you focus on the front page instead? I don't think the drag and drop work will get clobbered there.

src/shared/components/drop-target.jsx
```diff
@@ -0,0 +1,93 @@
...
+  justify-items: center;
+  padding: 20px;
+
+  width: ${props => (props.width ? `${props.width}px` : "100%")};
+
+  /* Specify an exact height so that the target doesn't grow or shrink when the
+     hover message is displayed */
+  height: ${props => (props.height ? `${props.height}px` : "100px")};
+
+  &.hovered {
```
hamilton: could we provide some kind of animation on hover?

src/shared/components/drop-target.jsx
```diff
@@ -0,0 +1,93 @@
+import React from "react";
+import PropTypes from "prop-types";
+import styled from "react-emotion";
+
+const StyledDropTarget = styled.div`
+  align-content: center;
+  align-items: center;
+  background: #ededed;
```
hamilton: I'm wondering if we shouldn't make this a pretty generic container (that is, have the option of changing some of the default styling). This component will likely need to be pretty flexible going forward, so let's consider various options. For instance, let's say we actually add a files drawer (like in vscode / other editors & ides) and we want to make it a drag-and-drop area. What styling choices would we need for this component? It would be great if we could solve this issue in a generalized way once and not have to re-implement it again in the future, so let's think through some ways we can keep it as such.

src/shared/__stories__/drop-target.stories.js
```diff
@@ -0,0 +1,47 @@
...
+  justify-items: center;
+`;
+
+export const actions = {
+  onDrop: action("onDrop")
+};
+
+const stories = storiesOf("DropTarget", module);
+
+stories.add("default", () => (
```
hamilton: it'd be great if there were some feedback after I drag and drop in these stories. Can you add something like listing the file names below the element? One way we've accomplished this is to just write a component function that utilizes the `DropTarget` component you have here + uses hooks to show local state in a list or something.

src/shared/components/drop-target.jsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+`;
+
+export default class DropTarget extends React.Component {
+  static propTypes = {
+    children: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.node),
+      PropTypes.node
+    ]),
+    getHoverMessage: PropTypes.func,
```
hamilton: `getHoverMessage` feels narrowly useful (just the number of files). Seems like it could just be a function that returns the array of files, and the user can return whatever they want (the total number, the names, return a bunch of React components that are to be rendered here). Could you do a pass where it is a bit more flexible?

With that in mind, it does seem like we could here return either a function (which for example could return a React component if we need access to the file contents / metadata) or a rendered React component.

src/shared/__stories__/drop-target.stories.js
```diff
@@ -0,0 +1,47 @@
...
+  <DropTarget {...actions}>
+    <DropTargetContent>
+      <ContainedButton>Add files</ContainedButton>
+      <span>or drag and drop files here</span>
+    </DropTargetContent>
+  </DropTarget>
+));
+
+stories.add("customization", () => (
```
hamilton: I mention this elsewhere, but it would be great to add more stories showing this component's flexibility. The hope is we will be able to use a generic drag and drop for (1) in-body elements where we need to give some clear direction that you can just drag something in, (2) elements where it's not 100% obvious but a user might expect _somethign_ like that to happen (Think a files drawer in the notebook, for example), and (3) a full-body drag and drop. I don't think (3) is probably achievable with what you have right now (and wasn't in your original figma sketches at any rate), but it does seem like (1) and (2) are.

src/shared/components/drop-target.jsx
hamilton: (I can see this done in the story and not the component)

src/editor/components/modals/file-modal/body.jsx
hamilton: actually, maybe I would just focus on the storybook as a starting point

src/shared/components/drop-target.jsx
openjck: Animation has been tricky. Not quite sure why my `transition`s aren't being applied. Something about re-rendering the component, maybe. I'll keep at it.

openjck:
Improved the two stories and pushed them up to this branch. The image gallery is a little finicky. I'll work on that more tomorrow.

src/shared/__stories__/drop-target/components/ImageGallery.jsx
```diff
@@ -0,0 +1,186 @@
...
+
+  /* Expand to fill the Storybook iframe */
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+
+  &.inactive {
+    display: none;
```
hamilton: I love this

openjck:
Superseded by #2327 honestly just because this branch name was misleading.

PR #2219, opened 2019-09-05
Add query remote kernel

jezdez:
This is the current design for remote kernels, specifically for query remote kernels.

### Pull Request checklist


.coveragerc
```diff
@@ -0,0 +1,15 @@
+[run]
```
wlach: As I mentioned on gitter, can we move this to a seperate PR?

.circleci/config.yml
```diff
@@ -130,7 +130,7 @@ jobs:
             # we need to re-collect static as docker-compose mounts the checkout directory over what we have in the image
             chmod 777 . # make sure we have permission to write the files in our volume
             # run the actual tests
-            make test
+            make leantest
```
wlach: I'm not 100% sure what this leantest stuff is, is it about not running redundant checks in CI? Probably easier addressed if we move to a seperate PR.

docs/remote_kernels.md
```diff
@@ -0,0 +1,119 @@
+# Remote kernels
+
```
wlach: 💯  on having docs right out of the gate. :) An introductory paragraph is probably a good idea here.

server/files/models.py
```diff
@@ -1,35 +1,57 @@
...
+        blank=True, null=True, help_text="The datetime when the file was last refreshed at"
+    )
+    remote_operation = models.ForeignKey(
+        RemoteOperation, null=True, blank=True, on_delete=models.SET_NULL
+    )
+
+    class Meta(BaseFile.Meta):
         verbose_name = "File"
         verbose_name_plural = "Files"
-        ordering = ("id",)
```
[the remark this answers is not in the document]
jezdez: Just moving it to a common abstract model class in https://github.com/iodide-project/iodide/pull/2219/files#diff-0a01a30df9908c89b347ea5b81c5b58eR27, not taking it out.

.coveragerc
jezdez: Sure thing, this is mostly developed while working on the feature, but I can split it off if you'd like.

.circleci/config.yml
jezdez: "leantest" is what we discussed a while ago that disables some pytest plugins in CI: https://github.com/iodide-project/iodide/pull/2219/files#diff-6b435363a6346c7164cb2f6e347ec389 Basically if you run `pytest --lean` it'll disable pytest-flake8, pytest-isort and pytest-coverage. Since you've said in our discussion that you dislike having all that as part of the CI run and since there is already an overlap with the lint scripts I added that ability. That means running `make test` during development will actually use flake8, isort and coverage.

docs/remote_kernels.md
jezdez: Yeah, this isn't done, good idea adding an introduction and maybe add that this is work in progress (e.g. see @bcolloran's recent proposal around remote language plugins).

jezdez:
> This is looking really good! I have a few preliminary comments:
> 
>     * For a proper review, I think it would be good to see the working plugin for redash somewhere. Either that or...

Yeah, that's my next step for the query backend, making the actual request to Redash, I also have some janky code for the client side but I'm less excited about that.

>     * ... we could consider creating a "demo" remote kernel plugin that does something useful on the public server both for illustration and validation purposes. Maybe we could use the [github graphql API](https://developer.github.com/v4/)? We already have people's github authentication information on the public server...

Hm, I'd rather not open a new construction site at the moment and focus on the PoC query backend for Redash.

>     * I think we need/should keep track of the user requesting the remote operation.

Yeah, I've been pondering that too, that would allow us doing interesting things in the future, such as sending over the username to Redash and auto-create a user account there to track queries per user, instead of using a catch-all user as we intended for the PoC. I don't recall if I discussed that with Rob or you before though.

.coveragerc
jezdez: Opened https://github.com/iodide-project/iodide/pull/2305.

.circleci/config.yml
wlach: awesome, makes sense!

PR #2244, opened 2019-09-13
upgrade to emotion 10 / refactor to accomodate the new API

hamilton:
### Pull Request checklist

- [-] **Documentation**: (PR does not require docs)
- [-] **Changelog**: This PR updates the [changelog](https://github.com/iodide-project/iodide/blob/master/CHANGELOG.md) with any user-visible changes.
- [-] **Tests**: This PR includes thorough tests or an explanation of why it does not

This PR effectively does what #1795 hints at – to actually upgrade to Emotion v10, we have to use the new namespacing, install the additional packages, and remove the deprecated `css` functionality where relevant. This PR does all that. 

There are a _lot_ of files that don't need to be reviewed. I would just check to see that nothing looks out of the ordinary via manual testing, and perhaps check `package.json` to see if there was anything I missed (I took a look myself but I think it's fine).

fwiw, this does not fix all the warnings in the client, but it does look like it boils it down to one, vs. the 4-5 from before.

Probably only need 1 of y'all to review this one.


bcolloran:
thanks for tackling this @hamilton!

PR #2246, opened 2019-09-13
adds a notebook files API + integrates into client

### LINKED ISSUE #2200: add a client-centric files API and have the manage files modal use it
This is a follow-on improvement to #1985. 

Currently, a notebook only receives a list of files from the server on page load. This caused some inefficiencies while authoring the original files modal, which ended up managing a lot of local state to keep up with what was added to the server and what wasn't.

With file sources, however, it could be that a file gets added to a notebook (for instance, clicking the "download now" button). The files list does not necessarily update automatically.

This will be an easy fix. We just need to copy the same pattern in the file sources API for files – when the modal appears or the modal tab is clicked, there should be a request to the server for the list of files.

Adding this API will also _drastically_ simplify the UX for the manage files tab. The specific UX refactor that needs to happen, however, should probably be tackled separately.

hamilton:
fixes #2200. I opted for the most minimal set of changes here, which is to say the `getFiles` action ~gets dispatched on notebook load only~ dispatches on page load, when the file modal opens, and after a manual file update operation successfully finishes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3645,6 +3645,7 @@
       "version": "9.2.11",
       "resolved": "https://registry.npmjs.org/babel-plugin-emotion/-/babel-plugin-emotion-9.2.11.tgz",
       "integrity": "sha512-dgCImifnOPPSeXod2znAmgc64NhaaOjGEHROR/M+lmStb3841yK1sgaDYAYMnlvWNz8GnpwIPN0VmNpbWYZ+VQ==",
+      "dev": true,
       "requires": {
         "@babel/helper-module-imports": "^7.0.0",
         "@emotion/babel-utils": "^0.6.4",
@@ -4604,7 +4605,7 @@
     },
     "callsites": {
       "version": "2.0.0",
-      "resolved": "http://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/callsites/-/callsites-2.0.0.tgz",
       "integrity": "sha1-BuuE8A7qQT2oav/vrL/7Ngk7PFA="
     },
     "camel-case": {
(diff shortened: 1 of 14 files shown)


package-lock.json
```diff
@@ -3645,6 +3645,7 @@
       "version": "9.2.11",
```
wlach: I think this got changed by mistake

server/files/api_views.py
```diff
@@ -61,6 +61,25 @@ def update(self, request, pk):
...
+    serializer_class = FilesSerializer
+
+    def get_serializer_context(self):
+        notebook_id = int(self.kwargs["notebook_id"])
+        if not Notebook.objects.filter(id=notebook_id).exists():
+            raise Http404("Notebook with id %s does not exist" % notebook_id)
+        return {"notebook_id": notebook_id}
+
+    def get_queryset(self):
+        base = File.objects.filter(notebook_id=self.kwargs["notebook_id"])
```
wlach: I would suggests `files` as a variable name instead of `base`

src/editor/index.jsx
```diff
@@ -43,6 +44,7 @@ window.addEventListener("message", listenForEvalFramePortReady, false);
 handleServerVariables(store);
 handleInitialIomd(store);
 store.dispatch(restoreLocalAutosave());
+store.dispatch(getFiles());
```
wlach: Is this something we currently need to do on startup?

I feel like it's worth the extra bit of effort to make this happen when the files modal is opened (or a file source is downloaded), and eliminate the line here. I know this is extra work, but this PR doesn't actually fix the bug in its current form.

hamilton: The purpose of this PR was not to fix the fact that the files modal doesn't make the request, but to prepare for refactoring that tab to do what the file sources tab does, which will require more cycles than I have. I'm happy to add the requesting but there is more UX work than you'd think to make the experience smoother.

hamilton: On second thought, sure, I'll add these in those spots for now. I think the UX weirdness is probably worth having those update.

hamilton: Also to clarify – we still need to dispatch `getFiles` for the programmatic API to work properly, so I'm going to keep this here.

hamilton: You make a good point – this PR should actually do the work necessary to close #2200, which isn't much additional code.

bcolloran:
looks great!

PR #2252, opened 2019-09-16
Feature Intercept Backspace (Fixes #1725) (Extends PR #1779)

### LINKED ISSUE #1725: [Feature Request] intercept backspace button press outside editor

Algentile:
Warn user of back navigation if backspace key bind is used. This PR extends the out of date PR #1779 and fixes #1725. Test notebook is hosted [https://local-iodide.herokuapp.com](https://local-iodide.herokuapp.com).

### Pull Request checklist


teonbrooks:
Hi @Algentile, I've just tried this out:
- if I have clicked into the console panel, this intercept works correctly
- if I am in report view, this intercept works correctly
- but in the explore mode if I click into the report view panel, it doesn't intercept

Algentile:
Hey @teonbrooks,

I added the initialization function to eval frame to intercept on editor / eval / and report view. A few other things I need to do is intercept on icon select as well as on root-modal, I deployed all the changes to the heroku instance listed above.

teonbrooks:
I tested this and it seems that most the backspace intercepts work except for the following scenario (and it might be edge case enough to not worry about it):


>     * in the explore mode if I click into the report view panel, it doesn't intercept

src/eval-frame/index.jsx
```diff
@@ -20,6 +20,7 @@ import messagePasserEval from "../shared/utils/redux-to-port-message-passer";
 import { iodide } from "./iodide-api/api";
 
 import "./port-to-editor";
+import { initializeDefaultKeybindings } from "../editor/initialization/keybindings";
```
bcolloran: can you move the relevant code for intercepting the backspace to a new file in `src/shared` and import that code here and in `src/editor`?

the reason for doing so is that importing `../editor/initialization/keybindings` into the eval-frame causes the import of
`import tasks from "../user-tasks/task-definitions"` from the editor side, which imports `actions` from the editor side, etc. in this ase, it would just import some extra code i think, but we try to maintain import hygiene by putting code that will be run in both the editor and eval-frame into `src/shared` because in some cases importing things in both sides of the app can cause conflicts and weird bugs.

thanks @Algentile ! :-)

src/shared/keybindings.js
```diff
@@ -0,0 +1,28 @@
+import Mousetrap from "mousetrap";
+import tasks from "../editor/user-tasks/task-definitions";
```
bcolloran: sorry @algentile, i should have been clearer in describing the objective -- regarding importing files between the folders `src/shared`, `src/editor`, and `src/eval-frame` we want to make sure that:

1.  no file in `src/editor` imports anything in `src/eval-frame`
2.  no file in `src/eval-frame` imports anything in `src/editor`
3.  no file in `src/shared` imports anything in `src/editor` or  `src/eval-frame`
4.  but files in  `src/editor` and  `src/eval-frame` may import from `src/shared`

you can think of  `src/editor` and  `src/eval-frame` as separate apps, so this import hygiene prevents us from accidentally pulling in a web of imports from one of the apps that will not be needed (or could even cause problems) in the other app.

i'd recommend putting the `initializeDefaultKeybindings` function back where it was in `src/editor`, and adding an additional file somewhere in `src/shared` that is just responsible for binding delete and backspace and handling/preventing back nav

does this make sense?

cc @wlach

Algentile: Hey @bcolloran @wlach, 

I moved keybindings back to initialization in editor and exported the intercept function from shared. The change should be redeployed in heroku.

Algentile:
I was able to trigger a code coverage report, but I had to add an empty commit as it appears that code coverage will still fail with 400 on a merge commit. We should reference this issue thread for tracking [codecov fails on merge commit](https://github.com/codecov/codecov-bash/issues/134).

wlach:
> I was able to trigger a code coverage report, but I had to add an empty commit as it appears that code coverage will still fail with 400 on a merge commit. We should reference this issue thread for tracking [codecov fails on merge commit](https://github.com/codecov/codecov-bash/issues/134).

This is actually due to the secret not being shared to forked PRs: see #2330 and #2332 for the gory details. In any case, if you rebase/merge against master this should be fixed.

bcolloran:
this looks great, thanks @Algentile :champagne: 

@teon does this fix #1725 to your satisfaction? if you can do a quick QA, i'll merge if you are happy.

teonbrooks:
just QA'd it. works great. thanks @Algentile!

PR #2271, opened 2019-09-18
Check latest local autosave against latest save (fixes #2259)

### LINKED ISSUE #2259: Notebook history does not hide "Unsaved Changes"
With the latest changes, expected behavior for notebook history is to hide "Unsaved Changes" when there are no differences. However, this does not hold in a case.

# Steps to Reproduce
1. Create a new notebook
2. In an empty line, press enter and delete
3. "Unsaved Changes" with "No difference." shows up
4. Page refresh fixes

mehmetoguzderin:
Fix for https://github.com/iodide-project/iodide/issues/2259

# Explanation
The current implementation shows unsaved changes whenever local autosave entry is created, which is never reduced back when there are no changes (e.g., create a new line and remove it immediately). This new implementation further checks the local autosave with the latest revision to ensure that there are no changes.

### DIFF
--- a/src/editor/actions/history-modal-actions.js
+++ b/src/editor/actions/history-modal-actions.js
@@ -2,7 +2,7 @@ import {
   getRevisionList,
   getRevisions
 } from "../../shared/server-api/revisions";
-import { getLocalAutosave, haveLocalAutosave } from "../tools/local-autosave";
+import { getLocalAutosave } from "../tools/local-autosave";
 import { getRevisionIdsNeededForDisplay } from "../tools/revision-history";
 import { getNotebookID, isLoggedIn } from "../tools/server-tools";
 import { updateAutosave } from "./autosave-actions";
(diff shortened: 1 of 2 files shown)


src/editor/actions/history-modal-actions.js
```diff
@@ -70,13 +70,28 @@ export function getNotebookRevisionList() {
...
+        getLocalAutosave(getState())
+          .then(localAutosaveState => {
+            let hasLocalOnlyChanges = false;
+            if (
+              "revisionContent" in getState().notebookHistory &&
+              "iomd" in localAutosaveState
+            ) {
+              if (
+                revisionList[0].id in getState().notebookHistory.revisionContent
+              ) {
```
wlach: Looking at this again, we could just rewrite this in a single expression (I know the formatting is wrong):

```js
const hasLocalOnlyChanges = ("iomd" in localAutosaveState  && 
  "revisionContent" in getState().notebookHistory && 
  revisionList.length &&
  revisionList[0].id in getState().notebookHistory.revisionContent &&                   
  localAutosaveState.iomd !== getState().notebookHistory.revisionContent[revisionList[0].id]);
```

This also checks for the edge case where the revision list has a length of 0 (should never happen but who knows).

wlach:
Thank you!

Issue #2275, opened 2019-09-19
migrate diff viewer to use monaco's diff viewer


bcolloran:
monaco editor comes with the diff viewer used by vscode. this is obviously very actively maintained and quite polished, and migrating would also remove a dependency which is always nice. we should migrate to it when we can.

bcolloran:
little more on this: i'm not sure how involved this will be; it might be trick or it might be pretty easy. we'd need to create a custom component to instantiate the monaco diff viewer, along the lines of the [editor component](https://github.com/iodide-project/iodide/blob/master/src/editor/components/iomd-editor/iomd-editor.jsx) but i think waaaay simpler since the state updating would not need to be so sophisticated -- i think we could write it as it's own component that could just take the same props as `ReactDiffViewer` and have it be a drop-in replacement, but i'm not 100% sure.

wlach:
FWIW, our usage of react-diff-viewer is very very simple (as is the component):

https://github.com/iodide-project/iodide/blob/dbfa9999eb48f989288830d3e399520cc743983b/src/editor/components/modals/revision-diff.jsx#L67

Issue #2260, opened 2019-09-17
Notebook history does not handle single new line


mehmetoguzderin:
Revision browser does not show one new empty line as difference. The new line is detected correctly, saved but the difference is not visible in the revision browser. There are no problems with more than one empty new lines.

# Steps to Reproduce
- Create a new notebook
- Add a single new line anywhere
- Check notebook history

wlach:
This sounds like a bug in `react-diff-viewer`, if you wanted to provide a test case and a PR there, we could pick up the fix on their next release.

bcolloran:
@wlach @mehmetoguzderin created a new issue #2275 about moving off of `react-diff-viewer`. just posting here as an fyi since you guys are looking at the diff viewer stuff. @mehmetoguzderin if this is of interest to you, you'd be very welcome to give it a shot.

mehmetoguzderin:
> @wlach @mehmetoguzderin created a new issue #2275 about moving off of `react-diff-viewer`. just posting here as an fyi since you guys are looking at the diff viewer stuff. @mehmetoguzderin if this is of interest to you, you'd be very welcome to give it a shot.

I think I can try it out!

PR #2298, opened 2019-09-25
Use `DiffEditor` of `monaco` (fix for #2275 and #2260)

mehmetoguzderin:
Aims to solve [#2275](https://github.com/iodide-project/iodide/issues/2275) and [#2260](https://github.com/iodide-project/iodide/issues/2260).

This pull request migrates generation of content for `revision-diff` from `ReactDiffViewer` to `DiffEditor` provided by `monaco` itself.

`revision-diff-content` serves as a non-redux wrapper for abstracting details of `monaco` away. Further connections such as theming should be communicated as properties from `revision-diff` to `revision-diff-content` in render function.


wlach:
This looks great @mehmetoguzderin! I'm not sure why the unit tests are failing though.

My only other comment is that the initial revision diff looks a little strange (it claims a line is deleted at the top): 
image

mehmetoguzderin:
> This looks great @mehmetoguzderin! I'm not sure why the unit tests are failing though.
> 
> My only other comment is that the initial revision diff looks a little strange (it claims a line is deleted at the top):
> image

Thanks!

It is strange indeed and have been noticed by others too:
https://github.com/Microsoft/monaco-editor/issues/630

Maybe we can specialize in the initial revision to solve this after we fix the unit tests issue.

wlach:
@mehmetoguzderin do you want to try rebasing this against latest master? The monaco upgrade in #2273 may have either fixed this or made the test error output more intelligible, if we're lucky.

PR #2314, opened 2019-09-30
Increase max length of file source urls from 200 to 8192

wlach:
Some URLs that people use can get quite long, so best to increase
the max amount here.

This can affect long bugzilla queries, like those used by @teonbrooks on itmo.

### Pull Request checklist

- [N/A] **Tests**: This PR includes thorough tests or an explanation of why it does not


wlach:
> LGTM, but I think this requires a new migration, right?

Totally does, I forgot to add it! Thanks for checking

PR #2332, opened 2019-10-04
Don't use CODECOV_TOKEN environment variables (fixes #2330)

### LINKED ISSUE #2330: Forked PR's fail CI
Since we enabled codecov, it looks like forked PRs are now failing the test-server job with this:

```
-> Pinging Codecov
https://codecov.io/upload/v4?package=bash-8a28df4&token=secret&branch=pull%2F2329&commit=e84326b038576f73971c632c3561be13ff94b607&build=&build_url=&name=&tag=&slug=iodide-project%2Fiodide&service=&flags=&pr=&job=
HTTP 400
Please provide the repository token to upload reports via `-t :repository-token`
Exited with code 1
```

See e.g. #2329

I think this is because the `CODECOV_TOKEN` variable is private to iodide. There is an option in circle [to pass secrets from forked PRs](https://circleci.com/docs/2.0/oss/#pass-secrets-to-builds-from-forked-pull-requests), but we probably don't want to blanket enable that: I don't know how sensitive the `CODECOV_TOKEN` is, but there are dockerhub credentials which we definitely don't want to leak here.

@jezdez do you have any ideas on how to proceed? [This page](https://docs.codecov.io/docs/about-the-codecov-bash-uploader) suggests that we might not even need a token?

image

wlach:
It should not be required for public projects like Iodide.

### Pull Request checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,7 @@ jobs:
             # we need to re-collect static as docker-compose mounts the checkout directory over what we have in the image
             chmod 777 . # make sure we have permission to write the files in our volume
             # run the actual tests
-            docker-compose run --rm $(CI_ENV) server ci
+            docker-compose run --rm $(CI_ENV) -e CI -e CIRCLECI -e CIRCLE_BRANCH -e CIRCLE_BUILD_NUM -e CIRCLE_NODE_INDEX -e CIRCLE_PR_NUMBER -e CIRCLE_SHA1 server ci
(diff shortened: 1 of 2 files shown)


wlach:
Turns out we also need to import the circle configuration into the docker environment for the right thing to happen. This is pretty unfortunate and probably brittle -- wonder if there are better options. In any case I'm going to merge this now so at least future PRs will start passing.

Issue #2175, opened 2019-08-20
Add ability for language plugins to auto-evaluate on chunk change?


wlach:
I've been tweaking a mermaid diagram in this notebook today:

https://definitely-staging.iodide.io/notebooks/60/

It's a bit annoying to have to constantly re-evaluate the cell as I make changes, is there any way we could allow language plugins to do the same thing as markdown cells -- that is, re-evaluate the content on every change?

@bcolloran @mdboom ?

mdboom:
Seems like a nice idea.  Obviously it should be a flag, because you wouldn't want this for something stateful like Python.

wlach:
I don't have an exact recipe for how to do this, but I think this issue wouldn't be that difficult for a motivated contributor to take on if they are willing to get their hands dirty and spelunk a bit through the codebase. 

The place to start would be probably be to consult the existing documentation for how language plugins work:

https://iodide-project.github.io/docs/language_plugins/

From there, I would dive into the codebase and see how markdown chunks are auto-evaluated, and see if we can't make language plugins for particular cells have the same behavior.

mehmetoguzderin:
I think this is highly related to issue [2265](https://github.com/iodide-project/iodide/issues/2265) since both aim to shorten artistic iteration times while keeping intuitive. Maybe these two could be tracked under one roof with more generalized objectives and exposed hooks to language plugins.

PR #2376, opened 2019-10-21
Simplify report pane logic and add tests

wlach:
Before this change, any change to an md/css cell would invalidate
all non-markdown cells (e.g. cells generated by a language plugin
output renderer). With this change, those changes are kept intact
while markdown/css changes are incorporated into the report.

### Pull Request checklist


wlach:
@bcolloran is there any reason not to do things this way? I based this approach on this document:

https://reactjs.org/docs/integrating-with-other-libraries.html

It seems to work great for notebooks like this, where you have one cell outputting some stuff in response to a language plugin, with explanatory markdown around it: https://definitely-staging.iodide.io/notebooks/168/

Without this change you are constantly having to re-run the language plugin cell as you edit text. This is not quite as good as #2175 but gets us a little closer.

wlach:
Went ahead and added some jest tests for this piece while I was at it.

bcolloran:
@wlach this code looks fine to me, but i'm confused about the behavior --
when i edit the md or css in this notebook on release the mermaid chart is not affected: https://alpha.iodide.io/notebooks/3470/

i don't seem to have to re-run the `mmd` chunk when i edit the css or md in either your notebook on staging or the the notebook on alpha, so i'm confused that maybe i'm missing something?

wlach:
I think I misunderstood the code here, and it turns out there is not an actual bug. But the simplifications here are still worthwhile. I'll remove the entry from the changelog and change the commit message and just land it.

PR #2382, opened 2019-10-22
Add codecov config to adjust target/threshold and disable patch track…

jezdez:
…ing.

### Pull Request checklist

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,28 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  # 50...75 results in any coverage less than 50% having a red background.
+  # The color would gradually change to green when approaching the 75% mark.
+  # Any coverage over 75% would result in a solid green color.
+  range: 50..75
+
+  status:
+    project:
+      default:
+        # the minimum coverage ratio that the commit must meet to be considered a success
+        target: 50%
+        # percentage to allow the coverage to drop by,
+        # without poting an error status
+        threshold: 1%
+
+comment:
+  # reach: coverage graph, diff: coverage diff to base, tree: list of files
+  layout: "reach,diff,tree"
+  # update, if exists. Otherwise post new. Skip if deleted.
+  behavior: once
+  # if yes, only post the comment if coverage changes
+  require_changes: no


.codecov.yml
```diff
@@ -0,0 +1,18 @@
...
+
+  status:
+    project:
+      default:
+        target: 90%
+        threshold: 1%
+    patch: no
+    changes: no
+
+comment: off
```
wlach: I feel like we probably want the comment

.codecov.yml
```diff
@@ -0,0 +1,18 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: 50..100
```
wlach: Any reason for these values in particular? Can you add a comment?

.codecov.yml
```diff
@@ -0,0 +1,18 @@
...
+
+  status:
+    project:
+      default:
+        target: 90%
+        threshold: 1%
```
wlach: Is there any reason for setting these thresholds if we're not actually going to post a status?

.codecov.yml
jezdez: Ah, ok!

.codecov.yml
jezdez: Yeah, in fact I'll change that to 50..75 meaning that anything less than 50% is considered bad and is marked as red, and the more we approach 75% the greener the graphs will get. I don't think we should only make 100% green since that implies that only 100% test coverage is "healthy".

Test coverage can be a misleading metric, not saying anything about the quality of the tests and sometimes even encouraging unrealistic and useless tests. Treating 75% as a goal is a good compromise that encourages test writing but staying in the realm of encouragement instead of penalization.

.codecov.yml
jezdez: Yeah, I fixed the reporting on changes and patch coverage.

.codecov.yml
```diff
@@ -0,0 +1,28 @@
...
+  # Any coverage over 75% would result in a solid green color.
+  range: 50..75
+
+  status:
+    project:
+      default:
+        # the minimum coverage ratio that the commit must meet to be considered a success
+        target: 50%
+        # percentage to allow the coverage to drop by,
+        # without poting an error status
```
wlach:
```suggestion
        # without posting an error status
```

jezdez: "poting" *snicker*

PR #2402, opened 2019-10-30
Use webpack production mode for building iodide (fixes #2023)

### LINKED ISSUE #2023: build tune-ups (webpack not shaking tree correctly / not reading NODE_ENV options correctly)
we should set our "build" mode to be a "production" mode (which will have minification and tree shaking and will prevent some debug messages from being surfaced in the devtools console), and deprecate the "build-prduction" mode, which we don't seem to be using anywhere.

thanks @dmose for noticing this.

cc @wlach

wlach:
Also, use terser-js (uglify doesn't support some ES6 constructs we now use).
This shaves over a megabyte off of the uncompressed file sizes for our
js bundles.


wlach:
I noticed other problems and cruft in our webpack config while doing this (e.g. our use of a ".dev" postfix on all our files is no longer necessary), but I'll file those as separate PRs after this lands in the interests of keeping things easy-to-review.

wlach:
Got a verbal r+ from @bcolloran, gonna merge

PR #2406, opened 2019-11-01
Remove APP_VERSION_STRING and all its use from all generated webpack files

wlach:
Originally envisioned as a way to "version" iodide client side code,
it is not useful in Iodide's current incarnation. We may want to bring
something like it back in the future, but it will probably take a different
form.


wlach:
@hamilton this is basically a followup to #2402, just some more opportunity for cleanup that I noticed.

bcolloran:
LGTM @wlach , r+ merge when you wish

PR #2461, opened 2019-11-20
Automatically prune revision history (fixes #2451)

### LINKED ISSUE #2451: New revisions should be marked as "draft", and should automatically prune drafts
Currently we save new revisions of each iodide notebook every 30 seconds: 

https://github.com/iodide-project/iodide/blob/6f3d61d8f559e27609140cc3f4e3ba0877b69eee/src/editor/actions/autosave-actions.js#L6

This leads to a lot of redundant revisions in notebooks with even a modest amount of history, which can make the history viewer hard to use. We should probably put a cap on the number of revisions we save to one per minute -- this would also let us save even more frequently.

I propose doing this by adding "draft" revisions. Draft revisions are subject to being deleted/replaced until they are at least a minute old.

Proposal:
* Add a new boolean field "is_draft" to the notebook revision model. Need to change something around here (you'll also need to create django migrations here, try googling "django migrations"): https://github.com/iodide-project/iodide/blob/6a9c68fb11e05f49accc0e5e2fa1cf15653ab7e3/server/notebooks/models.py#L36
* New revisions should have "is_draft=True" by default. Probably need to change something around here: https://github.com/iodide-project/iodide/blob/master/server/notebooks/api_views.py#L128
* Every time a revision is submitted, schedule a periodic task that does the following:
** Get a list of all "draft" revisions
** Make sure that all draft revisions are spaced at least one minute apart. If there any draft revisions within the same minute interval, delete them.
** After this process is complete, mark any draft revisions more than a minute old as being non-draft.
* Write some unit tests to make sure this is working as expected. You can see some existing ones under `server/tests/`: https://github.com/iodide-project/iodide/tree/master/server/tests

This is not a comprehensive set of things that need to be done but should be enough to get started. :)

zzl0:
### Pull Request checklist


zzl0:
The CI didn't pass, I will fix it later today.

wlach:
Thanks @zzl0! will take a look soon

server/notebooks/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+                revision.save()
+
+    def delete(notebook_revisions, now_utc, threshold_in_mins=TIME_INTERVAL_IN_MINS):
+        "Delete the revision if it is older than threshold_in_mins."
+        for revision in notebook_revisions:
+            if now_utc - revision.created > timedelta(minutes=threshold_in_mins):
+                revision.delete()
+
+    # ordered_draft_revisions is ordered by `created` time desc.
+    ordered_draft_revisions = NotebookRevision.objects.filter(
```
wlach: I think we can just call these `draft_revisions`

server/notebooks/tasks.py
```diff
@@ -0,0 +1,46 @@
+import pytz
+from datetime import datetime, timedelta
+from collections import defaultdict
+from spinach import Tasks
+
+from .models import NotebookRevision
+
+TIME_INTERVAL_IN_MINS = 1
```
wlach: Let's specify this in seconds, just for a bit more flexibility in the future. I'd also recommend defining this in settings.py, and making it configurable. Maybe call it something like `NOTEBOOK_SPACING_INTERVAL_SECS`

server/tests/test_notebook_revisions_cleanup.py
```diff
@@ -0,0 +1,82 @@
...
+@pytest.fixture
+def notebook_and_draft_revisions(fake_user):
+    notebook = Notebook.objects.create(owner=fake_user, title="Fake notebook")
+    notebook_revisions = []
+
+    for i, created in enumerate(CREATED_DATETIMES):
+        revision = NotebookRevision.objects.create(
+            notebook=notebook,
+            title="Revision %s" % i,
+            content="fake notebook content for revision %s" % i,
```
wlach:
```suggestion
            content=f"fake notebook content for revision {i}",
```

server/tests/test_notebook_revisions_cleanup.py
```diff
@@ -0,0 +1,82 @@
...
+
+@pytest.fixture
+def notebook_and_draft_revisions(fake_user):
+    notebook = Notebook.objects.create(owner=fake_user, title="Fake notebook")
+    notebook_revisions = []
+
+    for i, created in enumerate(CREATED_DATETIMES):
+        revision = NotebookRevision.objects.create(
+            notebook=notebook,
+            title="Revision %s" % i,
```
wlach:
```suggestion
            title=f"Revision {i}",
```

server/notebooks/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+    # ordered_draft_revisions is ordered by `created` time desc.
+    ordered_draft_revisions = NotebookRevision.objects.filter(
+        notebook_id=notebook_id,
+        is_draft=True)
+    revision_groups = group_by_time_interval(ordered_draft_revisions)
+    now_utc = now_utc or datetime.now(tz=pytz.utc)
+
+    for group in revision_groups:
+        clear_draft_flag(group[:1], now_utc)
+        delete(group[1:], now_utc)
```
wlach: I think it would be faster to just clear the draft flags in one operation at the end, rather than group-by-group. In that case you could simply filter for (remaining) revisions whose age is greater than the interval, and delete them in one go.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Allow running simple-serve from a port other than 8000 (fixes #2444)
 - Allow blank content on notebook revision creation (fixes #2445) (#2455)
 - Enable user to clear console history with a click (#1015)
+- Automatically prune revision history (#2451)
```
```suggestion
- Automatically prune revision history (fixes #2451) (#2461)
```

docs/key_concepts.md
```diff
@@ -29,7 +29,7 @@ As you go, you can also evaluate a [IOMD](iomd.md) code chunks pressing `ctrl+en
...
 
-When editing an iodide notebook, the content of your notebook should be saved at a regular interval (currently 30 seconds) as you type. The server has a collection of all the revisions explicitly saved to the during your session. To access the revisions for a notebook, go to the landing page and hover over a notebook of interest. You will see the option `revisions` that you can click, which will take you the history of revisions for that notebook.
+When editing an iodide notebook, the content of your notebook should be saved at a regular interval (currently 30 seconds) as you type. In order to reduce the number of old revisions, server side also prunes the old revisons to one per minute. The server has a collection of all the revisions explicitly saved to the during your session. To access the revisions for a notebook, go to the landing page and hover over a notebook of interest. You will see the option `revisions` that you can click, which will take you the history of revisions for that notebook.
```
wlach: I think we can skip the extra explanation here, I don't think this is super important from a user's perspective.

```suggestion
When editing an iodide notebook, the content of your notebook should be saved at a regular interval (currently 30 seconds) as you type. The server has a collection of all the revisions explicitly saved to the during your session. To access the revisions for a notebook, go to the landing page and hover over a notebook of interest. You will see the option `revisions` that you can click, which will take you the history of revisions for that notebook.
```

server/notebooks/models.py
```diff
@@ -43,6 +43,7 @@ class NotebookRevision(models.Model):
     title = models.CharField(max_length=Notebook.MAX_TITLE_LENGTH)
     created = models.DateTimeField(auto_now_add=True)
     content = models.TextField(blank=True)
+    is_draft = models.BooleanField(default=True)
```
wlach: I think perhaps we should not set a default for this field, to force any person using this class to think about what type of revision they're adding. Also, when creating the migration, any currently-existing revisions should *not* be considered draft.

server/tests/test_notebook_revisions_cleanup.py
```diff
@@ -0,0 +1,82 @@
...
+    return notebook, notebook_revisions
+
+
+def test_execute_notebook_revisions_cleanup(notebook_and_draft_revisions):
+    notebook, revisions = notebook_and_draft_revisions
+
+    # execute
+    execute_notebook_revisions_cleanup(notebook.id, NOW_UTC)
+
+    # verfiy
```
wlach:
```suggestion
    # verify
```

server/tests/test_notebook_revisions_cleanup.py
zzl0: I used `f-string`, but current lint complained about it. I didn't plan to fix this lint problem in this PR.

server/tests/test_notebook_revisions_cleanup.py
zzl0: same as above

docs/key_concepts.md
zzl0: sure

server/notebooks/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+        for revision in notebook_revisions:
+            if now_utc - revision.created > timedelta(minutes=threshold_in_mins):
+                revision.is_draft = False
+                revision.save()
+
+    def delete(notebook_revisions, now_utc, threshold_in_mins=TIME_INTERVAL_IN_MINS):
+        "Delete the revision if it is older than threshold_in_mins."
+        for revision in notebook_revisions:
+            if now_utc - revision.created > timedelta(minutes=threshold_in_mins):
+                revision.delete()
```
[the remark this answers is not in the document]
zzl0: thanks, I understand multiple transactions are slow. I thought in current settings (30s in UI, 1 minute interval in backend), normally only one revision need to be deleted. So I didn't try to optimize here. But since in the future, we might want to have more frequent drafts, so it might worth to add this now.

I also thought about create a partial index on `is_draft`, because usually draft revisions are just a small part of all revisions, but I prefer to do that after we have some performance data.

server/notebooks/tasks.py
zzl0: Sure

server/notebooks/tasks.py
zzl0: Sure

server/notebooks/tasks.py
zzl0: Sounds good, thanks.

zzl0:
Thank you @wlach for reviewing it, I will address the comments and send a new revision.

server/notebooks/models.py
zzl0: @wlach do you mean declare `is_draft` field as `is_draft = models.BooleanField()` and change JavaScript code to pass this parameter explicitly? Also, when creating the migration, Use `False` as a one-off default value for existing rows.

wlach: > @wlach do you mean declare `is_draft` field as `is_draft = models.BooleanField()` and change JavaScript code to pass this parameter explicitly? 

Yes, to not specifying a default, no to providing an interface to setting it via JavaScript / the web api (at least not yet). You may need to modify the django rest framework code to set `is_draft` is false when creating the actual item, in this file:

https://github.com/iodide-project/iodide/blob/master/server/notebooks/api_views.py

Let me know if you need any help with this.

> Also, when creating the migration, Use `False` as a one-off default value for existing rows.

Yup! :+1:

zzl0: Thanks for the explanation! 

> Let me know if you need any help with this

I am not familiar with Django REST Framework, but I think I can figure it out.

server/notebooks/api_views.py
```diff
@@ -147,3 +148,4 @@ def perform_create(self, serializer):
                 )
 
         serializer.save(**ctx)
+        execute_notebook_revisions_cleanup(ctx["notebook_id"])
```
jezdez: This should use the schedule API and not call the function itself otherwise it'll just be called in-process, e.g.:
```python
        tasks.schedule(execute_notebook_revisions_cleanup, ctx["notebook_id"])        
```
Here's an example:
https://github.com/iodide-project/iodide/blob/1891a5995e0e8a6bd318026e2988a207d7b71641/server/files/api_views.py#L150

zzl0: Thanks

zzl0:
@wlach I think I resolved all the comments. I am not familiar with this `changes requested` feature of GitHub (I have added those requested changes manually).

server/notebooks/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+    ]
+    NotebookRevision.objects.filter(id__in=intermediate_revision_ids).delete()
+
+    # mark old revisions as non-draft
+    non_draft_revision_ids = [
+        revision_window[0].id
+        for revision_window in revision_windows
+        if now_utc - revision_window[0].created >= threshold
+    ]
+    NotebookRevision.objects.filter(id__in=non_draft_revision_ids).update(is_draft=False)
```
wlach: Instead of getting a list, ry this instead:

```py    
NotebookRevision.objects.filter(id__in=draft_revisions, created__lt=now_utc - threshold).update(is_draft=False)
```

server/notebooks/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # group notebook revisions into fixed-size windows
+    revision_groups = defaultdict(list)
+    for revision in draft_revisions:
+        key = int(revision.created.timestamp()) // NOTEBOOK_REVISION_FIXED_WINDOW_LENGTH_SECS
+        revision_groups[key].append(revision)
+    revision_windows = [revision_groups[k] for k in sorted(revision_groups.keys(), reverse=True)]
+
+    threshold = timedelta(seconds=NOTEBOOK_REVISION_FIXED_WINDOW_LENGTH_SECS)
+    # delete the intermediate revisions
```
wlach:
```suggestion
    # delete intermediate revisions
```

server/notebooks/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+    intermediate_revision_ids = [
+        revision.id
+        for revision_window in revision_windows
+        for revision in revision_window[1:]
+        if now_utc - revision.created >= threshold
+    ]
+    NotebookRevision.objects.filter(id__in=intermediate_revision_ids).delete()
+
+    # mark old revisions as non-draft
```
wlach:
```suggestion
    # mark remaining old revisions as non-draft
```

server/notebooks/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+@tasks.task(name="notebooks:execute_notebook_revisions_cleanup")
+def execute_notebook_revisions_cleanup(notebook_id, now_utc=None):
+    """Prue revision hisotry.
+
+    Prune revision history so that there is only one revision in each
+    time window, also marking draft revisions that older than window length
+    as non-draft.
+
+    * Time window: This task groups draft revisions into fixed-size windows
+      (alsocalled Tumbling windows).
```
wlach: I didn't know about tumbling windows. Good catch 😄 I'd probably incorporate this text into the paragraph above, as suggested

server/notebooks/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+
+    threshold = timedelta(seconds=NOTEBOOK_REVISION_FIXED_WINDOW_LENGTH_SECS)
+    # delete the intermediate revisions
+    intermediate_revision_ids = [
+        revision.id
+        for revision_window in revision_windows
+        for revision in revision_window[1:]
+        if now_utc - revision.created >= threshold
+    ]
+    NotebookRevision.objects.filter(id__in=intermediate_revision_ids).delete()
```
wlach: To be consistent with my suggestion below, I'd suggest doing the filtering on time in this statement, rather than above. So something like:

```py
intermediate_revision_ids = [
    revision.id
    for revision_window in revision_windows
    for revision in revision_window[1:]
]
NotebookRevision.objects.filter(id__in=intermediate_revision_ids, created__lt=now_utc - threshold).delete()
```

server/notebooks/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+
+from ..settings import NOTEBOOK_REVISION_FIXED_WINDOW_LENGTH_SECS
+from .models import NotebookRevision
+
+tasks = Tasks()
+
+
+@tasks.task(name="notebooks:execute_notebook_revisions_cleanup")
+def execute_notebook_revisions_cleanup(notebook_id, now_utc=None):
+    """Prue revision hisotry.
```
wlach:
```suggestion
    """Prune revision hisotry.
```

server/notebooks/serializers.py
```diff
@@ -62,4 +62,7 @@ def validate(self, attrs):
     class Meta:
         model = NotebookRevision
         fields = ("id", "title", "created", "content")
-        write_only_fields = "notebook_id"
+        extra_kwargs = {
+            "notebook_id": {"write_only": True},
+            "is_draft": {"write_only": True},
```
wlach: I think we do in fact want to include the "is_draft" field in the serialized representation here. I think `is_draft` a read-only field:

https://www.django-rest-framework.org/api-guide/fields/#read_only

You can set the `is_draft` property when creating the revision around here:

https://github.com/iodide-project/iodide/blob/362eb451675b46268cc36ee7a34feb2139d3c630/server/notebooks/api_views.py#L149

server/settings.py
```diff
@@ -248,3 +248,5 @@
 REDIS_HOST = env.str("REDIS_HOST", default="redis")
 REDIS_URL = env.str("REDIS_URL", default=f"redis://{REDIS_HOST}:6379/1")
 SPINACH_BROKER = RedisBroker(redis.from_url(REDIS_URL, **recommended_socket_opts))
+
+NOTEBOOK_REVISION_FIXED_WINDOW_LENGTH_SECS = 60
```
wlach:
```suggestion
# Spacing for permanently-saved notebook revisions
NOTEBOOK_REVISION_SAVE_INTERVAL = 60
```

server/notebooks/tasks.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Prune revision history so that there is only one revision in each
+    time window, also marking draft revisions that older than window length
+    as non-draft.
```
wlach:
```suggestion
    as non-draft. See also: "tumbling windows"
```

server/notebooks/tasks.py
zzl0: Please correct me if I am wrong here. I understand that we want to [do database work in the database rather than in Python](https://docs.djangoproject.com/en/2.2/topics/db/optimization/#do-database-work-in-the-database-rather-than-in-python). If I am reading data from database, I prefer to put filters to database layer (filter pushdown), so that I will get less data.

But in above case, we already have the data in memory. so I did the filter in Python and used `NotebookRevision.objects.filter(id__in=small_set_of_ids).delete()` instead of `NotebookRevision.objects.filter(id__in=big_set_of_ids, created__lt=now_utc - threshold).delete()`. The assumption here is database is harder to scale than stateless application.

I am fine with both, I don't have strong preference here.

wlach: @zzl0 it really depends on what's going on and where the bottlenecks are. In general performing an operation at the database level is going to be an order of magnitude more efficient than doing so in python code -- though in this case I'm suggesting the change mainly for readability reasons (it's easier to see what's going on in a filter expression than a list comprehension)

server/notebooks/tasks.py
```diff
@@ -0,0 +1,47 @@
...
+        revision.id for revision_window in revision_windows for revision in revision_window[1:]
+    ]
+    NotebookRevision.objects.filter(
+        id__in=intermediate_revision_ids, created__lt=now_utc - threshold
+    ).delete()
+
+    # mark remaining old revisions as non-draft
+    NotebookRevision.objects.filter(id__in=draft_revisions, created__lt=now_utc - threshold).update(
+        is_draft=False
+    )
```
zzl0: This format is suggested by `black`, just to make lint happy.

wlach: Yup, was just trying to suggest what should be there, not how it should look. 😄

server/notebooks/api_views.py
```diff
@@ -145,5 +147,6 @@ def perform_create(self, serializer):
                     f"Based on non-latest revision {parent_revision_id} "
                     f"(expected: {last_revision.id})"
                 )
-
+        ctx["is_draft"] = True
         serializer.save(**ctx)
```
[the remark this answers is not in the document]
wlach: (manually changed this before merging, no need to do anything here)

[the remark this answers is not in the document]
zzl0: `ctx` is a local dict (`get_serializer_context()` always generate a new dict `return {"notebook_id": notebook_id}`), so I thought it might be okay to update it.

zzl0:
Thanks for testing it in stage environment.

> The one thing I noticed is that sometimes it will coalesce into an empty revision. I.e. if you have this series:
> A (non-draft) = aaa
> B (draft) = aab
> C (draft) = aaa

> This code will delete B but keep C, which is no different than A. In this case we should just delete both B and C.

That's a nice finding (How did you find this issue?). Yeah, I didn't think about this case. This is an interesting problem. I don't have a solution off the top of head. 

* I am not sure if we should delete both B and C, or just delete C and keep B.
* Seems we also need to get the newest non-draft revision in the auto prune task.

server/notebooks/api_views.py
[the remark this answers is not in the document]
wlach: Yeah, this is partly just a stylistic preference-- I like to treat stuff returned from other functions as immutable, where possible (imitates having a const keyword).

wlach:
I just filed an issue for the last item: #2517

Feel free to take a look at it and file another PR if you're so inclined! Note that I will be away next week.

zzl0:
> I just filed an issue for the last item: #2517
> 
> Feel free to take a look at it and file another PR if you're so inclined! Note that I will be away next week.

Yes, I am interested in fixing it, but I may not be able to work on it recently (I mean in December)

PR #2463, opened 2019-11-21
Make it more clear what the "go to report" button does

wlach:
Just "report" is a bit ambiguous (is it to file a bug report?). We have enough
space to use a small verb phrase ("view as report"), so let's just do that.


wlach:
Before:

image

After:

image

wlach:
> This looks great @wlach. I think we have a bigger UX problem around both viewing the report & also sharing it, but those UX issues are bigger than this PR. For now, I think this is a good direction

Yup, agreed, lots more to do here.

PR #2469, opened 2019-11-22
Use `.` instead of `$PWD` in docker-compose volume definition

wlach:


wlach:
This is just a trivial cleanup I noticed during the sprint when debugging docker stuff on people's linux machines. I'm just going to go ahead and merge.

PR #2526, opened 2019-12-06
Use a custom escaping function instead of lodash for unescaping iomd

wlach:
This fixes one edge case which should be very rare (where an
unescaped sequence becomes an escape sequence after the lodash
transformation)


wlach:
I realized later today that my earlier solution could fail in an obscure edge case. No rush on looking at / merging this one, since I think that should be exceedingly rare.

wlach:
As a last-minute addition, I changed the server-side unit test to expect a particular type of escaping, rather than just trusting the output of Django's escape function would be ok. This would have caught the original problem before we merged

PR #2529, opened 2019-12-09
Fix No Warning on Browser Button Navigation(Fixes #2403)

### LINKED ISSUE #2403: Navigating between explore and report page with back button does not work
A common thing I do is to write some code in the "explore" mode, and check on the "report" mode occasionally. My first reaction to go from the report mode back into explore is to hit the back button, which navigates back to the previous page (not the explore view). 

### What I Did

1. Start at https://alpha.iodide.io/ (the homepage)
2. Click on "A Brief Tour through Iodide" (or any other notebook)
3. Click on "REPORT" in the top right
4. Click browser back button

### What I Expected

I expect the page to navigate back to "EXPLORE", which is the last immediate view. An alternative is to provide the same dialog when "backspace" is pressed. 

### What Happened 

I navigate back to the homepage without warning dialog, losing any computed state. This is particularly felt when running python-heavy notebooks.

Algentile:
Test Instance can be found [https://local-iodide.herokuapp.com/](https://local-iodide.herokuapp.com/) for QA. This is a change to resolve #2403.

### Pull Request checklist

### DIFF
--- a/src/editor/actions/notebook-actions.js
+++ b/src/editor/actions/notebook-actions.js
@@ -43,6 +43,7 @@ export function setViewMode(viewMode) {
         "",
         `/notebooks/${notebookId}/?${objectToQueryString(params)}`
       );
+      console.log(window.history.length);
     }
     dispatch({
       type: "SET_VIEW_MODE",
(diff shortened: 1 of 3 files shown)


Algentile:
Here we just check for push state in history if a back navigation, or page refresh is called from the browser when a change is made. This can also be triggered by catching the overall onbeforeunload event.

bcolloran:
this works for me, thanks @Algentile. I've been thinking about the interaction model, and I think we can continue to refine it along the lines of what i mentioned in this this comment https://github.com/iodide-project/iodide/issues/2403#issuecomment-549942178 but this is a great step in preventing people from losing work!

PR #2546, opened 2019-12-12
Update Julia/JS integration tier

Keno:
As of https://github.com/JuliaLang/julia/pull/34016, when compiled to wasm, Julia supports integrating with and calling javascript class instances from within Julia code, which makes it Level 3 by the support level defined in this document. Of course the Julia/wasm integration is still quite a bit less mature than the python integration, so maybe that should be noted somewhere, but it does work at least.

cc @vchuravy


bcolloran:
this is awesome news! thanks @Keno !

PR #2550, opened 2019-12-14
Support bytes fetch type (fixes #1786)

### LINKED ISSUE #1786: proposal: binary fetch type
Inspired by the thread on #1686. It is very common to load binary data (pickled objects, arrow data frames, etc.) that need to get put through `Uint8Array`, that it might make sense to have an additional fetch type called `bytes` or `binary` that automatically does it so users don't have to deal with an additional implementation detail.

This would be as simple as just wrapping `fetch(url).then(r=>arrayBuffer()).then(data=>new Uint8Array(data))` most likely.

This would be very simple to implement and reduce the amount of cognitive overhead in figuring out how to get one step further w/ binary data in Iodide.

@mdboom @bcolloran @wlach thoughts here appreciated.

thanks for the idea @vadimkantorov

imranariffin:
### Pull Request checklist

This PR should fix #1786 

This PR is created based on my understanding of the issue after following the discussion on the issue itself & issue #1686 ie we need a new data fetch type called `bytes` that works the same way as `arrayBuffer`:
```
%% fetch
bytes: some_data = https://some-valid-host.com/some-data.some-binary-type
```

So this PR will:
1. Convert arrayBuffer to uint8Array on fetch bytes
2. Add `bytes` to valid fetch types
3. Support syntax highlight & auto-suggest for `bytes`

Some questions:
1. I am not very familiar with Javascript typed array nor Python memoryview nor data science in general that I couldn't find a sample binary data that I can test out on the notebook. The links in the issues are all not working anymore. Can someone provide me with a sample url?
2. I added a TODO flag in the documentation whereby I tried to describe the benefit of introducing this `bytes` fetch type. As I understand, it is so common for data scientists to convert ArrayBuffer into Uint8Array before doing something on it in Python world that it is worth shortcutting that with this new fetch type. Is that correct? I would appreciate some help with this so I can add a better description to the docs.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -167,8 +167,9 @@ available to [fetch chunks](https://iodide-project.github.io/docs/iomd/#fetch-ch
 
 - `json` (load this file as json and parse into a javascript object),
 - `text` (load this file as text),
-- `arrayBuffer` (load this file into an Array Buffer, especially useful when working with typed arrays) and
-- `blob` (load this file as a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)).
+- `arrayBuffer` (load this file into an Array Buffer, especially useful when working with typed arrays)
+- `blob` (load this file as a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)) and
+- `bytes` (load this file into an Array Buffer then convert it to [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) TODO:**Add more description here?**)
 
 These match the `serializerType` in `iodide.file.save` – if you save with a certain `serializerType`, it is recommended to load it with the same `fetchType`.
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -169,7 +169,7 @@ available to [fetch chunks](https://iodide-project.github.io/docs/iomd/#fetch-ch
 - `text` (load this file as text),
 - `arrayBuffer` (load this file into an Array Buffer, especially useful when working with typed arrays)
 - `blob` (load this file as a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)) and
-- `bytes` (load this file into an Array Buffer then convert it to [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) TODO:**Add more description here?**)
+- `bytes` (load this file into an Array Buffer then convert it to [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array), especially useful when working with APIs that expect a typed array of bytes)
 
 These match the `serializerType` in `iodide.file.save` – if you save with a certain `serializerType`, it is recommended to load it with the same `fetchType`.
 
--- a/docs/iomd.md
+++ b/docs/iomd.md
@@ -122,7 +122,7 @@ Each line in a fetch cell must specify:
 1. the "fetch type", one of `js`, `css`, `json`, `text`, `arrayBuffer`, `blob`, or `plugin`
 2. the url from which the resource will be fetched
 
-Additionally, data fetches (`json`, `text`, `blob`, `arrayBuffer`, `bytes`) must specify the variable name into which the data will be stored.
+Additionally, data fetches (`json`, `text`, `blob`, `arrayBuffer`, and `bytes`) must specify the variable name into which the data will be stored.
 
 This example demonstrates how a fetch chunk is used.
 
@@ -135,7 +135,7 @@ text: csvDataString = https://www.exmpl.co/a_csv_file.csv
 arrayBuffer: bigDataframe = https://www.exmpl.co/a_binary.arrow
 json: jsonData = https://www.exmpl.co/a_json_file.json
 blob: blobData = https://www.exmpl.co/a_binary_blob.arrow
-bytes: blobData = https://www.exmpl.co/a_binary_blob.arrow
+bytes: binaryData = https://www.exmpl.co/a_binary_file.hdf5
 ```
 
 All of the requested resources are downloaded in parallel (asynchronously), but if several evaluations are queued, following chunks will not be evaluated until all the resources are available. This allows you to manage the retrieval of assets in a more synchronous workflow, without having to deal with the asynchronous nature of Web APIs (of course, you are free to manage that complexity with JavaScript code and using those APIs if you need that extra control).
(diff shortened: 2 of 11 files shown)


docs/api.md
```diff
@@ -167,8 +167,9 @@ available to [fetch chunks](https://iodide-project.github.io/docs/iomd/#fetch-ch
 
 - `json` (load this file as json and parse into a javascript object),
 - `text` (load this file as text),
-- `arrayBuffer` (load this file into an Array Buffer, especially useful when working with typed arrays) and
-- `blob` (load this file as a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)).
+- `arrayBuffer` (load this file into an Array Buffer, especially useful when working with typed arrays)
+- `blob` (load this file as a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)) and
+- `bytes` (load this file into an Array Buffer then convert it to [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array) TODO:**Add more description here?**)
```
wlach: Maybe:

```suggestion
- `bytes` (load this file into an Array Buffer then convert it to [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array), especially useful when working with APIs that expect a typed array of bytes)
```

docs/iomd.md
```diff
@@ -122,7 +122,7 @@ Each line in a fetch cell must specify:
 1. the "fetch type", one of `js`, `css`, `json`, `text`, `arrayBuffer`, `blob`, or `plugin`
 2. the url from which the resource will be fetched
 
-Additionally, data fetches (`json`, `text` or `blob`) must specify the variable name into which the data will be stored.
+Additionally, data fetches (`json`, `text`, `blob`, `arrayBuffer`, `bytes`) must specify the variable name into which the data will be stored.
```
wlach:
```suggestion
Additionally, data fetches (`json`, `text`, `blob`, `arrayBuffer`, and `bytes`) must specify the variable name into which the data will be stored.
```

docs/iomd.md
```diff
@@ -135,6 +135,7 @@ text: csvDataString = https://www.exmpl.co/a_csv_file.csv
 arrayBuffer: bigDataframe = https://www.exmpl.co/a_binary.arrow
 json: jsonData = https://www.exmpl.co/a_json_file.json
 blob: blobData = https://www.exmpl.co/a_binary_blob.arrow
+bytes: blobData = https://www.exmpl.co/a_binary_blob.arrow
```
bcolloran:
```suggestion
bytes: binaryData = https://www.exmpl.co/a_binary_file.hdf5
```

imranariffin:
> Hi @imranariffin! Thank you so much, this looks great. Can you test it out with a variant of this notebook: https://definitely-staging.iodide.io/notebooks/221/ -- you will obviously need to download the attached excel file and reattach it to your notebook
> 
> If this PR works as expected, you should be able to remove the js pre-processing in the notebook after changing the fetch type to bytes.

Yay!

Question: how do I download the excel file the notebook? I couldn't find any user documentation on that. 

Another related question: If I were to upload my own excel file from my local file system using a relative fetch (I'm assuming that's what fetch with relative path is for) how do I know what's my root directory?

wlach:
> > Hi @imranariffin! Thank you so much, this looks great. Can you test it out with a variant of this notebook: https://definitely-staging.iodide.io/notebooks/221/ -- you will obviously need to download the attached excel file and reattach it to your notebook
> > If this PR works as expected, you should be able to remove the js pre-processing in the notebook after changing the fetch type to bytes.
> 
> Yay!
> 
> Question: how do I download the excel file the notebook? I couldn't find any user documentation on that.

We really should allow showing the files modal when the user doesn't own the notebook: I filed #2562 for this. But in any case the file should be downloadable anonymously like so:

https://definitely-staging.iodide.io/notebooks/221/files/Furniture_only.xls

> Another related question: If I were to upload my own excel file from my local file system using a relative fetch (I'm assuming that's what fetch with relative path is for) how do I know what's my root directory?

Hmm, I'm not sure if I understand the question? Files uploaded to iodide are always under `files/` path after the notebook on the server.

imranariffin:
> We really should allow showing the files modal when the user doesn't own the notebook: I filed #2562 for this. But in any case the file should be downloadable anonymously like so:
> 
> https://definitely-staging.iodide.io/notebooks/221/files/Furniture_only.xls

I see, gotcha. Thanks!

> Hmm, I'm not sure if I understand the question? Files uploaded to iodide are always under files/ path after the notebook on the server.

I was basically wondering how does one upload a fail to a notebook that they own, but never mind I finally figured out how do it -- totally missed the "managed files" from the top-left drop down!

So I was able to display the same output by using `bytes` fetch without having to convert it to Uint8Array beforehand:

```
%% fetch
bytes: file=Furniture_only.xls
```
```
%% js
var view = new Uint8Array(file);
view
```
```
%% py
from js import view
import io
b = io.BytesIO(view)
b
```
```
%% py
import micropip
micropip.install('xlrd')
```
```
%% py 
import pandas as pd
d = pd.read_excel(b)
d
```
-- See example notebook here: https://definitely-staging.iodide.io/notebooks/230/

**Note** that I used `arrayBuffer` as the best case instead of `blob` as seen on the notebook that @wlach has shared. I got an error converting the `blob` to `arrayBuffer` error when running the notebook:
```
TypeError: file0.arrayBuffer is not a function
[eval] (cell:2:7)
```

imranariffin:
> +1 what @wlach said, this pr looks really good. i've added one other little suggestion, and once you've tried @wlach suggested i think this will be good to merge!
> 
> Thank you!!

Thanks for the review! Test was a success.

wlach:
Thank you so much @imranariffin! Will be making a new release with this today.

imranariffin:
Yay, proud to be part of release! Thank you @wlach @bcolloran for the reviews!

imranariffin:
@wlach I realized while working on this PR that I wasn't able to easily tell if a notebook I was looking at was already forked by me. Is there a plan to display a message when a notebook has already been forked by the user?

Screenshot from 2019-12-30 21-35-50

wlach:
> @wlach I realized while working on this PR that I wasn't able to easily tell if a notebook I was looking at was already forked by me. Is there a plan to display a message when a notebook has already been forked by the user?

Not at present, seems like it would be a good idea to have something like this. Looks like github does this if you try to fork a repo twice:

image

In our case we would probably either offer to open up existing fork(s) in a new tab or allow the user to fork again. You could probably do this with a modal element. A bit complicated but might be a fun project!

imranariffin:
I see, I agree that a modal displaying existing forks and a button to fork it again would be great. Should I create an issue for this?

wlach:
> I see, I agree that a modal displaying existing forks and a button to fork it again would be great. Should I create an issue for this?

Sounds like a good idea! Others may have opinions about this as well.

Issue #2563, opened 2019-12-17
Print whole report to PDF


mehmetoguzderin:
When converting to PDF using browser's printing functionality, due to iframe, it produces a cropped page. Example notebook:
https://alpha.iodide.io/notebooks/3427/?viewMode=report

wlach:
@mehmetoguzderin as I mentioned in the meeting, I think it might be worth trying to print *just* the iframe using the window.print command:

https://developer.mozilla.org/en-US/docs/Web/API/Window/print

The mechanism to call this would be the eval frame's port to the editor:

https://github.com/iodide-project/iodide/blob/69409c8f70731ed83ee18da828e3cb66ca966982/src/eval-frame/port-to-editor.js

We could experiment with adding a "Print" button beside "Explore" in report view, and adding an option to the menu in explore view.

Are you interested in giving this a try? It might not be that difficult if we're lucky...

mehmetoguzderin:
@wlach Yes, I would like to try it on the weekend.

wlach:
> @wlach Yes, I would like to try it on the weekend.

Excellent! I'll assign this to you for now

mehmetoguzderin:
@wlach Great!

bcolloran:
re UI/UX, i think we need a better approach than adding another button in the top corner.

but it looks like maybe we can intercept attempts to print and trigger custom logic, so maybe we don't need to change the UI at all?--
https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeprint

wlach:
> re UI/UX, i think we need a better approach than adding another button in the top corner.

Yeah agreed, that was just a suggestion for a first pass.

> but it looks like maybe we can intercept attempts to print and trigger custom logic, so maybe we don't need to change the UI at all?--
> https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeprint

That sounds perfect! The best UI is no UI.

bcolloran:
fwiw: off the top of my head, i think the way to do thiswould be something like:

assuming the `print` call is sent to the top level window and doesn't somehow cascade down to the iframe:
 
   - we would want to intercept the `print` call in the editor window
   - if `viewMode = explore`, do the normal print action
   - if `viewMode = report`, send a signal down to the iframe to trigger printing...
        - re-layout the page for printing
        - call window print`window.print()`
        - restore the normal report view layout

mehmetoguzderin:
[...]

PR #2582, opened 2019-12-22
Add printing support for reports that are longer than one page

mehmetoguzderin:
Aims to fix https://github.com/iodide-project/iodide/issues/2563 issue.

This pull request adds a printing button that calls printing from iframe to the left of `EXPLORE` when using report view and changes the styling of the fixed container when in report view to make it possible to print a clean page without undesired bordering and page limiting.

### DIFF
--- a/src/editor/components/menu/view-mode-toggle-button.jsx
+++ b/src/editor/components/menu/view-mode-toggle-button.jsx
@@ -16,7 +16,15 @@ export class ViewModeToggleButtonUnconnected extends React.Component {
   };
   constructor(props) {
     super(props);
+    this.printReport = this.printReport.bind(this);
     this.toggleViewMode = this.toggleViewMode.bind(this);
+  }
+
+  printReport() {
+    if (this.props.isReportView) {
+      document.getElementById("eval-frame").contentWindow.focus();
+      document.getElementById("eval-frame").contentWindow.print();
+    }
   }
 
   toggleViewMode() {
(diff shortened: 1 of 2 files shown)


src/editor/components/menu/view-mode-toggle-button.jsx
```diff
@@ -16,9 +16,17 @@ export class ViewModeToggleButtonUnconnected extends React.Component {
   };
```
wlach: We probably would want to factor this into its own higher-level component, as it's not really a button anymore.

There are other minor UI Chrome elements that we want in the report view (like a spinner icon when the document is loading) so this might be an opportunity to pave the way for that.

src/editor/components/menu/view-mode-toggle-button.jsx
```diff
@@ -33,14 +41,26 @@ export class ViewModeToggleButtonUnconnected extends React.Component {
...
-        </Button>
+        <div>
+          {this.props.isReportView && (
+            <Button
+              style={this.props.style}
+              onClick={this.printReport}
+              variant="text"
+              mini
+            >
+              Print
```
wlach: Could probably use a printer icon as opposed to a textual button here, I'm sure material ui has one.

src/shared/components/fixed-position-container.jsx
```diff
@@ -53,6 +53,12 @@ export function mapStateToProps(state, ownProps) {
       width: "100%",
       height: "100%"
     };
+
+    if (state.viewMode === "REPORT_VIEW") {
+      style.position = "static";
+      style.border = "none";
+      style.overflow = "visible";
+    }
```
wlach: I suspect it is possible to do this with pure css / emotion, which is probably preferable.

bcolloran:
thanks for the PR @mehmetoguzderin -- unfortunately, i think this is not the way we want to go with this issue
see https://github.com/iodide-project/iodide/issues/2563#issuecomment-571278544

mehmetoguzderin:
@bcolloran Thanks for the feedback!

PR #2600, opened 2019-12-30
Handle cases where coalescing revisions results in no change (fixes #2517) 

### LINKED ISSUE #2517: Revision coalescing sometimes results in two adjacent revisions with no changes
#2461 added automatic revision coalescing of "draft" revisions, but there's one edge case where not taken into account:

Imagine a notebook with these three revisions:

```
A (non-draft) = aaa
B (draft) = aab
C (draft) = aaa
```

This code will delete B but keep C, which is no different than A. In this case we should just delete both B and C, since the notebook hasn't really changed.

zzl0:
### Pull Request checklist

This PR should fix #2517. As discussed in the issue, this PR deletes the revision if it's same as the previous one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # (Unreleased; add upcoming change notes here)
 
-- Solve duplicate revisions issue after revision coalescing (fixes 2517)
+- Handle cases where coalescing revisions results in no change (fixes #2517)
 
 # 0.17.0 (2019-12-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # (Unreleased; add upcoming change notes here)
 
+- Solve duplicate revisions issue after revision coalescing (fixes 2517)
+
 # 0.17.0 (2019-12-19)
 
 - Further refine escaping iomd returned by iodide server (#2526)
--- a/server/notebooks/tasks.py
+++ b/server/notebooks/tasks.py
@@ -22,21 +22,41 @@ def execute_notebook_revisions_cleanup(notebook_id, now_utc=None):
       (also called Tumbling windows).
     """
     draft_revisions = NotebookRevision.objects.filter(notebook_id=notebook_id, is_draft=True)
+    try:
+        latest_non_draft_revision = NotebookRevision.objects.filter(
+            notebook_id=notebook_id, is_draft=False
+        ).latest("created")
+    except NotebookRevision.DoesNotExist:
+        latest_non_draft_revision = None
 
     # group notebook revisions into fixed-size windows
     revision_groups = defaultdict(list)
     for revision in draft_revisions:
         key = int(revision.created.timestamp()) // NOTEBOOK_REVISION_SAVE_INTERVAL_SECS
         revision_groups[key].append(revision)
-    revision_windows = [revision_groups[k] for k in sorted(revision_groups.keys(), reverse=True)]
+    revision_windows = [revision_groups[k] for k in sorted(revision_groups.keys())]
 
     now_utc = now_utc or datetime.now(tz=pytz.utc)
     threshold = timedelta(seconds=NOTEBOOK_REVISION_SAVE_INTERVAL_SECS)
 
-    # delete intermediate revisions
+    # get intermediate revisions
     intermediate_revision_ids = [
         revision.id for revision_window in revision_windows for revision in revision_window[1:]
     ]
+    # for remaining_revisions, if a revision is same as the previous one (issue #2517), then
+    # also add it to the intermediate revisions.
+    remaining_revisions = [revision_window[0] for revision_window in revision_windows]
+    last_revision = latest_non_draft_revision
+    for revision in remaining_revisions:
+        if (
+            last_revision
+            and last_revision.title == revision.title
+            and last_revision.content == revision.content
+        ):
+            intermediate_revision_ids.append(revision.id)
+        else:
+            last_revision = revision
+    # delete intermediate revisions
     NotebookRevision.objects.filter(
         id__in=intermediate_revision_ids, created__lt=now_utc - threshold
     ).delete()
--- a/server/notebooks/tasks.py
+++ b/server/notebooks/tasks.py
@@ -45,7 +45,8 @@ def execute_notebook_revisions_cleanup(notebook_id, now_utc=None):
     ]
 
     # for remaining_revisions, if a revision is same as the previous one (issue #2517), then
-    # also add it to the intermediate revisions, so we don't accidentally coalesce down to a duplicate revision.
+    # also add it to the intermediate revisions, so we don't accidentally coalesce down to
+    # a duplicate revision.
     remaining_revisions = [revision_window[0] for revision_window in revision_windows]
     last_revision = latest_non_draft_revision
     for revision in remaining_revisions:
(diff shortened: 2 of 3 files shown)


server/notebooks/tasks.py
```diff
@@ -22,21 +22,41 @@ def execute_notebook_revisions_cleanup(notebook_id, now_utc=None):
       (also called Tumbling windows).
     """
     draft_revisions = NotebookRevision.objects.filter(notebook_id=notebook_id, is_draft=True)
+    try:
+        latest_non_draft_revision = NotebookRevision.objects.filter(
+            notebook_id=notebook_id, is_draft=False
+        ).latest("created")
+    except NotebookRevision.DoesNotExist:
+        latest_non_draft_revision = None
```
wlach: I don't think this should ever happen. In theory it might currently be possible to delete all the revisions of a notebook from the user page, but I would expect other weird things to happen after that. I would suggest adding something like this (you will need to set up logging inside this file, look at `server/files/tasks.py` for an example)

```suggestion
        logger.warning("Attempted to prune revisions from notebook %s with no revisions", notebook_id)
        return  # should never happen
```

server/notebooks/tasks.py
```diff
@@ -22,21 +22,41 @@ def execute_notebook_revisions_cleanup(notebook_id, now_utc=None):
...
+    for revision in remaining_revisions:
+        if (
+            last_revision
+            and last_revision.title == revision.title
+            and last_revision.content == revision.content
+        ):
+            intermediate_revision_ids.append(revision.id)
+        else:
+            last_revision = revision
+    # delete intermediate revisions
```
wlach: Add a newline here:

```suggestion

    # delete intermediate revisions
```

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # (Unreleased; add upcoming change notes here)
 
+- Solve duplicate revisions issue after revision coalescing (fixes 2517)
```
```suggestion
- Handle cases where coalescing revisions results in no change (fixes #2517)
```

server/tests/test_notebook_revisions_cleanup.py
```diff
@@ -43,6 +49,25 @@ def notebook_and_draft_revisions(fake_user):
     return notebook, notebook_revisions
 
 
+@pytest.fixture
+def notebook_and_duplicate_draft_revisions(fake_user):
```
wlach: I would just call this `notebook_with_duplicate_draft_revisions` and then only return the notebook object from this fixture. It looks like you don't actually use the notebook_revisions list in your test.

server/notebooks/tasks.py
```diff
@@ -22,21 +22,41 @@ def execute_notebook_revisions_cleanup(notebook_id, now_utc=None):
...
     now_utc = now_utc or datetime.now(tz=pytz.utc)
     threshold = timedelta(seconds=NOTEBOOK_REVISION_SAVE_INTERVAL_SECS)
 
-    # delete intermediate revisions
+    # get intermediate revisions
     intermediate_revision_ids = [
         revision.id for revision_window in revision_windows for revision in revision_window[1:]
     ]
+    # for remaining_revisions, if a revision is same as the previous one (issue #2517), then
+    # also add it to the intermediate revisions.
```
wlach:
```suggestion
    # also add it to the intermediate revisions, so we don't accidentally coalesce down to a duplicate revision.
```

server/notebooks/tasks.py
```diff
@@ -22,21 +22,41 @@ def execute_notebook_revisions_cleanup(notebook_id, now_utc=None):
...
 
     now_utc = now_utc or datetime.now(tz=pytz.utc)
     threshold = timedelta(seconds=NOTEBOOK_REVISION_SAVE_INTERVAL_SECS)
 
-    # delete intermediate revisions
+    # get intermediate revisions
     intermediate_revision_ids = [
         revision.id for revision_window in revision_windows for revision in revision_window[1:]
     ]
+    # for remaining_revisions, if a revision is same as the previous one (issue #2517), then
```
wlach:
```suggestion

    # for remaining_revisions, if a revision is same as the previous one (issue #2517), then
```

server/tests/test_notebook_revisions_cleanup.py
```diff
@@ -43,6 +49,25 @@ def notebook_and_draft_revisions(fake_user):
...
+    notebook_revisions = []
+
+    for i, created in enumerate(CREATED_DATETIMES_FOR_DUPLICATE_REVISIONS):
+        revision = NotebookRevision.objects.create(
+            notebook=notebook,
+            title=f"Revision",
+            content=f"fake notebook content for revision",
+            is_draft=True,
+        )
+        revision.created = created
```
wlach:
```suggestion
        revision.created = created  # manually override the value provided by auto_now_add
```

server/notebooks/tasks.py
zzl0: This can happen, e.g.: when you create a new notebook, there is only one draft revision and there is no non-draft revision.

This logic here is to find the latest **non-draft** revision.

server/tests/test_notebook_revisions_cleanup.py
zzl0: Sure

server/notebooks/tasks.py
wlach: Ah right, sorry, you are correct. This makes perfect sense.

server/tests/test_notebook_revisions_cleanup.py
```diff
@@ -38,13 +43,29 @@ def notebook_and_draft_revisions(fake_user):
...
+
+
+@pytest.fixture
+def notebook_with_duplicate_draft_revisions(fake_user):
+    notebook = Notebook.objects.create(owner=fake_user, title="Fake notebook")
+
+    for i, created in enumerate(CREATED_DATETIMES_FOR_DUPLICATE_REVISIONS):
+        revision = NotebookRevision.objects.create(
+            notebook=notebook,
+            title=f"Revision",
```
wlach: Format string not needed

```suggestion
            title="Revision",
```

server/tests/test_notebook_revisions_cleanup.py
```diff
@@ -66,6 +87,21 @@ def test_execute_notebook_revisions_cleanup(notebook_and_draft_revisions):
     assert updated_revisions[3].created == CREATED_DATETIMES[-6]
 
 
+def test_execute_notebook_revisions_cleanup_for_duplicate_revisions(
+    notebook_with_duplicate_draft_revisions,
+):
```
wlach:
```suggestion
):
    # tests the case where a draft revision is not actually different from the last non-draft revision
```

zzl0:
> @zzl0 apparently I don't have the rights to make changes to your fork -- could you apply my last suggestions? Happy to merge after that. Thanks again for digging into this!

Thanks for your suggestions , I committed them to my branch.

server/tests/test_notebook_revisions_cleanup.py
```diff
@@ -66,6 +87,22 @@ def test_execute_notebook_revisions_cleanup(notebook_and_draft_revisions):
...
+    # tests the case where a draft revision is not actually different from the last non-draft revision
```
wlach:
```suggestion
    # tests the case where a draft revision is not actually different from the 
    # last non-draft revision
```

wlach:
Oh grump, I suggested something that wouldn't pass the linter. Can you apply this one last suggestion?

zzl0:
> Oh grump, I suggested something that wouldn't pass the linter. Can you apply this one last suggestion?

I just changed the file before noticing your suggestion. It should pass now.

server/tests/test_notebook_revisions_cleanup.py
zzl0: Thanks, I already changed this line.

wlach:
Merged!

Issue #2101, opened 2019-07-30
"clear all variables" menu item doesn't actually clear all the variables...


bcolloran:
i just noticed that the "clear all variables" menu item clears the variables out of the redux store (because they disappear from the "workspace pane), but the  variables remain in the eval frame JS scope (so for example you can still reference them in code or from the console)

PR #2621, opened 2020-01-10
Move "clear declared variables" to workspace pane

wlach:
After doing this, it is a bit unclear to me whether we should keep this
functionality or just remove it altogether -- it's a bit of a footgun,
doesn't work as advertised (e.g. #2101) and doesn't seem that useful even
on its own terms (I have never actually tried to use it outside of testing
this PR). But in any case, this seems like a more logical place for the action
than the menu.


wlach:
Exciting screen shot:

image

wlach:
We talked about this in today's iodide meeting. tl;dr is that this feature needs a re-think, best to just take this feature out altogether for now. Will try to get to that soon.

wlach:
Actually let's just close this

PR #2676, opened 2020-02-10
Allow creation of new or tryit notebook from iomd template

### LINKED ISSUE #2565: Add a simplified way of creating notebooks with no authentication credentials
We added the capability for services to create notebook on a user's behalf in #, but it's a little bit of a pain to set up: administrative effort is required to configure both the iodide server and the service creating the notebook to use a special key.

It would be nice if we could just do everything on the client, wouldn't it? I originally thought this was intractable, but have since come up with an approach that I think is secure:

1. Allow notebooks to be owned by *no user* initially and allow anonymous creation of notebooks of this kind
2. Allow notebooks to be addressed by a unique hash of their id (maybe using https://github.com/nshafer/django-hashid-field) in addition to (at first) their numeric id. Anonymous notebooks can only be viewed by using their hash (which is returned when they are created)
3. When an anonymous notebook is viewed by a logged in user, it will automatically be claimed. When it is viewed by an unauthenticated one (only applicable to public sites like alpha.iodide.io), it will show the usual message about needing to be logged in to save your changes.

So in this case, to add an "open in iodide button" to a web app (or a web extension), you would do something like:

1. POST a new notebook owned by no user. Get the content of the notebook back, including the hash id (which should be known only to the person creating the notebook)
2. Open a new tab at (e.g. iodide.telemetry.mozilla.org/notebooks/&lt;hash id&gt;/)

This should be resilient against the following types of attacks:

1. Users won't be able to claim/view notebooks they didn't create (since they won't have the hash)
2. On a server where we require authentication (e.g. iodide.telemetry.mozilla.org), we can allow anonymous creation of notebooks with low risk (creating a notebook doesn't grant access to any other resources)

With this in place, we should be able to easily add "explore this dataset in iodide" buttons to all our telemetry tools, including ones that use static frontends. No auth credential setup required.

/cc @jezdez @openjck @hamilton , I'm curious what you think of this idea. I may give it a shot in the next couple weeks if there aren't any glaring flaws with it.

wlach:
Partially fixes #2565, currently this is limited to creating a notebook with a specific iomd-- ideally we would also allow attaching files to the payload as well, but that will require handling these as POST requests (which is considerably more technically involved)

This also needs tests and documentation.

### Pull Request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # (Unreleased; add upcoming change notes here)
 
+- Can now specify notebook iomd via a URL parameter with new or
+  tryit endpoints (fixes #2565) (#2676)
+
 # 0.18.0 (2020-01-22)
 
 - Center "Help" and "Play" icons in some browsers (fixes #2567) (#2570)
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -128,18 +128,31 @@ def notebook_revisions(request, pk):
     )
 
 
+def _get_new_notebook_content(iomd):
+    if iomd:
+        return iomd
+    # default is to create a new notebook from our default template
+    return get_template("new_notebook_content.iomd").render()
+
+
+def _get_iomd_params(iomd):
+    if iomd:
+        return f"?iomd={urllib.parse.quote_plus(iomd)}"
+    return ""
+
+
 @ensure_csrf_cookie
 def new_notebook_view(request):
+    iomd = request.GET.get("iomd")
     if not request.user.is_authenticated:
-        return redirect(reverse("try-it"))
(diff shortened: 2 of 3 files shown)


wlach:
Put this PR on staging, here's an example of it in action: https://definitely-staging.iodide.io/tryit/?iomd=%25%25%20md%0A%0AHi%20this%20is%20a%20markdown%20cell%21%0A%0A%25%25%20js%0A%0A%2F%2F%20Oh%20hai%20this%20is%20a%20js%20cell%0Avar%20x%20%3D%20123%3B%0Ax%20%2F%203.14

wlach:
Added tests and changelog, I think I'm going to skip documentation of this feature for now.

PR #2690, opened 2020-02-17
allow pre-attached template and data file (#issue2681)

Ralts1337:
Adds a function that returns a iodide page with customized template file and data
parameter: httpPostRequest
returns: iodide Django view with data attached to customized template.

file uploads must be done through POST like this:
<form action= " addressToIodide/customize">:
<input name = "template"> for template file  
<input name = "data"> for data file

### Pull Request checklist

### DIFF
--- a/server/notebooks/templates/notebook_uploaded_files.html
+++ b/server/notebooks/templates/notebook_uploaded_files.html
@@ -7,7 +7,11 @@
 <div id="notebook-header"></div>
 <div id="editor-react-root"></div>
 <script id="iomd" type="text/iomd">{{ iomd|escape }}</script>
-<script id="file0" type="text">{{datafile|safe}}</script>
+
+{% for data in datalist %}
+<script id="file{{forloop.counter0}}" type="text">{{data|safe}}</script>
+{% endfor %}
+
 <iframe
   id="eval-frame"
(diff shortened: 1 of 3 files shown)


server/notebooks/views.py
```diff
@@ -186,3 +187,36 @@ def tryit_view(request):
...
+        
+        postedTemplate = request.FILES['template']
+        postedData = request.FILES['data']
+    
+        template = postedTemplate.read().decode()
+        data = postedData.read().decode()
+
+        iomd = template + data
+
+        return render(
```
wlach: This doesn't look right to me. What I'd expect is a response that looks like this:

```html
<script id="iomd" type="text/iomd">%% md
example md content
</script>
<script id="file0" type="application/json">
{
 "a": 1,
 ...
}
</script>
```

Putting the response into the script tag will let us easily extract and process the response into a file from javascript.

server/notebooks/templates/notebook_uploaded_files.html
```diff
@@ -0,0 +1,29 @@
+{% extends "base.html" %}
```
wlach: I think we can use the existing `notebook.html` template here with a few adjustments-- see my branch for details.

(but don't worry about making this change, I can do it when I merge your changes in)

PR #2707, opened 2020-02-23
Always show clear history

cglotr:
Always show the "Clear history" button in the _Console_ dropdown.

In my opinion, hiding the "Clear history" button is a bad UX since:
- It changes the user expectation of the dropdown. Let say the user opens the dropdown when the history is empty, the user will put in his/her mind that the last button in the dropdown is some button. When the user opens the dropdown again, his/her expectation should be that the last button is the same button. But, depending on the history the user expectation will be violated. Also, it's hard to establish a habit since the brain needs to figure out the pattern first. If the dropdowns remain the same, it's much more intuitive & easier for the brain to develop a habit. For example, some developers liked _clearing_ an already _cleared_ console since they developed that habit.
- It makes it hard for new users to discover the "Clear history" functionality.

### Pull Request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Can now specify notebook iomd via a URL parameter with new or
   tryit endpoints (fixes #2565) (#2676)
+- "Clear history" button is always shown in the Console dropdown
 
 # 0.18.0 (2020-01-22)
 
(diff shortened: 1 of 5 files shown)


cglotr:
My pleasure to contribute too & thanks for the merge :) @wlach

PR #2727, opened 2020-03-06
Temporary files + files in template strings

### LINKED ISSUE #2680: Allow attaching files to a trial notebook
For the purposes of experimentation, it would be nice to allow users to attach/use files with a trial notebook. Obviously we can't *actually* add the files to the notebook, since we don't have permissions to do so, but we could probably fake it by adding a cached copy of the files and then uploading them if/when the user logs in and saves the notebook.

Off the top of my head, I think this would have several parts:

* keeping an optional data structure in the redux store of "unuploaded files" (with virtual urls + content)
* adding a user interface element to the files dialog to add these files (and displaying them if they are present)
* modifying the fetch cell to retrieve the contents from the data structure if they are unuploaded

wlach:
WIP - this is a bit of a giant pull request

* Refactor files modal to use redux
* Allow "uploading" unpersisted files to a notebook you don't own (fixes #2680)
* Probably a bunch of other junk I've forgotten about

### Pull Request checklist

### DIFF
--- a/server/notebooks/templates/notebook.html
+++ b/server/notebooks/templates/notebook.html
@@ -20,5 +20,8 @@
 <script>
   window.evalFrameOrigin = "{{ eval_frame_origin }}";
 </script>
+{% for filename, filecontent, mimetype in files %}
+<script id="file-{{ filename }}" type="application/base64" mimetype="{{ mimetype }}">{{ filecontent }}</script>
+{% endfor %}
 <script src="/iodide.js"></script>
 {% endblock %}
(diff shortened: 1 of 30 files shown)


wlach:
I'm probably going to violate some of the historical protocols of the project when landing this:

* Not going to provide documentation (this is more of an experimental feature to enable another project, we can say more if it becomes widely used)
* Just going to land without review (people on the team have given the ok, and we've discussed the ideas in detail)
* Not going to spend a long time writing new tests (I'm strapped for time-- I made things work as expected for common cases-- again, we can circle back if new energy comes to the project)

PR #2816, opened 2020-04-24
file upload test draft

Ralts1337:
Draft.
Test `GET` request with file parameters to `/try-it` endpoint


server/tests/test_notebook_view.py
```diff
@@ -149,6 +149,36 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+@pytest.mark.parametrize("logged_in", [True, False])
+@pytest.mark.parametrize("request_method", ["POST", "GET"])
+def test_tryit_view_file_uploads(client, fake_user, logged_in, request_method):
+    import tempfile
+    import mimetypes
+
+    # we should see the same script blocks
+    # for both logged_in and both request_method
+
+    with tempfile.NamedTemporaryFile(mode="w+") as f:
```
wlach, line 161: There is no need to create an actual temporary file here. You can just upload some arbitrary content (e.g. a string).

server/tests/test_notebook_view.py
```diff
@@ -149,6 +149,36 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+        f.name = "data.txt"
+        file_content = "testing"
+        file_mimetype = mimetypes.guess_type(f.name)
+        f.write(file_content)
+        f.seek(0)
+        path = reverse("try-it")
+
+        def http_response(request_method):
+            if request_method == "POST":
```
wlach, line 170: I didn't get into this earlier, but we won't be using `/tryit/` for `POST` requests-- for that, we'll be using a specialized endpoint (probably `/from-template/`). Let's create a seperate test as I think it'll be pretty different from this one.

This also eliminates the need to create a special helper function

server/tests/test_notebook_view.py
```diff
@@ -149,6 +149,36 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+                return client.post(path, {"filename": f.name, "file": f}, follow=True)
+            else:
+                return client.get(path, {"filename": f.name, "file": open(f.name)}, follow=True)
```
wlach, line 173: Try just encoding this in the URL:

client.get(f'/tryit?file={file_content}&filename={filename}')

server/tests/test_notebook_view.py
```diff
@@ -149,6 +149,25 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+def test_tryit_view_get_api(client, fake_user, logged_in):
+    """
+    Test if /try-it handles GET request file upload correctly
+    """
+
+    import mimetypes
+
+    file_name = "data.txt"
+    file_content = "testing"
+    file_mimetype = mimetypes.guess_type(file_name)
```
wlach, line 162: We can skip this line, and just hardcode `text/plain` below

```suggestion
```

server/tests/test_notebook_view.py
```diff
@@ -149,6 +149,25 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
         assert len(response.redirect_chain) == 0
 
 
+@pytest.mark.parametrize("logged_in", [True, False])
+def test_tryit_view_get_api(client, fake_user, logged_in):
+    """
+    Test if /try-it handles GET request file upload correctly
```
wlach, line 155:
```suggestion
    Test that we can pass file parameters to the try it view
```

server/tests/test_notebook_view.py
```diff
@@ -149,6 +149,25 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+    """
+
+    import mimetypes
```
wlach, line 158: Not needed (see below):

```suggestion
```

server/tests/test_notebook_view.py
```diff
@@ -149,6 +149,25 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+
+    file_name = "data.txt"
+    file_content = "testing"
+    file_mimetype = mimetypes.guess_type(file_name)
+
+    if logged_in:
+        client.force_login(fake_user)
+
+    resp = client.get(f"/tryit?file={file_content}&filename={file_name}", follow=True)
+    assert get_file_script_block(resp.content, file_name, file_mimetype) == file_content
```
wlach, line 168: Looking again, this is only correct if `logged_in` is false. If it is true, you check that the file is created as expected:

https://github.com/iodide-project/iodide/blob/54d127d7f7ff1d2e4bb2c3049e33c75805e745c9/server/notebooks/views.py#L181

In this case there should be no file script block (since the files are already created on the server, there is no need to provide "fake" files in script blocks to the client)

server/tests/test_notebook_view.py
```diff
@@ -149,6 +150,35 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+        assert resp.status_code == 200
+        assert NotebookRevision.objects.count() == 1
+        assert Notebook.objects.count() == 1
+        assert File.objects.count() == 1
+        assert File.objects.first().filename == file_name
+        assert bytes(File.objects.first().content) == bytes(file_content, encoding="utf8")
+        assert File.objects.first().notebook.id == Notebook.objects.first().id
+
+    else:
+        resp = client.get(f"/tryit?iomd=123&file={file_content}&filename={file_name}", follow=True)
```
wlach, line 175:
```suggestion
        resp = client.get(reverse('try-it') + f"?iomd=123&file={file_content}&filename={file_name}", follow=True)
```

wlach:
Looks good now, thanks! Will merge

PR #2849, opened 2020-05-10
Allow creation of Iodide notebooks from POST request

wlach:
This adds the POST method, enabling the creation of relatively large (trial) notebooks via standard web APIs. /cc @Ralts1337

Could probably use a few more tests before landing.

### Pull Request checklist

### DIFF
--- a/server/notebooks/views.py
+++ b/server/notebooks/views.py
@@ -9,7 +9,8 @@ from django.shortcuts import get_object_or_404, redirect, render
 from django.template.loader import get_template
 from django.urls import reverse
 from django.views.decorators.clickjacking import xframe_options_exempt
-from django.views.decorators.csrf import ensure_csrf_cookie
+from django.views.decorators.csrf import csrf_exempt, ensure_csrf_cookie
+from django.views.decorators.http import require_http_methods
 
 from ..base.models import User
 from ..files.models import File
@@ -199,6 +200,50 @@ def tryit_view(request):
         files = _get_files_from_request(request, base64_encode=True)
     except ValidationError as e:
         return HttpResponseBadRequest(content=e)
+
+    return render(
+        request,
+        "notebook.html",
+        {
+            "user_info": {},
+            "notebook_info": {
+                "connectionMode": "SERVER",
+                "tryItMode": True,
+                "title": title if title else "Untitled notebook",
+            },
+            "iomd": _get_new_notebook_content(iomd),
+            "files": files,
+            "iframe_src": _get_iframe_src(),
+            "eval_frame_origin": EVAL_FRAME_ORIGIN,
+        },
+    )
+
+
+@require_http_methods(["POST"])
+@csrf_exempt
+def from_template_view(request):
+    iomd = urllib.parse.unquote(request.POST.get("iomd", ""))
+    if not iomd:
+        return HttpResponseBadRequest("Must specify iomd template")
+
+    title = request.POST.get("title")
+
+    # NOTE: We can not sadly not use files (request.FILES) attached to forms, as
+    # it is not possible to attach those programatically from the client side via
+    # JavaScript
+    files = []
+    for filename in [
+        k.replace("file-", "", 1) for k in request.POST.keys() if k.startswith("file-")
+    ]:
+        if not filename:
+            return HttpResponseBadRequest("Filename must have at least one character")
+        file_content = request.POST[f"file-{filename}"]
+        actual_file_size = len(file_content) * 3 / 4  # base64 size = 4 * actual / 3
+        if actual_file_size > MAX_FILE_SIZE:
+            return HttpResponseBadRequest(
+                f"File {filename} exceeds maximum file size {MAX_FILE_SIZE}"
+            )
+        files.append((filename, file_content, mimetypes.guess_type(filename)[0]))
 
     return render(
         request,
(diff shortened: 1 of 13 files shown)


wlach:
Demo application for creating an iodide notebook on the command line via a python script: https://github.com/wlach/create-iodide-notebook

server/notebooks/views.py
```diff
@@ -216,3 +217,47 @@ def tryit_view(request):
...
+def from_template_view(request):
+    iomd = urllib.parse.unquote(request.POST.get("iomd", ""))
+    if not iomd:
+        return HttpResponseBadRequest("Must specify iomd template")
+
+    title = request.POST.get("title")
+
+    # NOTE: We can not sadly not use files (request.FILES) attached to forms, as
+    # it is not possible to attach those programatically from the client side via
+    # JavaScript
```
jezdez, line 233: I'm not sure if that is true, IIRC you can use the fetch API to do that, and it should show up on the Django side correctly. [StackOverflow comment talking about this](https://stackoverflow.com/a/51109645)

server/notebooks/views.py
```diff
@@ -216,3 +217,47 @@ def tryit_view(request):
...
+        k.replace("file-", "", 1) for k in request.POST.keys() if k.startswith("file-")
+    ]:
+        if not filename:
+            return HttpResponseBadRequest("Filename must have at least one character")
+        file_content = request.POST[f"file-{filename}"]
+        actual_file_size = len(file_content) * 3 / 4  # base64 size = 4 * actual / 3
+        if actual_file_size > MAX_FILE_SIZE:
+            return HttpResponseBadRequest(
+                f"File {filename} exceeds maximum file size {MAX_FILE_SIZE}"
+            )
```
jezdez, line 245: I'm really worried that this could be used to DOS the server, I would feel much better if we would be able to handle this validation via a Django FileField that has been well tested and has a good approach to streaming files to /tmp if needed.

server/notebooks/views.py
wlach: It is possible to do this with fetch, but that won't work since we're doing this anonymously (and we don't want to persist anything on the server). 

What we need to do is use an HTML form and attach the files to it programatically using JavaScript, but my initial investigations seemed to indicate that was impossible. On more recent browsers it is possible to do this, but only if the user has already made files available through a user input action (so it doesn't help us so much):

https://stackoverflow.com/questions/47515232/how-to-set-file-input-value-when-dropping-file-on-page

I'll update the comment before landing.

server/notebooks/views.py
wlach: We should be respecting the existing limit on the size of uploads, which limits the total size of the form to 2.5 MB:

https://docs.djangoproject.com/en/3.0/ref/settings/#data-upload-max-memory-size

I'm not sure if there's really any more risk here than there would be of the user spamming us through any of the existing mechanisms provided by the server. Note that this method doesn't actually *create* anything on the server -- it just echoes the content of the files back to the client.

jezdez: Right, I meant that `len(file_content)` is risky in terms of memory usage when accepting any content without rate limiting. The big advantage of FileField is that it uses file system calls to determine the size of a temporarily stored file, which is a lot less problematic.

server/notebooks/views.py
jezdez: I would recommend using a technique like the Django FileField to do this since even k8s nodes have a `/tmp` directory to persist files temporarily. My worry is mostly around this writing an own file handling, which for a public and anonymous upload API is really a bad idea.

That said, maybe I'm missing how this API will be used, usually all kinds of CLIs like curl can send regular form files, and if there is an integration in a frontend scenario there would need to be a file selector anyway. Could you elaborate what the use case of the API is?

server/notebooks/views.py
wlach: Gotcha, I would certainly prefer to use FileField as well were it possible. Unfortunately it doesn't seem like it is given the above. :(

server/notebooks/views.py
wlach: There are a few use cases:

* Creating a notebook via a python script by opening a page and submitting a hidden form (see the demo repo here: https://github.com/wlach/create-iodide-notebook)
* Creating a notebook via a web extension (by creating a hidden form and submitting it)
* Creating a notebook from a web application (by creating a hidden form and submitting it)

There are no cases that I have in mind that would use a file selector directly, although it would be nice to allow it.

--

Doing even more research, it looks like what I thought impossible *is* possible, using a DataTransfer object! See this github issue and the associated fiddle:

https://github.com/whatwg/html/issues/3269#issue-278914102

I'll give this a try and get back to you later :)

wlach: Woohoo! It worked 🎉 We're now using Django's file handling here.

server/notebooks/views.py
wlach: Turns out it was possible to use Django's file handling after all by using some client side tricks to add the files to the form.

wlach:
@jezdez feel no obligation or rush to review again, but if you could take a quick look at the python parts I would be grateful

wlach:
I went ahead and merged this, but feel free to open issues for anything you see.

jezdez:
Uh, sorry, hadn't time yet to look at it, happy you found a way.

PR #2855, opened 2020-05-12
Don't warn when navigating away when unnecessary (fixes #2537)

### LINKED ISSUE #2537: further refinements to navigation interception
along the lines of what I described in https://github.com/iodide-project/iodide/issues/2403#issuecomment-549942178 , i think there is more room to refine how and when we intercept navigation.

I think the desired behavior is that we should always make sure a user really intends to navigate away if doing so would cause them to lose work, either edits or novel computed state.

- if you arrive at the page in Report view and never enter explore view, don't alert on nav away (since no novel state can be lost)
- if you make any edits and then eval and then attempt to navigate away, alert since novel state may be lost

are there other rules or cases we want to consider? @wlach @Algentile what  do you guys think?

Moreover, i think we probably don't need to specifically handle delete/backspace keys since the nav actions they trigger are a subset of all nav actions, and we _always_ want to confirm a users intent during a nav action, no matter how it is triggered. In this case, we can probably factor out the key handling code.

wlach:
Only warn when navigating away when some novel state will be lost--
this is only when modifications were made locally and not saved
(e.g. due to interaction with a trial notebook) or the kernel
was executed after a modification (indicating the user may be in
the middle of some complicated operation)

### Pull Request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # (Unreleased; add upcoming change notes here)
 
+- Don't warn unnecessarily when navigating away from a notebook (fixes #2537)
+
 # 0.19.1 (2020-04-06)
 
 - GitHub authentication no longer uses URL query parameters (fixes #2682) (#2751)
--- a/src/editor/actions/__tests__/editor-actions.test.js
+++ b/src/editor/actions/__tests__/editor-actions.test.js
@@ -15,9 +15,10 @@ const mockStore = configureMockStore(middlewares);
 describe("basic editor actions", () => {
   it("updateIomdContent works as expected", () => {
     const iomd = "foo";
-    expect(updateIomdContent(iomd)).toEqual({
+    expect(updateIomdContent(iomd, true)).toEqual({
       type: "UPDATE_IOMD_CONTENT",
-      iomd
+      iomd,
+      userInitiated: true
     });
   });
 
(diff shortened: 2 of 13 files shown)


bcolloran:
thansk for taking this up will! couple comments:
* this can all be handled in existing reducer cases without introducing new cases, nor the actions that go with them, and also without needing to dispatch those actions from within components -- the "UPDATE_IOMD_CONTENT" reducer can check whether `iomdModifiedLocally` is false, and set it to true (which will catch the first edit). likewise, the "SET_KERNEL_STATE" case can check `hasUniqueKernelState`
* "hasUniqueKernelState" and "iomdModifiedLocally" are slightly ambiguous, i think the concepts we're really after here are about what the user has done during this session. "iomdModifiedLocally" sounds like it relates to a client/server concept, "hasUniqueKernelState" sounds a bit more complicated than needed. to get to the idea that we're really tracking what has happened during the user's session, maybe something like `sessionHasUserEvals` and `sessionHasUserEdits` (or `sessionHasEvals` and `sessionHasEdits`)


thanks!

wlach:
Hey @bcolloran -- thanks for the feedback! Unfortunately we can't reuse the `UPDATE_IOMD_CONTENT` reducer, as a bunch of things can cause its evaluation besides the user modification of the notebook. I think the kernel state one can be reused though-- I was on the fence about doing that, so if you think it's best I'm happy to make the change.

I also like your naming suggestions.

wlach:
Actually figured out a way to incorporate this into `UPDATE_IOMD_CONTENT` reducer, by specifying a userInitiated argument to it.

wlach:
Let's merge this, can always follow up later.

PR #2895, opened 2020-06-04
test_from_template_view

Ralts1337:
@wlach


server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,43 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
 
+@pytest.mark.parametrize("file_size", [1, 10485764])
+def test_from_template_view(client, file_size):
+    endpoint = "from-template"
+
+    # endpoint should not accept a GET request
+    resp = client.get(
+        reverse(endpoint)
+    )
+    assert resp.status_code == 405
```
wlach, line 163: Each of these should be separate test cases, it makes it easier to debug what's working / what isn't when there are changes

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,43 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
         assert len(response.redirect_chain) == 0
 
 
+@pytest.mark.parametrize("file_size", [1, 10485764])
```
wlach, line 155: There's no need to create an actual 10 megabyte file here, you can use the settings fixture to change `MAX_FILE_SIZE` for this test. There's an existing example of this here:

https://github.com/iodide-project/iodide/blob/cfc9722816ec5d2de70bc43ced56fcdc19b3d639/server/tests/test_file_update_operations.py#L56

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,43 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+
+    # HTTPBadRequest if no iomd is provided
+    resp = client.post(
+            reverse(endpoint),
+            {
+                "iomd": "",
+            },
+        )
+    assert resp.status_code == 400
+    assert resp.content == b"Must specify iomd template"
```
wlach, line 173: Same, this should be its own test

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,43 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+    with tempfile.NamedTemporaryFile(mode="w+") as f:
+        for _ in range(file_size):
+            f.write("0")
+        f.seek(0)
+        resp = client.post(
+            reverse(endpoint),
+            {
+                "iomd": "Test IOMD",
+                "file": f.name,
+            },
```
wlach, line 184: This is not actually posting the file, try this technique:

https://stackoverflow.com/a/11170472

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,43 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+        )
+        if file_size == 10485764:
```
wlach, line 186: Should be something like `if file_size > settings.MAX_FILE_SIZE:` if you follow the suggestion above

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,43 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+            {
+                "iomd": "",
+            },
+        )
+    assert resp.status_code == 400
+    assert resp.content == b"Must specify iomd template"
+
+    with tempfile.NamedTemporaryFile(mode="w+") as f:
+        for _ in range(file_size):
+            f.write("0")
```
wlach, line 177:
```suggestion
        f.write("0" * file_size)
```

server/tests/test_notebook_view.py
```diff
@@ -151,6 +151,72 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+def test_from_template_view_reject_no_iomd(client):
+
+    resp = client.post(reverse("from-template"), {"iomd": "",},)
+    assert resp.status_code == 400
+    assert resp.content == b"Must specify iomd template"
+
+
+def test_from_template_view_file_too_big(client, settings):
+    settings.MAX_FILE_SIZE = 8
+    with tempfile.NamedTemporaryFile(mode="w+") as f1, tempfile.NamedTemporaryFile(mode="w+") as f2:
```
wlach, line 169: It would probably be better to use a text i/o object here, instead of a temporary file:

https://docs.python.org/3/library/io.html#text-i-o

server/tests/test_notebook_view.py
```diff
@@ -151,6 +151,72 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+                "filename": "tempFile",
+                "file1": open(f1.name),
+                "file2": open(f2.name),
+            },
+        )
+        assert resp.status_code == 400
+        real_file_name = f2.name[f2.name.rfind("/") + 1 :]
+        assert resp.content == bytes(
+            f"File {real_file_name} exceeds maximum file size {settings.MAX_FILE_SIZE}", "utf-8"
+        )
```
wlach, line 188: This is causing a linter error. I think actually the name of the file is not so important here, something like this should be fine:

```suggestion
        assert "exceeds maximum file size" in resp.content.decode("utf-8")
```

server/tests/test_notebook_view.py
```diff
@@ -151,6 +151,72 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+            reverse("from-template"),
+            {
+                "iomd": "Test IOMD",
+                "title": "Test title",
+                "file1": open(f1.name),
+                "file2": open(f2.name),
+            },
+        )
+        f1_name = f1.name[f1.name.rfind("/") + 1 :]
+        f2_name = f2.name[f2.name.rfind("/") + 1 :]
```
wlach, line 208: Try just using `os.path.basename` here, a bit simpler and won't cause linter errors:

https://docs.python.org/3/library/os.path.html#os.path.basename

(in fact you can just inline that in the check below instead of calculating a temporary variable)

server/tests/test_notebook_view.py
Ralts1337: > It would probably be better to use a text i/o object here, instead of a temporary file:
> 
> https://docs.python.org/3/library/io.html#text-i-o
> 
> Not strictly necessary to change (it's not that big a deal), just letting you know.

great way to solve this problem. since we don't really need file names. thank you

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,70 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+    import io
+
+    f1 = io.StringIO("7 bytes")
+    f2 = io.StringIO("more than 8 bytes")
+    resp = client.post(
+        reverse("from-template"),
+        {
+            "iomd": "Test IOMD",
+            "title": "Test title",
+            "filename": "tempFile",
```
wlach, line 179: Pretty sure this isn't needed?

```suggestion
```

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,70 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
+def test_from_template_view_reject_empty_iomd(client):
+
+    resp = client.post(reverse("from-template"), {"iomd": ""},)
+    assert resp.status_code == 400
+    assert resp.content == b"Must specify iomd template"
+
+
+def test_from_template_view_file_too_big(client, settings):
+    settings.MAX_FILE_SIZE = 8
+    import io
```
wlach, line 170: Can you put this on top of the file?

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,70 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
...
 
 
+def test_from_template_view_reject_get_request(client):
+
+    resp = client.get(reverse("from-template"))
+    assert resp.status_code == 405
+
+
+def test_from_template_view_reject_empty_iomd(client):
+
```
wlach, line 162:
```suggestion
```

server/tests/test_notebook_view.py
```diff
@@ -151,6 +152,70 @@ def test_tryit_view(client, fake_user, logged_in, iomd):
         assert len(response.redirect_chain) == 0
 
 
+def test_from_template_view_reject_get_request(client):
+
```
wlach, line 156:
```suggestion
```

wlach:
💯

PR #2935, opened 2020-08-06
Wait for database before running migrations

nrgoncalves:
This prevents a race condition while launching the containers for
local development, whereby the migrations script is sometimes being
executed before the database is ready to accept connections.

This borrows the approach currently used in the CI pipeline, as suggested by @wlach 

https://github.com/iodide-project/iodide/blob/5c38c734420edb37ccff5ea3243d6125a498be8c/.circleci/config.yml#L140

### Pull Request checklist

- [ ] **Documentation**: Not sure if this needs to be documented
- [ ] **Changelog**: Infrastructure related change, so probably does not apply 
- [ ] **Tests**: Infrastructure related change, so probably does not apply

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # (Unreleased; add upcoming change notes here)
 
+- Wait for database before running migrations (#2935)
+
 # 0.20.2 (2020-07-16)
 
 - Allow new windows opened from iodide to submit forms (#2919)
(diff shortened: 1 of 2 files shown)


wlach:
Looks good @nrgoncalves, just wanted to confirm that it fixes the problem for you before landing?

wlach:
Oh, also, a quick Changelog entry would be good. Just using the title for this PR in the log should be enough.

nrgoncalves:
thanks @wlach, I confirm that this works for me -- I tested on a couple of different machines and could not reproduce the issue.

wlach:
Thanks so much!

PR #3001, opened 2020-12-15
Clarify pyodide status, link to starboard

wlach:
Pyodide development continues, also let's link to Starboard as I think 
it's a project in a similar spirit to Iodide and worth supporting.


wlach:
Got a verbal ok from Mike on this and it's been a few days, going to land this.

PR #3015, opened 2020-12-31
Link to fixed pyodide version in documentation

dalcde:
The relevant page may be (re)moved in future pyodide versions.


wlach:
Thanks @dalcde -- I may at some point update pyodide to use the new public API (I don't actually expect it to be very hard), but this is a good stopgap.

